Issue #11, opened 2020-09-17
Always return the Pipeline subclass/implementation from the module inspection


tdruez:
The current `get_pipeline_class` implementation may return the `Pipeline` base class in place of the actual subclass we are looking for. This appears to be random since it is based on the order of the list returned by `inspect.getmembers`.

PR #12, opened 2020-09-17
Fix the get_pipeline_class to only return subclass of Pipeline #11

tdruez:


pombredanne (approved):
LGTM. Thanks!

PR #47, opened 2020-11-13
32 Add license app with Django views from the ScanCode.io UI

tdruez:
This PR is about making the licenses data available from the ScanCode.io server UI.


pombredanne (approved):
LGTM!

PR #53, opened 2020-11-17
Simplify the Makefile `clean` rule

cco3:
This change updates the Makefile `clean` rule to delete everything
listed in .gitignore.

Nothing too important seemed to be missed, but this seemed like a
simpler approach.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,8 +80,7 @@ valid: isort black
 
 clean:
 	@echo "-> Cleaning the Python env"
-	rm -rf bin/ lib/ lib64/ include/ build/ dist/ pip-selfcheck.json pyvenv.cfg scancodeio.egg-info
-	find . -type f -name '*.py[co]' -delete -o -type d -name __pycache__ -delete
+	git clean -Xdf
 
 migrate:
 	@echo "-> Applying migrations"


pombredanne (commented):
@cco3 I did not knew about this git clean. That's awesome!
Though it has one minor drawback that I need to pounder: it assumes that we are working from a checkout and that git is available/installed on the target host, and I am not sure that's a given thing.

cco3:
Good point.

cco3:
Just came across one example where this would be helpful.  `make clean` doesn't actually cleanup after `make envfile`.

pombredanne:
@cco3 re:

> Just came across one example where this would be helpful. make clean currently doesn't actually cleanup after make envfile.

Good catch but this is actually by design as we did not want the .env file to be wiped out by a mere (quite routine) clean.

That said we could have a "gitclean" target in the Makefile that would be separate (and would assume that git is there and that you want to really "git clean"?

cco3:
Ah, OK.  If you really want clean to just be a subset of generated files, we can just close this.  I just started using the git command in my routine with playing with the tool and thought it might be helpful for others, but if not, no worries.

Issue #52, opened 2020-11-17
Create a database-backed commoncode.resource.Codebase-like class


pombredanne:
We should be able to walk and navigate a CodebaseResource the same we we do it with commoncode.resources.Codebase and VirtualCodebase but using the DB as a base (as opposed to the real filesystem or a JSON scan)

Issue #24, opened 2020-09-25
design-needed: Add GUI to review scanpipe results 


pombredanne:
We should have some UI that's better than the JSON API and better than the Django Admin.
May be something such as the scancode-workbench?

JonoYang:
I think using the scancode-workbench layout as a start is a good idea. It has the basics in terms of how we want to view/filter Resource data. I find using the tree view on the side to navigate/filter for the files under a certain directory to be a more intuitive/quicker way to navigate the rows of data.

PR #60, opened 2020-11-25
Add ProjectCodebase to walk and navigate CodebaseResource loaded from the DB #52

tdruez:
Add also a get_tree function compatible with scanpipe.CodebaseResource and commoncode.Resource
This function is a pre-requisite for the upcoming review UI https://github.com/nexB/scancode.io/issues/24 and will be the data source for the tree view.


scanpipe/tests/test_models.py
```diff
@@ -243,6 +248,7 @@ def test_scanpipe_run_model_get_run_id_method(self):
 
     def test_scanpipe_run_model_queryset_methods(self):
         now = timezone.now()
+        Run.objects.all().delete()
```
pombredanne, line 251: Is this because tests are not independent?

scanpipe/pipes/codebase.py
```diff
@@ -0,0 +1,83 @@
...
+    return resource["name"].lower()
+
+
+def get_tree(resource, fields, codebase=None):
+    """
+    Return a tree as a dict structure starting from the provided `resource`.
+
+    The following classes are supported for the input `resource` object:
+     - scanpipe.models.CodebaseResource
+     - commoncode.resource.Resource
```
pombredanne, line 38: That's great that you can support `commoncode.resource.Resource`. Yet I wonder if this may also be needed the other way too? e.g. feed a scancode plugin `process_codebase()` method a codebase object subclass or simialr that's behaving the exact same way as `commoncode.resource.Codebase` and that would yield objects that are behaving exactly like a `commoncode.resource.Resource` each being actually wrapping CodebaseResource instances.

An example of the `process_codebase()` is there https://github.com/nexB/scancode-toolkit/blob/develop/src/scancode/plugin_mark_source.py#L63

The benefit would be that scancode plugins could then be used as-is and fed a codebase that they process alright (and that could be filtered as some query set). This includes all the output plugins for instance.

pombredanne (commented):
LGTM. +1  See my comments inline

scanpipe/pipes/codebase.py
tdruez: The `walk` is available on the ProjectCodebase for that purpose.
Also, on the CodebaseResource side, `is_file`, `is_dir`, `children`, `descendants` are implemented.
We can add more in the future as needed.

scanpipe/tests/test_models.py
tdruez: The Run QS should be scoped to the project instead, this may be cleaner.
I'll commit a fix before the merge.

Issue #82, opened 2021-02-04
Remove the dependency on Metaflow


tdruez:
The benefits from using Metaflow are now outweighed by the issues and lack of customization (logging for example) forcing us into complex workarounds.

Since we do not use much of the Metaflow features (we've only used Metaflow for the structure of the Pipelines), we do not have a lot of code that depends on it. 
The required code changes will mostly be needed on the Pipelines side.

Those change will also allow us to have more control over the execution of Pipelines (logging, profiling, branching, ...)
It will remove 1 level of indirection in the execution stack since Metaflow is always run as a subprocess call (it cannot be called within Python by design).

mjherzog:
@tdruez Does this mean that most of what we are using from Metaflow is the approach to structuring a Pipeline without really using much Metaflow code?

tdruez:
@mjherzog Yes, we are basing our Pipelines on the Metaflow Flow structure but inherit from its execution approach (through subprocess) that is not great in the ScanCode.io context (asynchronous tasks).

PR #89, opened 2021-02-05
Remove the dependency on Metaflow #82

tdruez:


scanpipe/management/commands/__init__.py
```diff
@@ -64,6 +64,13 @@ def get_run_status_code(self, run):
         return status
 
 
+def indent(text, by=0, sep="\n"):
+    """
+    Indent each line of the provided `text`.
+    """
+    return sep.join(" " * by + line for line in text.split(sep))
```
pombredanne, line 71: Could this be a job for the stdlib textwrap? https://docs.python.org/3/library/textwrap.html

scanpipe/management/commands/graph.py
```diff
@@ -26,19 +26,46 @@
...
 
 def is_graphviz_installed():
-    exitcode, _ = subprocess.getstatusoutput("which dot")
+    exitcode = subprocess.getstatusoutput("which dot")[0]
     if exitcode == 0:
         return True
     return False
 
 
+def pipeline_graph_dot(pipeline_class, fontname="Helvetica", shape="record"):
```
pombredanne, line 42: I wonder if in the future we could eschew using `dot` at all.

scanpipe/management/commands/run.py
```diff
@@ -30,33 +30,18 @@
...
+        run = self.project.get_next_run()
 
         if not run:
-            raise CommandError(f"No pipelines to {action} on project {self.project}")
-
-        self.stdout.write(f"Pipeline {run.pipeline} {action} in progress...")
-        getattr(run, task_function)()
+            raise CommandError(f"No pipelines to run on project {self.project}")
 
+        self.stdout.write(f"Pipeline {run.pipeline} run in progress...")
```
pombredanne, line 41: What is `self.stdout`?

scanpipe/pipelines/docker.py
```diff
@@ -67,92 +47,77 @@ def extract_layers(self):
...
-        self.next(self.end)
-
-    @step
-    def end(self):
-        """
-        Analysis completed.
-        """
-
 
-if __name__ == "__main__":
```
pombredanne, line 157: we are missing out the direct execution of a pipeline, but I will not miss this at all. A proper managed CLI is better.

scanpipe/tests/pipelines/do_nothing.py
```diff
@@ -0,0 +1,40 @@
...
+#
+# ScanCode.io is a free software code scanning tool from nexB Inc. and others.
+# Visit https://github.com/nexB/scancode.io for support and download.
+
+from scanpipe.pipelines import Pipeline
+
+
+class DoNothing(Pipeline):
+    """
+    A pipeline that does nothing, in 2 steps.
```
pombredanne, line 28: :+1:

pombredanne (approved):
LGTM!

scanpipe/management/commands/run.py
tdruez: This is the default Django management command API https://docs.djangoproject.com/en/3.1/howto/custom-management-commands/#module-django.core.management

scanpipe/management/commands/graph.py
tdruez: We can get rid of the feature now we have a way to visualize Pipeline in the UI.

scanpipe/management/commands/__init__.py
tdruez: Yes, thanks for the tip.

pombredanne:
BTW, we may need to update the documentation accordingly?

tdruez:
@pombredanne the only reference to Metaflow in the docs is this:
> ScanPipe was originally developed using Django, ScanCode Toolkit, and Metaflow to help boost productivity of code analysts who work on a wide variety of software composition analysis projects.

Issue #91, opened 2021-02-09
Add Pipelines as plugins


tdruez:
Until now pipelines have been plain Python scripts living a specific Python module. To improve modularity, pipelines should be installable instead as Python packages that one can add to a scancode.io installation.
A simple design is to use setuptools entry points as way for a Python package to signal it is contributing a pipeline.

PR #95, opened 2021-02-12
Pipeline registration through distribution entry points #91

tdruez:
Those changes allow to install new external pipelines using the "scancodeio_pipelines" entry point group as a plugin mechanism:

```
entry_points={
        "scancodeio_pipelines": [
            "docker = scanpipe.pipelines.docker:Docker",
            ...
        ],
    },
```


tdruez:
@pombredanne can you take a look at those changes and give your input?
The new pipeline registration logic is at https://github.com/nexB/scancode.io/pull/95/files#diff-450348e775480cb66b78166191948b47e5e49d1a97563f22a219269b4f977071R38

docs/scanpipe-tutorial-1.rst
```diff
@@ -59,13 +59,13 @@ Step-by-step
...
-    "[SUCCESS] scanpipe/pipelines/docker.py"
+    "[SUCCESS] docker"
 
 - Get the results of the pipeline run as a JSON file using the ``output`` command::
 
-    $ scanpipe output --project staticbox results.json
+    $ scanpipe output --project staticbox --format json
 
-- Open the ``results.json`` in your preferred viewer.
+- Open the ``output/results-<timestamp>.json`` in your preferred viewer.
```
pombredanne, line 68: IMHO the UX would be better if we instead require the output file to be named? Or at least it should be possible to name the file. I find sometimes frustrating to have files created or display and I cannot control their name? To discuss!

scanpipe/api/serializers.py
```diff
@@ -144,7 +144,7 @@ def create(self, validated_data):
             project.add_input_file(upload_file)
 
         if pipeline:
-            project.add_pipeline(pipeline, start_run=bool(upload_file))
+            project.add_pipeline(pipeline, start=bool(upload_file))
```
pombredanne, line 147: Nit: I wonder if start if the proper name to use there: what about `run` or `launch` or `execute` ? e.g. a pipeline has a start and end step, but the action to execute may not be best described as "starting" which is state more than an action?

scanpipe/api/views.py
```diff
@@ -91,11 +89,10 @@ def results_download(self, request, *args, **kwargs):
     @action(detail=False)
     def pipelines(self, request, *args, **kwargs):
         data = {}
-        for location, name in scanpipe_app_config.pipelines:
+        for name, pipeline_class in scanpipe_app_config.pipelines.items():
             data[name] = {
-                "location": location,
-                "description": get_pipeline_class(location).get_doc(),
-                "steps": get_pipeline_graph(location),
+                "description": pipeline_class.get_doc(),
```
pombredanne, line 94: Is this the structure returned by the API? and if so does it have this shape?
```
"name" : {
    "description": "lorem ipsum.....",
    "steps":  [.....]
}
```


If yes, I wonder if have a pipeline name as a mapping key make sense as this is not our ABC data API convention to always use names/value pairs for data and avoid to have values as keys, e.g. something like:

```
[{
    "name" :  "some name",
    "description": "lorem ipsum.....",
    "steps":  [.....]
}]
```

scanpipe/apps.py
```diff
@@ -20,45 +20,51 @@
...
-        project_root = Path(__file__).parent.parent.absolute()
-        pipelines_dir = project_root / "scanpipe" / "pipelines"
+        entry_points = importlib_metadata.entry_points()
 
-        for child in pipelines_dir.iterdir():
-            if child.name.endswith(dot_py_suffix) and not child.name.startswith("_"):
-                location = str(child.relative_to(project_root))
-                name = remove_dot_py_suffix(child.name)
-                self.pipelines.append((location, name))
+        # Ignore duplicated entries caused by duplicated paths in `sys.path`.
```
pombredanne, line 49: 1. are duplicated entries possible?
2. should we instead raise an exception and fail to avoid any possible confusion when this happens? ... especially since below you are also raising exceptions https://github.com/nexB/scancode.io/pull/95/files#diff-450348e775480cb66b78166191948b47e5e49d1a97563f22a219269b4f977071R65

scanpipe/pipelines/__init__.py
```diff
@@ -98,7 +124,7 @@ def save_errors(self, *exceptions):
         try:
             yield
         except exceptions as error:
-            self.project.add_error(error, model=self.__class__.__name__)
+            self.project.add_error(error, model=self.get_name())
 
 
 def is_pipeline_subclass(obj):
```
pombredanne, line 130: Could this be called instead just `is_pipeline`?

scanpipe/apps.py
```diff
@@ -20,45 +20,51 @@
...
+        pipeline_entry_points = set(entry_points.get("scancodeio_pipelines"))
 
-    def is_valid(self, pipeline):
+        for entry_point in pipeline_entry_points:
+            pipeline_class = entry_point.load()
+            self.register_pipeline(entry_point.name, pipeline_class)
```
pombredanne, line 54: Are we are saying here that the name given in the entry IS the name that we will use throughout? eg. as in https://github.com/nexB/scancode.io/pull/95/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R58 ?

I wonder if we should instead may be consider this name to be an internal "key" and instead rely on a name provided as a class attribute instead? ... e.g. we want that key unique but would not use that in the UI directly?

scanpipe/management/commands/__init__.py
```diff
@@ -78,10 +78,9 @@ def validate_pipelines(pipelines):
     """
     Raise an error if one of the `pipelines` is not available.
     """
-    for pipeline_location in pipelines:
-        if not scanpipe_app_config.is_valid(pipeline_location):
-            pipeline_list = [location for location, _ in scanpipe_app_config.pipelines]
+    for pipeline_name in pipelines:
```
pombredanne, line 81: nit: since `for pipeline_name in pipelines` iterates over names, may be `pipelines` should be really be called `pipeline_names` to distinguish with cases where the list of pipelines is really a list of classes or a dict?
I see `pipelines` used as collection name for things that are very different

scanpipe/management/commands/graph.py
```diff
@@ -85,10 +83,9 @@ def add_arguments(self, parser):
 
     def handle(self, *pipelines, **options):
         if options["list"]:
-            for location, _ in scanpipe_app_config.pipelines:
-                self.stdout.write("- " + self.style.SUCCESS(location))
-                pipeline_doc = get_pipeline_doc(location)
-                self.stdout.write(indent(pipeline_doc, "  "), ending="\n\n")
+            for name, pipeline_class in scanpipe_app_config.pipelines.items():
```
pombredanne, line 86: same nit as above wrt. using `pipelines`... what is a pipeline? is this a Pipeline subclass? if that's the case, then I would think that `pipelines` would be a sequence of Pipelines, `pipeline_names` or `pipeline_keys` a sequence of Pipeline keys and `pipelines_by_key` or `pipelines_by_name` mappings? 
for discussion

pombredanne (changes requested):
Looking good! I have some nits for your consideration.

scanpipe/api/serializers.py
tdruez: `Run` is the current name of the Run that wrap a pipeline execution, so "run a run" is not great.
Execute is the action on the Pipeline class to execute each step.

> a pipeline has a start and end step
This is not the case anymore.

So the idea here is to "start" a Run which is a bit more (celery task management) than a Pipeline execution.

docs/scanpipe-tutorial-1.rst
tdruez: This requires its own ticket.

scanpipe/api/serializers.py
tdruez: After discussion, we decided to use `execute` instead of `start`.

PR #107, opened 2021-03-06
Add OpenAPI 3 Schema using drf-spectacular

### LINKED ISSUE #106: Using Swagger for API Documentation
The root api view (`/api`) currently gives us the default view from Django Rest Framework. Would it be better if we use Swagger, like on vulnerablecode? If yes, I can make that change.

hanif-ali:
Adds a Swagger schema accessible through the navigation bar. I think adding this would help us document the API in a standardized manner. I went with `drf-spectacular` because it was being used in vulnerablecode already and provides more extensibility than the alternatives.

closes #106


hanif-ali:
@tdruez Please review this.

scancodeio/urls.py
```diff
@@ -39,6 +41,8 @@
 
 urlpatterns = [
     path("admin/", admin.site.urls),
+    path("api/schema/", SpectacularAPIView.as_view(), name="schema"),
+    path("api/schema/swagger-ui/", SpectacularSwaggerView.as_view(), name="swagger-ui"),
```
majurg, line 45: Would the swaggerUI only effect these schema endpoints?

scanner/models.py
```diff
@@ -199,7 +199,7 @@ def delete(self, *args, **kwargs):
         return deleted
 
     @property
-    def status(self):
+    def status(self) -> str:
```
majurg, line 202: Not sure we want to mix and match type annotations. We should probably remove this.

scanpipe/api/serializers.py
```diff
@@ -133,11 +133,11 @@ class Meta:
             "discovered_package_summary",
         ]
 
-    def get_codebase_resources_summary(self, project):
+    def get_codebase_resources_summary(self, project) -> dict:
```
majurg, line 136: Not sure we want to mix and match type annotations. We should probably remove this.

scanpipe/api/serializers.py
```diff
@@ -133,11 +133,11 @@ class Meta:
...
         queryset = project.codebaseresources.all()
         return count_group_by(queryset, "status")
 
-    def get_discovered_package_summary(self, project):
+    def get_discovered_package_summary(self, project) -> dict:
```
majurg, line 140: Not sure we want to mix and match type annotations. We should probably remove this.

scanpipe/models.py
```diff
@@ -118,7 +118,7 @@ def task_state(self):
         return AsyncResult(str(self.task_id)).state
 
     @property
-    def execution_time(self):
+    def execution_time(self) -> int:
```
majurg, line 121: Not sure we want to mix and match type annotations. We should probably remove this.

scanpipe/models.py
```diff
@@ -280,15 +280,15 @@ def get_root_content(directory):
         return [str(path.relative_to(directory)) for path in directory.glob("*")]
 
     @property
-    def input_root(self):
+    def input_root(self) -> list:
```
majurg, line 283: Not sure we want to mix and match type annotations. We should probably remove this.

scanpipe/models.py
```diff
@@ -280,15 +280,15 @@ def get_root_content(directory):
...
         """
         Return the list of all files and directories of the input/ directory.
         Only the first level children are listed.
         """
         return self.get_root_content(self.input_path)
 
     @property
-    def output_root(self):
+    def output_root(self) -> list:
```
majurg, line 291: Not sure we want to mix and match type annotations. We should probably remove this.

majurg (changes requested):

pombredanne:
@hanif-ali gentle ping... are you planning to finish this? Otherwise we may close it for now.

PR #112, opened 2021-03-16
Do not create Package if one of the required fields is missing...

tdruez:
...create a ProjectError instead.
Also, move the package creation from the DiscoveredPackage to the CodebaseResource model.

### DIFF
--- a/scanpipe/models.py
+++ b/scanpipe/models.py
@@ -22,10 +22,10 @@
 
 import re
 import shutil
-import traceback
 import uuid
 from contextlib import suppress
 from pathlib import Path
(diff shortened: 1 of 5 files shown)


scanpipe/models.py
```diff
@@ -432,12 +432,16 @@ def add_error(self, error, model, details=None):
         Create a ProjectError record from the provided `error` Exception for this
         project.
         """
+        traceback = ""
```
pombredanne, line 435: Have you tried if the simpler `traceback.format_exc` may work, e.g. if you are already in an exception?

PR #115, opened 2021-03-23
Fix/scanpipe project delete modal

Sagarsangwan:
Delete modal close button not working and cancel button working as delete


tdruez:
Fixed in https://github.com/nexB/scancode.io/commit/2b31db6c7bfc536d554b8fe8a0c5671e4f2180b4

PR #118, opened 2021-03-23
fix :: project name clean before saving to DB

sagarsangwan:


scanpipe/forms.py
```diff
@@ -117,6 +124,10 @@ def __init__(self, *args, **kwargs):
         name_field.widget.attrs["autofocus"] = True
         name_field.help_text = "The unique name of your project."
 
+    def clean(self):
```
tdruez, line 127: No need to override the `clean` method here. You can simply use the `clean_<fieldname>()` Django's API.
https://docs.djangoproject.com/en/3.1/ref/forms/validation/#cleaning-a-specific-field-attribute

tdruez (changes requested):
Thanks for the PR. 
Could you add a unit test along the code changes?

scanpipe/forms.py
Sagarsangwan: > Thanks for the PR.
> Could you add a unit test along the code changes?

Yes, will do that shortly.

tdruez: Hi @Sagarsangwan, any progress on this change?

tdruez:
Closing, see https://github.com/nexB/scancode.io/pull/158

Issue #109, opened 2021-03-11
Add option to drag and drop files in the UI.


Abhirup-99:
This would make it easier to drop a complete folder rather than individually clicking on the files to upload them. 
Check this out: https://www.dropzonejs.com/

tdruez:
@Abhirup-99 PR welcome :)

Abhirup-99:
Would push one shortly :).

AvishrantsSh:
I am a little confused regarding this. So I'll just break it into parts.
1. How do you scan a complete folder? I mean is this feature supported, and if so how to do this?
2. Are there any plans to improve upon the UI of this project? If so, is it required to properly format and place the dropzone?

tdruez:
> How do you scan a complete folder? I mean is this feature supported, and if so how to do this?

You can compress your folder in a zip archive (or any format supported by extract code) and upload it as an input file.
Alternatively, you can copy any content you want to scan in a project "work directory" directly on the file system. The "work directory" of a project is displayed in the header of the project detail view.

tdruez:
> Are there any plans to improve upon the UI of this project?

Yes, the web UI is fairly new and constantly evolving. Any ideas on improving the UI are welcome.

> If so, is it required to properly format and place the dropzone?

Can you expand on your thoughts on this? I don't know what you mean by "properly format and place".

tdruez:
@Abhirup-99 any progress on the Dropzone.js implementation?

tdruez:
Drag and drop feature added for files input. Thanks @AvishrantsSh

PR #144, opened 2021-04-09
Drag and Drop Feature

AvishrantsSh:
Added initial functionality to drag and drop files while adding a new project. #109


tdruez:
@AvishrantsSh thanks for the PR, could you revert all the formatting/indenting changes? It makes a lot of noise and makes it harder to review the real changes. We can make a pass on formatting as a separate PR.

Issue #130, opened 2021-04-06
incorrect pipeline status when restarted the server


chinyeungli:
I've created couple projects and launch a docker pipeline on each of them, but then the server is dead. Therefore, I restarted the server, and when I open the project page again, the "status" next to the pipelines is still showing "Running" which is incorrect.

tdruez:
@AvishrantsSh I'd like to get your input on this one. Any ideas on how we could improve this?

To reproduce:
- Start the webserver: `$make run`
- Start any pipeline execution
- Kill the webserver process
- Restart the webserver and look at your pipeline state, still displayed as "running"

The pipeline will show as "running" forever while not running anymore. That status display is based on the database state of the Pipeline run, it was marked as "started" at the start of the pipeline execution, but since the worker process was "killed", it never got the chance to be marked as "ended/failed".

AvishrantsSh:
@tdruez, I've looked into this issue and I believe there are a few workarounds:

- We can implement a custom `signal` for server shutdown and perform database cleansing at that time.
- When the server is started, we can look for entries in `Run` model that have no valid `task_exitcode` and perform cleansing accordingly. It can be implemented in `wsgi.py` using a filter like `task_exitcode__isnull = True`.

Personally, I feel that the second solution makes more sense as it is guaranteed that no celery job will be running at that time that might interfere with the cleansing. Besides, it is also simpler to implement than signals.

AvishrantsSh:
I tried to implement the second solution and it seems to work perfectly.

tdruez:
@AvishrantsSh thanks for your input. This approach works alright in local "runserver" mode.

But in the full stack mode (redis + celery worker + webserver + ALWAYS_EAGER = False) the "When the server is started" would not work since each service is run separately. The celery worker could still be working on a task while you restart the webserver for example.

AvishrantsSh:
@tdruez As you rightly said, the approach that I mentioned should work in "offline" mode.
Now as for the complete solution, I have some doubts. First, can Celery update the db when django server is shut down? If it can then I don't think there will be any issue with this approach as we can set the cleansing to work only with local development.
However, if Celery can't update the record, then it is a different story.

tdruez:
> Celery update the db when django server is shut down?
Yes, of course, the celery workers access the database independently from the django app.

Also, we can have multiple celery workers running, so doing the cleanup on starting a worker would not be good enough.

AvishrantsSh:
[...]

PR #171, opened 2021-05-04
Fix pipeline status when restarting the server #130

AvishrantsSh:
Fixed Issue #130.
Signed-off-by: Avishrant Sharma <avishrants@gmail.com>

### DIFF
--- a/scancodeio/settings/base.py
+++ b/scancodeio/settings/base.py
@@ -222,6 +222,7 @@ CRISPY_TEMPLATE_PACK = "bootstrap3"
 
 # Celery
 
+CELERY_BROKER_TRANSPORT_OPTIONS = {"socket_timeout": 5, "max_retries": 1}
 CELERY_BROKER_URL = env.str("CELERY_BROKER_URL", default="redis://")
 CELERY_RESULT_BACKEND = env.str("CELERY_RESULT_BACKEND", default="redis://")
 CELERY_TRACK_STARTED = env.bool("CELERY_TRACK_STARTED", default=True)
--- a/scancodeio/wsgi.py
+++ b/scancodeio/wsgi.py
@@ -33,6 +33,28 @@ import os
 
 from django.core.wsgi import get_wsgi_application
 
+from scancodeio.celery import app
+from scanpipe.models import Run
+
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "scancodeio.settings")
+
+
+def get_celery_worker_status():
+    i = app.control.inspect()
+    # availability = i.ping()
+    # stats = i.stats()
+    # registered_tasks = i.registered()
+    active = i.active()
+    scheduled = i.scheduled()
+    result = {
+        "active": active[list(active.keys())[0]] if active else 0,
+        "scheduled": active[list(active.keys())[0]] if active else 0,
+    }
+    return result
+
+
+result = get_celery_worker_status()
+if not result["active"] and not result["scheduled"]:
+    Run.objects.filter(task_exitcode__isnull=True).update(task_exitcode=1)
 
 application = get_wsgi_application()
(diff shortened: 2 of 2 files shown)


scancodeio/settings/base.py
```diff
@@ -222,6 +222,7 @@
 
 # Celery
 
+CELERY_BROKER_TRANSPORT_OPTIONS = {"socket_timeout": 5, "max_retries": 1}
```
pombredanne, line 225: Could i make sense to use an environment variable here too?

scancodeio/wsgi.py
```diff
@@ -33,6 +33,27 @@
...
+
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "scancodeio.settings")
 
+
+def get_celery_worker_status():
+    try:
+        i = app.control.inspect()
+        active = i.active()
+        scheduled = i.scheduled()
+        if active[list(active.keys())[0]] or scheduled[list(scheduled.keys())[0]]:
```
pombredanne, line 49: Could there be a way to make this easier to read and understand? may be some intermediate variable? and is there a risk to use the first element? can this fail if the list are empty? Can they be empty? Is  there a difference between `list(active.keys())` and `list(active)` assuming `active` is some kind of mapping?
What are you trying to achieve here? get the first value of either one of two mappings?

scancodeio/wsgi.py
```diff
@@ -33,6 +33,27 @@
...
+            return True
+    except OperationalError:
+        return False
+
+
+result = get_celery_worker_status()
+if not result:
+    Run.objects.filter(task_exitcode__isnull=True).update(task_exitcode=1)
```
pombredanne, line 57: Can you elaborate what this does?

pombredanne (changes requested):
Thanks! See my comments inline. Is there a way to get a test for this too? or is the setup too complex?

scancodeio/wsgi.py
AvishrantsSh: I will change variable names and try to make it more readable. When testing on my PC, `i.active()` returned a dictionary with one key of the format `celery@<-pc name->`, so I had to use the first key by using `list(active.keys())[0]`.  If no worker is avaliable, it simply returns a dictionary object like `{'celery@avishrant-Latitude-3500': []}`.

`can this fail if the list are empty?` - @pombredanne both active() and scheduled() will always work provided that the broker is reachable.

`What are you trying to achieve here?` - This code snippet simply checks if any job is running or scheduled. If it finds such a job, the Database won't be updated. Alternatively, if no job is running and any 'Run' database entry has `running`  task status, it simply updates that field, thereby solving the problem.

scancodeio/wsgi.py
AvishrantsSh: `Run.objects.filter(task_exitcode__isnull=True).update(task_exitcode=1)` - This command is executed when no worker is available at the moment when server is started. The filter extracts all entries having `task_exitcode = NULL`, which represents `Running` state,  and updates them to represent failure in execution `task_exitcode = 1`.

scancodeio/settings/base.py
AvishrantsSh: > Is there a way to get a test for this too? or is the setup too complex?

I'm not sure how to make a test case for this. Any ideas?

AvishrantsSh: > Could i make sense to use an environment variable here too?

Yes, environment variables can be used to add Broker Configuration.

pombredanne:
@tdruez @AvishrantsSh gentle ping... what's left to do to finish this?

pombredanne (approved):

pombredanne (changes requested):
@tdruez your review is needed

tdruez:
@pombredanne This is an incomplete solution, see https://github.com/nexB/scancode.io/issues/130#issuecomment-860731952 for details.
I think we should close this PR.

tdruez:
Closing for now, see https://github.com/nexB/scancode.io/issues/130 for discussion.

PR #181, opened 2021-05-25
Add support for Windows containers

pombredanne:


scanpipe/pipes/docker.py
```diff
@@ -61,23 +61,23 @@ def extract_layers_from_images(project, images):
     Return the `errors` that may have happen during the extraction.
     """
     errors = []
-
+    # FIXME: use container-inspector extract instead
```
tdruez, line 64: Could you provide some context and reason about this change?

etc/requirements/base.txt
```diff
@@ -24,14 +24,14 @@ redis==3.5.3
 gunicorn==20.1.0
 
 # Docker
-container_inspector>=3.1.2
+container_inspector>=21.5.25
```
tdruez, line 27: Using `>=` with CalVer is problematic for API changes, see https://github.com/nexB/scancode.io/pull/181/commits/7b7fc2b11dd15a7cdae8d84912ce76e02d53f761
We should revert back to exact versions.

etc/requirements/base.txt
```diff
@@ -27,12 +27,12 @@ gunicorn==20.1.0
...
 # ScanCode-toolkit
-scancode-toolkit[packages]==21.6.7
+scancode-toolkit[packages]==21.6.11
 extractcode[full]==21.6.2
-commoncode==21.5.25
+commoncode==21.6.11
 
 # FetchCode
-fetchcode-container==1.2.3.210512; sys_platform == "linux"
+fetchcode-container>=1.2.3.210512; sys_platform == "linux"
```
tdruez, line 35: Let's use pinned down version instead ;)

scanpipe/pipelines/docker.py
```diff
@@ -63,7 +63,7 @@ def extract_layers(self):
         if errors:
             self.add_error("\n".join(errors))
 
-    def find_images_linux_distro(self):
+    def find_images_os_and_distro(self):
         """
         Find the linux distro of the images.
```
tdruez, line 68: The docstring needs an update too.

scanpipe/pipelines/windows_docker.py
```diff
@@ -0,0 +1,59 @@
...
+            cls.tag_known_software_packages,
+            cls.tag_uninteresting_codebase_resources,
+            cls.tag_empty_files,
+            cls.scan_for_application_packages,
+            cls.scan_for_files,
+            cls.analyze_scanned_files,
+            cls.tag_not_analyzed_codebase_resources,
+        )
+
+    def tag_known_software_packages(self):
```
tdruez, line 51: We need a docstring for each step.

tdruez (changes requested):

tdruez:
@JonoYang we need docstrings for all pipeline steps and pipes as parts of the documentation are automatically generated from that content.

etc/requirements/base.txt
JonoYang: I've pinned the dep.

JonoYang:
@tdruez Thanks for the review. I added/updated docstrings to the functions that were missing it.

scanpipe/pipes/docker.py
tdruez: @JonoYang Could you enter a new issue with some details of the problem? This would have more chance to be worked on than a comment in the code.
Thanks!

etc/requirements/base.txt
pombredanne: That's a container-inspector issue ... and that's my making: this ship has sailed there now and I cannot see how I can revert there.
Note that the API for this container-inspector library is fast evolving but I will thrive to keep it backward compatible as much as possible.

scanpipe/pipes/docker.py
pombredanne: @tdruez there are some issues that @JonoYang uncovered in how code is being extracted AFAIRC.

tdruez: @pombredanne I would prefer a ticket nonetheless as those `# FIXME` tend to never get dealt with.

pombredanne: Good point, https://github.com/nexB/scancode.io/issues/251 added

scanpipe/pipes/docker.py
```diff
@@ -61,7 +61,7 @@ def extract_layers_from_images(project, images):
...
```
pombredanne, line 64:
```suggestion
```

scanpipe/pipes/windows.py
```diff
@@ -0,0 +1,237 @@
...
+            license_expression="python",
+            copyright="Copyright (c) Python Software Foundation",
+            homepage_url="https://www.python.org/",
+        )
+        tag_installed_package_files(
+            project=project, root_dir_pattern=python_path, package=python_package
+        )
+
+    qs = project.codebaseresources.no_status()
+    openjdk_root_directory_name_pattern = r"/Files/(open)?jdk(-((\d*)(\.\d+)*))*"
```
pombredanne, line 154: Would this catch a non-openjdk?

scanpipe/pipes/windows.py
```diff
@@ -0,0 +1,237 @@
...
+    there are CodebaseResources that are under a known software root directory,
+    a DiscoveredPackage is created for that software package and all files under
+    that software package's root directory are considered installed files for
+    that package.
+
+    Currently, we are only checking for Python and openjdk in Windows Docker
+    image layers.
+    """
+    qs = project.codebaseresources.no_status()
+    python_root_directory_name_pattern = r"/Files/Python(\d*)"
```
pombredanne, line 118: The `/Files` in the path should be made somewhat configurable such that we can also support non-Docker based rootfs images. This is not for here but everywhere.

scanpipe/pipes/windows.py
```diff
@@ -0,0 +1,237 @@
...
+    openjdk_paths_by_versions = {}
+    for openjdk_codebase_resource in qs.filter(
+        rootfs_path__regex=openjdk_root_directory_name_pattern
+    ):
+        _, open_prefix, _, openjdk_version, _, _, _ = re.split(
+            openjdk_root_directory_name_pattern_compiled,
+            openjdk_codebase_resource.rootfs_path,
+        )
+        if not openjdk_version:
+            openjdk_version = "nv"
```
pombredanne, line 167: We need to document with a comment... or more. This can surprising

scanpipe/pipes/windows.py
JonoYang: I've updated the regex to look for openjdk root dirs in the root of the image. either `/` or `/Files/`

scanpipe/pipes/windows.py
JonoYang: I've added a note to the docstring of this function

scanpipe/pipes/windows.py
JonoYang: Regex has been updated such that we are only looking for CodebaseResources that start with `/Python*` or `/Files/Python*`

scanpipe/pipes/rootfs.py
```diff
@@ -336,3 +340,94 @@ def tag_uninteresting_codebase_resources(project):
...
+
+
+def tag_media_files_as_uninteresting(project):
+    """
+    Tags CodebaseResources that are media files to be uninteresting.
+
+    `mimes` and `types` are taken from TypeCode:
+    https://github.com/nexB/typecode/blob/main/src/typecode/contenttype.py#L528
+    """
+    mimes = (
```
tdruez, line 391: This should be directly imported from `typecode`.
It would require to make https://github.com/nexB/typecode/blob/main/src/typecode/contenttype.py#L528 available as a module variable.
In the short term, we can keep it as-is and enter a ticket on the `typecode` side.

scanpipe/pipelines/windows_docker.py
```diff
@@ -0,0 +1,81 @@
...
+# ScanCode.io is a free software code scanning tool from nexB Inc. and others.
+# Visit https://github.com/nexB/scancode.io for support and download.
+
+from scanpipe.pipelines.docker import Docker
+from scanpipe.pipes import docker
+from scanpipe.pipes import rootfs
+from scanpipe.pipes import windows
+
+
+class WindowsDocker(Docker):
```
tdruez, line 29: What about `DockerWindows` and `docker_windows` instead? This would keep all Docker based pipeline grouped in the UI.
I'm not sure about this though, what's your take?

scanpipe/pipes/windows.py
```diff
@@ -0,0 +1,250 @@
...
+    for file_name in uninteresting_files:
+        lookups |= Q(rootfs_path__iendswith=file_name)
+    for file_extension in uninteresting_file_extensions:
+        lookups |= Q(extension__icontains=file_extension)
+
+    qs = project.codebaseresources.no_status()
+    qs.filter(lookups).update(status="ignored-not-interesting")
+
+
+def tag_installed_package_files(project, root_dir_pattern, package, q_objects=[]):
```
tdruez, line 96: Do not use mutable default arguments.

scanpipe/pipes/windows.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+def _tag_openjdk_software(project):
+    qs = project.codebaseresources.no_status()
+    openjdk_root_pattern = r"^(/(Files/)?(open)?jdk(-((\d*)(\.\d+)*))*)/.*$"
+    openjdk_root_pattern_compiled = re.compile(openjdk_root_pattern)
+
+    openjdk_versions_by_path = {}
+    for openjdk_codebase_resource in qs.filter(rootfs_path__regex=openjdk_root_pattern):
+        _, openjdk_root_path, _, _, _, openjdk_version, _, _, _ = re.split(
```
tdruez, line 169: What about using the match system https://docs.python.org/3/library/re.html#match-objects for improved readability and future maintenance?
Some examples in https://github.com/package-url/packageurl-python/blob/master/src/packageurl/contrib/url2purl.py

tdruez (changes requested):
@JonoYang good work, looks like we're almost there for the merge.
I've made a round of adjustments in https://github.com/nexB/scancode.io/pull/181/commits/a48eb4c46da31c77a63b1edc4f837b35c6e8fbd5

See my comments for a last round of code improvements.

scanpipe/pipes/windows.py
tdruez: This applies to the 3 `re.split()` of the PR.

scanpipe/pipes/rootfs.py
JonoYang: @tdruez On the other hand, what if we added the `is_media` field from the license scan to the `CodebaseResource` model?

tdruez: Sure, that sounds good.

scanpipe/pipelines/windows_docker.py
JonoYang: I think it makes sense to rename the pipeline to `DockerWindows` for grouping.

scanpipe/pipes/rootfs.py
JonoYang: I've added the `is_media` field to `CodebaseResource`, removed the part that excludes `is_media` from `unsupported_fields` in `scanpipe.pipes.scancode.get_resource_info`, and updated the tests results.

scanpipe/pipes/windows.py
JonoYang: I replaced all useage of `re.split()` with `re.match()`

tdruez:
@JonoYang nice work, PR merged!
See https://github.com/nexB/scancode.io/pull/181/commits/2645864c7fb4273eb5d67ef070dc0d82ff9642b6#diff-30934e49c9b2107144b0ea45fd86cb0693bf7544f0195a3caec127e0118720a1R96 for a better fix for empty list as default parameter.

PR #226, opened 2021-07-05
GLC Scan Pipeline

tdruez:
Signed-off-by: Avishrantsh <avishrants@gmail.com>


scanpipe/pipelines/glc_scan.py
```diff
@@ -0,0 +1,62 @@
...
+# for any legal advice.
+#
+# ScanCode.io is a free software code scanning tool from nexB Inc. and others.
+# Visit https://github.com/nexB/scancode.io for support and download.
+
+from scanpipe.pipelines import scan_codebase
+from scanpipe.pipes import glc
+
+
+class GLicenseScan(scan_codebase.ScanCodebase):
```
tdruez, line 27: A more explicit name would be nice. Something like: LicenseClassifierScan

scanpipe/pipelines/glc_scan.py
```diff
@@ -0,0 +1,62 @@
...
+            cls.build_inventory_from_scan,
+            cls.csv_output,
+        )
+
+    def run_glc(self):
+        """
+        Scan extracted codebase/ content.
+        """
+        self.scan_output = self.project.get_output_file_path("scancode", "json")
+        # print(self.scan_output)
```
tdruez, line 47: Leftover?

scanpipe/pipelines/glc_scan.py
```diff
@@ -0,0 +1,62 @@
...
+            cls.run_glc,
+            cls.build_inventory_from_scan,
+            cls.csv_output,
+        )
+
+    def run_glc(self):
+        """
+        Scan extracted codebase/ content.
+        """
+        self.scan_output = self.project.get_output_file_path("scancode", "json")
```
tdruez, line 46: Since it's not a scancode ouput, let's use a proper filename, such as : "licenseclassifier"

scanpipe/pipes/glc.py
```diff
@@ -0,0 +1,97 @@
...
+from scanpipe.models import CodebaseResource
+
+
+def run_glc(location, output_file):
+    """
+    Scan `location` content and write results into `output_file`.
+    """
+    classifier = LicenseClassifier()
+    _ = classifier.analyze(location, output=output_file)
+    return
```
tdruez, line 38: `return` is not necessary here

scanpipe/pipes/glc.py
```diff
@@ -0,0 +1,97 @@
...
+                value = scanned_resource.get(field.name, None)
+
+            if value is not None:
+                resource_data[field.name] = value
+
+        resource_type = "FILE" if isfile(scanned_resource["path"]) else "DIRECTORY"
+        resource_data["type"] = CodebaseResource.Type[resource_type]
+        resource_path = get_path(root, scanned_resource["path"], strip_root=True)
+
+        _, flag = CodebaseResource.objects.get_or_create(
```
tdruez, line 93: Since you are not using `_` not `flag` you do not have to create those variables.

tdruez (commented):
@AvishrantsSh That branch looks much better!

Make sure to read https://aboutcode.readthedocs.io/en/latest/contributing/writing_good_commit_messages.html
"GLC Scan Pipeline" is not a great commit message.

AvishrantsSh:
Is this what are you talking about @tdruez? For using the tool on a single file?
https://scancodeio.readthedocs.io/en/latest/scanpipe-tutorial-2.html

tdruez:
@AvishrantsSh I was talking about the ability to run ScanCode-toolkit on the whole codebase (similar to a command line usage) or to call the internal API directly from Python.

See https://github.com/nexB/scancode.io/blob/main/scanpipe/pipes/scancode.py#L143 and https://github.com/nexB/scancode.io/blob/main/scanpipe/pipes/scancode.py#L159 for calling the API on 1 file.
And https://github.com/nexB/scancode.io/blob/main/scanpipe/pipes/scancode.py#L301 for running on a directory.

In any case, it's an example of various ways to run the toolkit, not meant to be duplicated on your side.

For now, it'll be great to have a new pipe function in your new module that take a location of a single file and return the scan results.

AvishrantsSh:
> For now, it'll be great to have a new pipe function in your new module that take a location of a single file and return the scan results.
Right now the module only supports json file as output. I am working on returning the result as a dictionary type object to Python. Once done, I'll do the necessary amendments in the pipes as you said.

tdruez:
@AvishrantsSh I'm having my Python process killed when calling `glc.run_glc` on some directories.

At first, it seems to be random, but after some investigation, it appears that the `Killed` is not directly related to a file specific content but instead happens when the size of the files in that directory is above a certain sum. In my case that threshold is around 18MB but I'm guessing it may vary depending on the machine.
If I remove a file from that directory, the `run_glc` completes fine, but let's add any other file of ~1MB and the process dies.

This looks a lot like a memory related issue.

To reproduce locally:

```
import tempfile
from scanpipe.pipes import glc

glc.run_glc(
    location="<debug directory location>",
    output_file=tempfile.mkstemp()[1],
)
```

AvishrantsSh:
Can you share the memory stats while execution of the program?

AvishrantsSh:
> This looks a lot like a memory related issue.

It is possible because JSON is generated only when all the data is available. So it kind of waits until the scan is complete.
Also can you point me to some resources for testing the module, because I've been able to successfully run it on files as big as 100MB.

tdruez:
> Can you share the memory stats while execution of the program?

It'll be easier if you can reproduce the issue locally so you can debug from there.

> Also can you point me to some resources for testing the module

Pretty much any large input fails on my side. For example: `docker://centos:8`

scanpipe/pipelines/glc_scan.py
AvishrantsSh: Done

scanpipe/pipelines/glc_scan.py
AvishrantsSh: Removed

AvishrantsSh:
> For now, it'll be great to have a new pipe function in your new module that take a location of a single file and return the scan results.

I've implemented a scan_file function in glc pipe. Hope it meets the requirement.

> Pretty much any large input fails on my side.

I've also added a maxRoutine parameter in `analyze` function to reduce the number of concurrent goroutines. By default it is set to 10. Also I've made some changes to the main module to limit memory usage. Still if it doesn't help, try reducing the `maxRoutines` to 1.

tdruez:
@AvishrantsSh same issue with `maxRoutines=10` and `maxRoutines=1`

```
>>> classifier.analyze("codebase/", output=tempfile.mkstemp()[1], maxRoutines=1)
Killed
```

tdruez:
The `glc.scan_file` appears to work fine on most files but also get "Killed" on large one:

```
>>> glc.scan_file("codebase/centos_8.tar")
Killed
```

AvishrantsSh:
I think I've figured out the issue. The module currently tries to read the file in its entirety before analysing it. That being said, it will take me some time to modify the module to make it work with larger packages.

AvishrantsSh:
For now, I think we should work on developing some kind of framework to support newer pipelines. We can work on the module part in parallel.

AvishrantsSh:
@tdruez, I've been looking at the issue and it seems there is not much that I can do. Google Licenseclassifier requires the file to be in byte array format and doesn't support data streams.
```
func (c *Classifier) match(in []byte) Matches {
	id := c.createTargetIndexedDocument(in)
```
So until changes are made in the upstream, it will be difficult.
One thing that we can do is skip the files which are too big to analyze.
Thoughts?

tdruez:
> One thing that we can do is skip the files which are too big to analyze.

Yes, but instead of skipping, it'd be better to catch the error when it occur (so we can log it) and move on the next file.

scanpipe/pipes/glc.py
```diff
@@ -0,0 +1,63 @@
...
+    result = classifier.scan_file(location)
+    return result
+
+
+def update_codebase_resources(project, scan_data):
+    """
+    Update CodebaseResource model with scan results from golicense-classifier
+    """
+    for resource in project.codebaseresources.no_status():
+        scan_result = next(
```
tdruez, line 54: This needs to be simplified. 
Also performance wise, it's not great to have to iterate over the full results set for each resource. 
What about making an index of those results using the path as key instead?

scanpipe/tests/test_pipes.py
```diff
@@ -700,6 +701,43 @@ def test_scanpipe_pipes_rootfs_has_hash_diff(self):
...
+        input_location = str(self.data_location / "notice.NOTICE")
+        scan_results = glc.scan_file_for_license(input_location)
+        expected = [
+            "path",
+            "licenses",
+            "license_expressions",
+            "copyrights",
+            "scan_errors",
+        ]
+        self.assertEqual(sorted(expected), sorted(scan_results.keys()))
```
tdruez, line 714: It could be nice to test not only on the keys but for the full results.

scanpipe/tests/test_pipes.py
```diff
@@ -700,6 +701,43 @@ def test_scanpipe_pipes_rootfs_has_hash_diff(self):
...
+        codebase_resource2 = CodebaseResource.objects.create(
+            project=project1, path="notice.NOTICE"
+        )
+
+        scan_results = glc.scan_file_for_license(codebase_resource2.location)
+        glc.update_codebase_resources(project1, [scan_results])
+
+        codebase_resource2.refresh_from_db()
+        self.assertEqual("scanned", codebase_resource2.status)
+
```
tdruez, line 740: We could add somme extra assertion on the values set on the `codebase_resource2`.
Maybe on `licenses`, `copyrights`, ... everything that have been updated by the update_codebase_resources.

tdruez (commented):
@AvishrantsSh The tests looks good! See my various comments.
Also, you may want to add a unit test for the scan_directory function as well.

scanpipe/tests/test_pipes.py
AvishrantsSh: It was my original plan but licenseclassifier is not able to produce results for `notice.NOTICE`. If we can have some other file in data folder then it might help.

tdruez: @AvishrantsSh You're welcome to add test data that produce relevant results for your tests.

scanpipe/pipes/glc.py
AvishrantsSh: This is something that pertains to the output structure we use for the result. Can you have a look at #231?

pombredanne:
@AvishrantsSh the code we worked out together was fine IMHO https://github.com/nexB/scancode.io/blob/a4a451f809c4e9cabd353c7e036fac8b822e7861/scanpipe/pipelines/glc_scan.py

The two commits you made afterwards make no sense to me.
1. we do not want to scan a whole directory at once for now at all
2. wrt. #231 I think you should not get hung up on the format. Rather try to update a single thing for now which would be the `license_expressions` field and nothing else directly in https://github.com/nexB/scancode.io/blob/a4a451f809c4e9cabd353c7e036fac8b822e7861/scanpipe/pipelines/glc_scan.py#L60 and do notr save all the data you get for now.
  

So IMHO you should revert https://github.com/nexB/scancode.io/pull/226/commits/9041af6ebc25aee12f2ed5a459d1dec1cce8759f entirely
https://github.com/nexB/scancode.io/pull/226/files#diff-0297e88d813e7ee05472fe75fd201ca5277f7e8f71085456558416c43230b39fR49 does not make sense at if you scan 10,000 files you would get 100,000,000 loops done.

Please do not try to work in batch on a full directory for now. Focus on correctness, one file at a time. And not on any optimization.

scanpipe/pipes/glc.py
pombredanne: Please see https://github.com/nexB/scancode.io/pull/226#issuecomment-879912975 
This nested loop is not needed and is not related to the format documentation of #231 . If you work one file ta a time that's a non-issue.

AvishrantsSh:
@pombredanne, I have reverted the changes as you mentioned. Now the pipeline scans individual files and saves them in a linear fashion, just as we designed the other day. The pipeline is now able to store all the parameters in scancode.io, including copyright info and holders.

Also, I have made some modifications in the module itself to only scan files of size 50MB or less for now (@tdruez mentioned high memory usage with some files). Anything above that, a `scan_error` will be generated.

AvishrantsSh:
@tdruez, I have updated the tests. Can you have a look?

scanpipe/pipes/glc.py
```diff
@@ -0,0 +1,81 @@
...
+    ----------
+    location : str
+        Location of directory to scan.
+    """
+    classifier = LicenseClassifier()
+    result = classifier.scan_directory(location)
+    return result
+
+
+def scan_file(classifier, location) -> dict:
```
tdruez, line 44: We are not using type hints in the ScanCode.io codebase. In general, you want to match the style of the codebase you are contributing to, for consistency.

scanpipe/pipes/glc.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    Run Golicense-Classifier on `project` and save results in CodebaseResources model
+
+    Parameter
+    ----------
+    project : Project()
+            Instance of `Project` to scan
+
+    scan_threshold : float
+            Threshold for license scan results. `0 < scan_threshold <= 1.0`
```
tdruez, line 71: Could you add details about the consequences of modifying that value?

scanpipe/tests/test_pipes.py
```diff
@@ -700,6 +702,76 @@ def test_scanpipe_pipes_rootfs_has_hash_diff(self):
...
+        scan_results = glc.scan_file(classifier, input_location)
+        expected_keys = [
+            "path",
+            "licenses",
+            "license_expressions",
+            "copyrights",
+            "holders",
+            "scan_errors",
+        ]
+        self.assertEqual(sorted(expected_keys), sorted(scan_results.keys()))
```
tdruez, line 717: It'll be nice to test for the full results instead of just the keys.

tdruez (commented):
@AvishrantsSh The tests are fine, see my comments for improvements.

tdruez:
> Also, I have made some modifications in the module itself to only scan files of size 50MB or less for now (@tdruez mentioned high memory usage with some files). Anything above that, a scan_error will be generated.

Pipelines are now able to complete without crashing, much better :)

scanpipe/pipes/glc.py
AvishrantsSh: I used type hints while documenting the module. But I removed the same from the recent commit.

scanpipe/pipes/glc.py
AvishrantsSh: Done :)

tdruez:
Closing the PR, the code is now available at https://github.com/nexB/scancodeio.io-pipeline-glc_scan

Issue #191, opened 2021-06-08
Enhance Alpine package scan results


aalexanderr:
Alpine packages lack some important info like copyrights or where the source code is located. This info can't be gathered from the packages themselves as its just not there. To get this info we need to:
download aports repo & for each pkg check it out on commit specific to alpine package (via fetchcode)
parse APKBUILD nexB/scancode-toolkit#2541
download package sources (fetchcode) & amend new info to package's scan results

Discussed a bit with @pombredanne
Most likely @quepop will PR it

The question is- should it be standard behavior when alpine based docker is being scanned or should it be a separate pipeline?

quepop:
Related: https://github.com/nexB/scancode-toolkit/issues/2541#issuecomment-856904754

quepop:
@pombredanne should I PR it?

pombredanne:
@quepop please do! that's much easier for review
Of note:
- I am not super comfy (that's an understatement) with sourcing arbitrary shell scripts at https://github.com/quepop/scancode.io/commit/2f59440b97cfc89e4bc8291d7839188d25e0fb34#diff-a6e238494a11c38a54d401da6211e8e3461898f10e22ff6673a7896ffedbbd3eR43
- I am not sure we want the alpine package getter to always fetch packages. If anything I know several users run in an airgap'ed environment and they surely will not want (and will not be able anyway) to fetch things from the outside world

quepop:
@pombredanne I'dont think there is any other way. In order to extract source variable value (or any other value) we need to run the script (APKBUILD file), because in some of the alpine packages, source variable is built in a for loop (not to mention that sometimes bash string formatting is used)

From `pkg:alpine/alpine-keys@2.2-r0` APKBUILD file:
`_arch_keys="
        aarch64:alpine-devel@lists.alpinelinux.org-58199dcc.rsa.pub
        armhf:alpine-devel@lists.alpinelinux.org-524d27bb.rsa.pub
 
        x86:alpine-devel@lists.alpinelinux.org-5243ef4b.rsa.pub
        x86,x86_64:alpine-devel@lists.alpinelinux.org-4a6a0840.rsa.pub
        x86_64:alpine-devel@lists.alpinelinux.org-5261cecb.rsa.pub
 
        ppc64le:alpine-devel@lists.alpinelinux.org-58cbb476.rsa.pub
 
        s390x:alpine-devel@lists.alpinelinux.org-58e4f17d.rsa.pub
 
[...]

PR #272, opened 2021-08-05
Add AlpinePackages pipeline

quepop:
Fixes: #191
Depends on: https://github.com/nexB/fetchcode/pull/54, https://github.com/nexB/fetchcode/pull/56, https://github.com/nexB/scancode-toolkit/pull/2598

When developing I ran into some issues that I couldn't fix on my own so I've decided to list them here and mark this PR as a draft:
- The function `fetch_via_git` from https://github.com/nexB/fetchcode/pull/54 doesn't provide any successful/failed checkout feedback which is essential to this PR.
- I'm not confident with how I split the entire commit into functions. Especially `complement_missing_packages_data`. It feels too big for a pipeline function (compared to other pipelines). I tried to be as clean as possible but at this point I'm out of ideas.
- I'm not sure about file headers/copyrights.
- Do package copyrights require a year (in a legal sense)? This commit uses `--summary` scancode option which is very convenient but it also prunes any year from the original copyright message.

Signed-off-by: Mateusz Perc <m.perc@samsung.com>

### DIFF
--- a/scanpipe/pipelines/alpine_packages.py
+++ b/scanpipe/pipelines/alpine_packages.py
@@ -1,0 +1,107 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# http://nexb.com and https://github.com/nexB/scancode.io
+# The ScanCode.io software is licensed under the Apache License version 2.0.
+# Data generated with ScanCode.io is provided as-is without warranties.
+# ScanCode is a trademark of nexB Inc.
+#
+# You may not use this software except in compliance with the License.
+# You may obtain a copy of the License at: http://apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software distributed
+# under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations under the License.
+#
+# Data Generated with ScanCode.io is provided on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, either express or implied. No content created from
+# ScanCode.io should be considered or used as legal advice. Consult an Attorney
+# for any legal advice.
+#
+# ScanCode.io is a free software code scanning tool from nexB Inc. and others.
+# Visit https://github.com/nexB/scancode.io for support and download.
+
+from scanpipe.pipelines import Pipeline
+from scanpipe.pipes.alpine import download_or_checkout_aports
+from scanpipe.pipes.alpine import extract_summary_fields
+from scanpipe.pipes.alpine import get_unscanned_packages_from_db
+from scanpipe.pipes.alpine import prepare_scan_dir
+from scanpipe.pipes.scancode import run_extractcode
+from scanpipe.pipes.scancode import run_scancode
+
+
+class AlpinePackages(Pipeline):
+    """
+    A pipeline to complement missing alpine package data.
+    Downloads and extracts needed information from aports repository and package source files.
+    Alpine Linux does not provide copyrights and (in some cases) licenses for it's packages.
+    """
+
+    @classmethod
+    def steps(cls):
+        return (
+            cls.create_alpine_versions_dict,
+            cls.download_aports_repo,
+            cls.complement_missing_package_data,
+        )
+
+    scancode_options = ["--copyright", "--summary"]
+
+    def create_alpine_versions_dict(self):
+        """
+        Create a dict mapping alpine image ids from the database to alpine versions.
+        """
+        self.alpine_versions = {
+            i["image_id"]: i["distro"]["version_id"]
+            for i in self.project.extra_data["images"]
+            if i["distro"]["identifier"] == "alpine"
+        }
+
+    def download_aports_repo(self):
+        """
+        Set pipeline's `aports_dir_path` variable to it's project temporary path.
+        Iterate over every alpine version associated with this project.
+        Download corresponding aports repository branches (alpine versions).
+        """
+        self.aports_dir_path = self.project.tmp_path
+        for image_id, alpine_version in self.alpine_versions.items():
+            download_or_checkout_aports(
+                aports_dir_path=self.project.tmp_path, alpine_version=alpine_version
+            )
+
+    def complement_missing_package_data(self):
+        """
+        Iterate over alpine packages associated with this project.
+        Checkout aports repository to the corresponding alpine version and a commit.
+        Prepare scan target directory - download and extract package's sources.
+        Run scancode and extract missing data (only copyrights for now).
+        Update and save package's missing data to database.
+        """
+        for (
+            alpine_version,
+            commit_id,
+            scan_target_path,
+            scan_result_path,
+            package,
+        ) in get_unscanned_packages_from_db(
+            project=self.project, alpine_versions=self.alpine_versions
+        ):
+            if not download_or_checkout_aports(
+                aports_dir_path=self.aports_dir_path,
+                alpine_version=alpine_version,
+                commit_id=commit_id,
+            ) or not prepare_scan_dir(
+                package_name=package.name, scan_target_path=scan_target_path
+            ):
+                continue
+            run_extractcode(location=str(scan_target_path))
+            run_scancode(
+                location=str(scan_target_path),
+                output_file=str(scan_result_path),
+                options=self.scancode_options,
+            )
+            package.update_extra_data(
+                data=extract_summary_fields(
+                    scan_result_path=scan_result_path,
+                    summary_field_names=["copyrights"],
+                )
+            )
--- a/scanpipe/pipes/alpine.py
+++ b/scanpipe/pipes/alpine.py
@@ -20,7 +20,113 @@
 # ScanCode.io is a free software code scanning tool from nexB Inc. and others.
 # Visit https://github.com/nexB/scancode.io for support and download.
 
+
+import json
+from shutil import copytree
+
+from fetchcode import fetch
+from fetchcode.vcs.git import fetch_via_git
 from packagedcode import alpine
+
+from scanpipe.models import DiscoveredPackage
+
+APORTS_URL = "https://gitlab.alpinelinux.org/alpine/aports.git"
+APORTS_DIR_NAME = "aports"
+APORTS_SUBDIRS = ["main", "non-free", "testing", "community", "unmaintained"]
+
+
+def download_or_checkout_aports(aports_dir_path, alpine_version, commit_id=None):
+    """
+    Download aports repository and it's branch based on `alpine_version`.
+    Checkout to a branch (alpine version).
+    If `commit_id` is provided also checkout to a commit.
+    Return `aports_dir_path` if checkout(s) succeded. #TODO Proper fetchcode patch required (extending #54)
+    """
+    major, minor = alpine_version.split(".")[:2]
+    aports_dir_path = str(aports_dir_path / APORTS_DIR_NAME)
+    fetch_via_git(
+        url=f"git+{APORTS_URL}@{major}.{minor}-stable", location=aports_dir_path
+    )
+    if commit_id:
+        fetch_via_git(url=f"git+{APORTS_URL}@{commit_id}", location=aports_dir_path)
+    return aports_dir_path
+
+
+def get_unscanned_packages_from_db(project, alpine_versions):
+    """
+    Return an iterator of 5-tuples (alpine_version, commit_id, scan_target_path, scan_result_path, package) where:
+    `alpine_version` is an alpine version from which a package comes from (obtained from `alpine_versions` dict),
+    `commit_id` is an id of aports repository commit that added corresponding version of a package,
+    `scan_target_path` is a path of the directory on which a scan will be performed,
+    `scan_result_path` is a path of the scan result json file,
+    `package` is a DiscoveredPackage instance that belongs to a `project` with an alpine package type.
+    The returned iterator contains not-a-subpackage alpine packages that don't have an existing scan result file.
+    """
+    for package in DiscoveredPackage.objects.filter(project=project, type="alpine"):
+        scan_id = f"{package.name}_{package.version}"
+        scan_result_path = project.output_path / (scan_id + ".json")
+        alpine_version = alpine_versions.get(package.extra_data["image_id"])
+        commit_id = package.vcs_url.split("id=")[1]
+        scan_target_path = project.tmp_path / scan_id
+        not_a_subpackage = (
+            not package.source_packages or package.source_packages[0] in package.purl
+        )
+        scan_result_nonexistent = not scan_result_path.exists()
+        if not_a_subpackage and scan_result_nonexistent:
+            yield alpine_version, commit_id, scan_target_path, scan_result_path, package
+
+
+def prepare_scan_dir(package_name, scan_target_path, aports_dir_path=None):
+    """
+    A function to gather all the package's source files in `scan_target_path`.
+    Source files of an alpine package are obtained from it's aports directory whose location has to be guessed.
+    Such directory is present in one of the five aports repository subdirectories (main, non-free, testing, community, unmaintained).
+    It's name is the same as the value of the corresponding package's `name` field (hence the `package_name` parameter).
+    Here are some path examples:
+    .../aports/main/acf-db
+    .../aports/non-free/mongodb
+    Inside, there are some extra files (patches) and an APKBUILD which contains urls to source tarballs.
+    The function copies all these files (including APKBUILD) and downloads all the source tarballs to `scan_target_path`.
+    The default value of `aports_dir_path` is set to the parent of the `scan_target_path`.
+    If the package's aports path is found/guessed and it's also not empty the returned value is `scan_target_path`.
+    """
+    if aports_dir_path is None:
+        aports_dir_path = scan_target_path.parent
+    for subdir_name in APORTS_SUBDIRS:
+        apkbuild_dir = aports_dir_path / APORTS_DIR_NAME / subdir_name / package_name
+        if not apkbuild_dir.exists():
+            continue
+        if not any(apkbuild_dir.iterdir()):
+            break
+        copytree(apkbuild_dir, scan_target_path)
+        package_sources = (
+            alpine.parse_apkbuild(scan_target_path / "APKBUILD")
+            .to_dict()
+            .get("extra_data")
+            .get("sources")
+            or []
+        )
+        for source in package_sources:
+            source_url = source.get("url")
+            if source_url:
+                fetch(source_url, scan_target_path)
+        return scan_target_path
+
+
+def extract_summary_fields(scan_result_path, summary_field_names):
+    """
+    Having a scancode result file extract all the values from the `summary` section of the scan result file (`scan_result_path`).
+    Put them in the arrays inside the `result` object (result[`field_name`]).
+    Return `result`.
+    """
+    scan_result = open(scan_result_path)
+    summaries = json.load(scan_result)["summary"]
+    scan_result.close()
+    result = {}
+    for field_name in summary_field_names:
+        values = (summary["value"] for summary in summaries.get(field_name, []))
+        result[field_name] = [v for v in values if v]
+    return result
 
 
 def package_getter(root_dir, **kwargs):
--- a/scanpipe/pipes/docker.py
+++ b/scanpipe/pipes/docker.py
@@ -122,6 +122,7 @@ def scan_image_for_system_packages(project, image, detect_licenses=True):
 
     for i, (purl, package, layer) in enumerate(installed_packages):
         logger.info(f"Creating package #{i}: {purl}")
+        package.extra_data = {"image_id": image.image_id}
         created_package = pipes.update_or_create_package(project, package.to_dict())
 
         # We have no files for this installed package, we cannot go further.
--- a/scanpipe/tests/data/example_scan_summary.json
+++ b/scanpipe/tests/data/example_scan_summary.json
@@ -1,0 +1,34 @@
+{
+  "summary": {
+    "copyrights": [
+      {
+        "value": "Copyright (c) A B",
+        "count": 51
+      },
+      {
+        "value": "Copyright (c) C D",
+        "count": 8
+      }
+    ],
+    "holders": [
+      {
+        "value": "A B",
+        "count": 51
+      },
+      {
+        "value": "C D",
+        "count": 41
+      }
+    ],
+    "authors": [
+      {
+        "value": "A B",
+        "count": 2
+      },
+      {
+        "value": "C D",
+        "count": 1
+      }
+    ]
+  }
+}
--- a/scanpipe/tests/test_pipes.py
+++ b/scanpipe/tests/test_pipes.py
@@ -38,6 +38,7 @@ from scancode.interrupt import TimeoutError as InterruptTimeoutError
 from scanpipe.models import CodebaseResource
 from scanpipe.models import DiscoveredPackage
 from scanpipe.models import Project
+from scanpipe.pipes import alpine
 from scanpipe.pipes import codebase
 from scanpipe.pipes import docker
 from scanpipe.pipes import fetch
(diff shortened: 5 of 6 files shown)


scanpipe/pipes/alpine.py
```diff
@@ -20,7 +20,91 @@
...
+
+
+def download_or_checkout_aports(aports_dir_path, alpine_version, commit_id=None):
+    """
+    Download aports repository and it's branch based on `alpine_version`.
+    Checkout to a branch (alpine version).
+    If `commit_id` is provided also checkout to a commit.
+    Return False if checkout failed otherwise True. #TODO Proper fetchcode patch required (extending #54)
+    """
+    ver = alpine_version.split(".")
```
pombredanne, line 43: What about using a more stable `.partition(".")` instead? or assigning a variable to each of the version segments [0] and [1] for clarity and readability by giving them a name?

scanpipe/pipes/alpine.py
```diff
@@ -20,7 +20,91 @@
...
+    Yield only root and not yet scanned packages along with paths needed when performing a scan.
+    """
+    for package in DiscoveredPackage.objects.project(project):
+        scan_id = f"{package.name}_{package.version}"
+        scan_target_path = project.tmp_path / scan_id
+        scan_result_path = project.output_path / (scan_id + ".json")
+        if (
+            package.type == "alpine"
+            or (package.source_packages and package.source_packages[0] in package.purl)
+            or not scan_result_path.exists()
```
aalexanderr, line 63: Seems to me that among others it would yield non-alpine packages that have no scan results?

scanpipe/pipes/alpine.py
```diff
@@ -20,7 +20,91 @@
...
+def download_or_checkout_aports(aports_dir_path, alpine_version, commit_id=None):
+    """
+    Download aports repository and it's branch based on `alpine_version`.
+    Checkout to a branch (alpine version).
+    If `commit_id` is provided also checkout to a commit.
+    Return False if checkout failed otherwise True. #TODO Proper fetchcode patch required (extending #54)
+    """
+    ver = alpine_version.split(".")
+    aports_dir_path = str(aports_dir_path / APORTS_DIR_NAME)
+    fetch_via_git(f"git+{APORTS_URL}@{ver[0]}.{ver[1]}-stable", aports_dir_path)
```
pombredanne, line 45: Here and in general, do you mind to use named keyword arguments rather than un-named positional arguments? This makes reading much easier and is more resistant to refactorings that adds or reorders arguments

scanpipe/pipes/alpine.py
```diff
@@ -20,7 +20,91 @@
...
+    """
+    Get alpine packages from the database that belong to a `project`
+    Yield only root and not yet scanned packages along with paths needed when performing a scan.
+    """
+    for package in DiscoveredPackage.objects.project(project):
+        scan_id = f"{package.name}_{package.version}"
+        scan_target_path = project.tmp_path / scan_id
+        scan_result_path = project.output_path / (scan_id + ".json")
+        if (
+            package.type == "alpine"
```
pombredanne, line 61: This should be a filter in the query set above (e.g. `DiscoveredPackage.objects.project(project)`)

scanpipe/pipes/alpine.py
```diff
@@ -20,7 +20,91 @@
...
+            or (package.source_packages and package.source_packages[0] in package.purl)
```
pombredanne, line 62: What does `package.source_packages[0] in package.purl)` mean? it would help to have that in a named variable for readability.

scanpipe/pipes/alpine.py
```diff
@@ -20,7 +20,91 @@
...
+        Download all package's sources into `scan_target_path`
+    The default value of `aports_dir_path` is set to the parent of the `scan_target_path`
+    """
+    if aports_dir_path is None:
+        aports_dir_path = scan_target_path.parent
+    for subdir_name in APORTS_SUBDIRS:
+        apkbuild_dir = aports_dir_path / APORTS_DIR_NAME / subdir_name / package_name
+        if not apkbuild_dir.exists():
+            continue
+        copytree(apkbuild_dir, scan_target_path)
```
pombredanne, line 81: Why do you need a copy?

scanpipe/pipes/alpine.py
```diff
@@ -20,7 +20,91 @@
...
+def prepare_scan_dir(package_name, scan_target_path, aports_dir_path=None):
+    """
+    Find package's aports path and if found execute the following steps:
+        Copy all the files from that path into `scan_target_path`
+        Download all package's sources into `scan_target_path`
+    The default value of `aports_dir_path` is set to the parent of the `scan_target_path`
+    """
+    if aports_dir_path is None:
+        aports_dir_path = scan_target_path.parent
+    for subdir_name in APORTS_SUBDIRS:
```
pombredanne, line 77: Rather than trying all the paths and since each package name should be unique within a given release what about a different logic:
1. walk the aports tree once and build a mapping of `{<package name>: <package path> }` (or maybe even better <APKBUILD path>` )
2. use this in your loop below

scanpipe/pipes/alpine.py
```diff
@@ -20,7 +20,91 @@
...
+        for source in (
+            alpine.parse_apkbuild(scan_target_path / "APKBUILD")
+            .to_dict()
+            .get("extra_data")
+            .get("sources")
+            or []
+        ):
+            if source["url"]:
+                fetch(source["url"], scan_target_path)
+        return True
```
pombredanne, line 91: Why do you return True and what does this mean? this is not documented anywhere

scanpipe/pipes/alpine.py
```diff
@@ -20,7 +20,91 @@
...
+            continue
+        copytree(apkbuild_dir, scan_target_path)
+        for source in (
+            alpine.parse_apkbuild(scan_target_path / "APKBUILD")
+            .to_dict()
+            .get("extra_data")
+            .get("sources")
+            or []
+        ):
+            if source["url"]:
```
pombredanne, line 89: This would be more robust
```suggestion
            source_url = source.get("url")
            if source_url:
```

scanpipe/pipes/alpine.py
```diff
@@ -20,7 +20,91 @@
...
+    """
+    if aports_dir_path is None:
+        aports_dir_path = scan_target_path.parent
+    for subdir_name in APORTS_SUBDIRS:
+        apkbuild_dir = aports_dir_path / APORTS_DIR_NAME / subdir_name / package_name
+        if not apkbuild_dir.exists():
+            continue
+        copytree(apkbuild_dir, scan_target_path)
+        for source in (
+            alpine.parse_apkbuild(scan_target_path / "APKBUILD")
```
pombredanne, line 83: Create a variable with the results of this rather than chaining many calls in one place. Do this before the for loop, this will make the code more readable.

scanpipe/pipes/alpine.py
```diff
@@ -20,7 +20,91 @@
...
+        return True
+
+
+def extract_summary_fields(scan_result_path, summary_fields):
+    """
+    Having a scancode result file extract all the 'summary_fields' values from the `summary` section.
+    Return an object mapping 'summary_fields' to the arrays of values mentioned above.
+    """
+    with open(scan_result_path) as scan_result:
+        json_obj = json.load(scan_result)
```
pombredanne, line 100: what does the `_obj` suffix means and brings here and below?

scanpipe/pipes/alpine.py
```diff
@@ -20,7 +20,91 @@
...
+            or []
+        ):
+            if source["url"]:
+                fetch(source["url"], scan_target_path)
+        return True
+
+
+def extract_summary_fields(scan_result_path, summary_fields):
+    """
+    Having a scancode result file extract all the 'summary_fields' values from the `summary` section.
```
pombredanne, line 96: Are you somehow dependent on the summary features of scancode-toolkit?

scanpipe/pipes/alpine.py
```diff
@@ -20,7 +20,91 @@
 # ScanCode.io is a free software code scanning tool from nexB Inc. and others.
 # Visit https://github.com/nexB/scancode.io for support and download.
 
+import json
+
 from packagedcode import alpine
+from fetchcode.vcs.git import fetch_via_git
```
pombredanne, line 26: Please sort your imports

pombredanne (changes requested):
Thank you ++ for this! I made a few comments for your review.
Also we would need some automated tests for sure. 
Can we setup a time to discuss all these in a live session?

scanpipe/pipes/alpine.py
```diff
@@ -20,7 +20,91 @@
...
+            package.type == "alpine"
+            or (package.source_packages and package.source_packages[0] in package.purl)
+            or not scan_result_path.exists()
+        ):
+            yield scan_target_path, scan_result_path, package
+
+
+def prepare_scan_dir(package_name, scan_target_path, aports_dir_path=None):
+    """
+    Find package's aports path and if found execute the following steps:
```
aalexanderr, line 70: If it is not found shouldn't it be indicated somehow to provide input for further investigation?

scanpipe/pipes/alpine.py
```diff
@@ -20,7 +20,91 @@
...
+        ):
+            if source["url"]:
+                fetch(source["url"], scan_target_path)
+        return True
+
+
+def extract_summary_fields(scan_result_path, summary_fields):
+    """
+    Having a scancode result file extract all the 'summary_fields' values from the `summary` section.
+    Return an object mapping 'summary_fields' to the arrays of values mentioned above.
```
aalexanderr, line 97: above where?

scanpipe/pipes/alpine.py
aalexanderr: I think current solution might be quicker (as we're not iterating over all packages in alpine - just checking for existance of pkgs we're interested in. the APORTS_SUBDIRS is sorted by probability of having package we're looking for (main->non-free->contrib...)

scanpipe/pipes/alpine.py
quepop: Because we have to download and extract the sources somwhere  them and it shouldnt be inside of the aports repo directory. Furthermore doing a scan on a single directory is much more readable and elegant.

scanpipe/pipes/alpine.py
quepop: Line 96

scanpipe/pipes/alpine.py
quepop: JSON Object Literal

scanpipe/pipes/alpine.py
quepop: Yes

scanpipe/pipes/alpine.py
quepop: It indicates to the Pipeline function if preparing the scan directory succeeded. It is essential because why proceed with a scan if you for e.g. didn't even found package's APKBUILD? Right now it only fails in that case and i plan to add returning false when the scan directory is empty.

scanpipe/pipes/alpine.py
quepop: My bad, it should be `and` instead of `or`.

scanpipe/pipelines/alpine_packages.py
```diff
@@ -0,0 +1,95 @@
...
+        Checkout aports repository to the corresponding alpine version and commit.
+        Prepare scan target directory, download and extract package's sources.
+        Run scancode and extract missing data (only copyrights for now).
+        Update and save package's missing data to database.
+        """
+        for scan_target_path, scan_result_path, package in get_packages_from_db(
+            self.project
+        ):
+            if (
+                not download_or_checkout_aports(
```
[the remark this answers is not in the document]
quepop: Because checking out to a package's alpine version and commit id (provided by the package manager) gives us the best probability of finding correct APKBUILD (for a corresponding package version) and correct patch files, making our scans much more accurate. Also, package versions can change from commit to commit within alpine version branch.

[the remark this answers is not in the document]
quepop: > Can we setup a time to discuss all these in a live session?

Yeah, it would be great.

quepop:
> Can we setup a time to discuss all these in a live session?

Yeah, it would be great.

scanpipe/pipes/alpine.py
```diff
@@ -20,7 +20,91 @@
...
+    """
+    with open(scan_result_path) as scan_result:
+        json_obj = json.load(scan_result)
+        result_obj = {}
+        for field in summary_fields:
```
pombredanne, line 102: The whole code block could be rewritten this way:
```Python
    summaries = scan_result["summary"]
    for field_name in summary_field_names:
        values = (summary["value"] for summary in summaries.get(field_name, []))
        result[field_name] = [v for v in values if v]
```

scanpipe/pipes/alpine.py
```diff
@@ -20,7 +20,91 @@
...
+            .get("extra_data")
+            .get("sources")
+            or []
+        ):
+            if source["url"]:
+                fetch(source["url"], scan_target_path)
+        return True
+
+
+def extract_summary_fields(scan_result_path, summary_fields):
```
pombredanne, line 94: Is there a default tuple of `summary_field_names` that you could define here? This would also help understanding the code better.

scanpipe/pipes/docker.py
```diff
@@ -123,6 +123,7 @@ def scan_image_for_system_packages(project, image, detect_licenses=True):
     for i, (purl, package, layer) in enumerate(installed_packages):
         logger.info(f"Creating package #{i}: {purl}")
         created_package = pipes.update_or_create_package(project, package.to_dict())
+        created_package.update_extra_data({"image_id": image.image_id})
```
pombredanne, line 126: Can you elaborate on why you need to keep this?

scanpipe/pipes/alpine.py
pombredanne: OK, so if I understand correctly you are:
1. making a copy of the aports directory of a given package (which would typically include the APKBUILD and some patches)
2. in this copied directory, you will also fetch the sources (or at least only the remote sources as identified by a URL)
3. finally you will (extract then run a scan of sorts on this directory? )

I think it could be better if you separate each operation and the process could benefit from more documentation.

scanpipe/pipes/alpine.py
```diff
@@ -20,7 +20,91 @@
...
+    """
+    ver = alpine_version.split(".")
+    aports_dir_path = str(aports_dir_path / APORTS_DIR_NAME)
+    fetch_via_git(f"git+{APORTS_URL}@{ver[0]}.{ver[1]}-stable", aports_dir_path)
+    if commit_id:
+        fetch_via_git(f"git+{APORTS_URL}@{commit_id}", aports_dir_path)
+    return True
+
+
+def get_packages_from_db(project):
```
pombredanne, line 51: Is the fact you get the packages from the db the most interesting thing this function does? or maybe it is rather the fact that it gets not-yet-scanned-for-source Alpine packages instead? So IMHO we can find a better name.

scanpipe/pipes/alpine.py
```diff
@@ -20,7 +20,91 @@
...
+    """
+    Get alpine packages from the database that belong to a `project`
```
pombredanne, line 53: I think that this docstring could be improved: You are yielding three-tuples of (scan_target_path, scan_result_path, package)... so say it as in 

> Yield three-tuples of (scan_target_path, scan_result_path, package) where  scan_target_path is xxx, scan_result_path is yyy and package  is a DiscoveredPackage instance with an ``alpine``  package type,  given a Project ``project``.

> Do not return DiscoveredPackage for which we have already fetched and scanned its corresponding source package source code.

pombredanne (changes requested):
Thank for the quicke replies!
I added some extra comments for your review.

scanpipe/pipes/docker.py
tdruez: Also, in that shape, this will trigger 1 extra SQL query per packages.

quepop: The pipeline needs to know package's alpine version (It is stored in `project.extra_data["images"][...]["distro"]["version_id"]`).

scanpipe/pipelines/alpine_packages.py
```diff
@@ -0,0 +1,95 @@
...
+                    self.project.tmp_path,
+                    self.alpine_versions[package.extra_data["image_id"]],
+                    package.vcs_url.split("id=")[1],
```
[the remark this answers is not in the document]
quepop: I wrote it this way because I don't want to run `prepare_scan_directory` if `download_or_checkout_aports` fails and in cases like these (`A or B`) boolean expression parsing stops  at A if A is True.  If i wanted to retain this behavior and write it your way it would look something like this:
```
checkout_succeded = download_or_checkout_aports(...)
if checkout_succeded:                                                 
   continue                                                                  `
scan_directory_ready = prepare_scan_directory(...)
if scan_directory_ready:
   continue
```
which in my opinion is worse. Also, if i split prepare_scan_directory into smaller functions it would be even longer.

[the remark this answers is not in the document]
quepop: You mean `package.vcs_url.split("id=")[1]` or `not download_or_checkout_aports(...)`?

scanpipe/pipelines/alpine_packages.py
```diff
@@ -0,0 +1,95 @@
...
+    scancode_options = ["--copyright", "--summary"]
+
+    def create_alpine_versions_dict(self):
+        """
+        Create a dict, mapping alpine image ids from the database to alpine versions.
+        """
+        self.alpine_versions = {
+            i["image_id"]: i["distro"]["version_id"]
+            for i in self.project.extra_data["images"]
+            if i["distro"]["identifier"] == "alpine"
```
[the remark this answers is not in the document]
quepop: If these fields are not available should the function do nothing or maybe throw an exeption?

scanpipe/pipes/alpine.py
quepop: It checks if the package is not a sub-package (I mentioned it in the doc string `Yield only root...`). Looking at some alpine docker scans i figured out that the only diffrence between `package.purl` and `package.source_packages[...]`  is that the former has `arch...` at the end of it. So, if a package is not a sub-package it will have `package.purl` minus `arch...` as the first element of `package.source_packages[...]`.

scanpipe/pipes/alpine.py
quepop: > OK, so if I understand correctly you are:

Yes, exactly.

> I think it could be better if you separate each operation and the process could benefit from more documentation.

If i separated 1. from 2. it would have only `copytree(apkbuild_dir, scan_target_path)` as the function body.

scanpipe/pipes/alpine.py
quepop: > Is there a default tuple of summary_field_names that you could define here?

We don't know what parts of the 'summary' we will use in the end (copyrights are the only certain thing). Also, what is the priority here, inside `AlpinePackages` runs or the outside ones? There are certainly no default values that satisfy both.

scanpipe/pipes/alpine.py
quepop: I wanted function names to be as short as possible. I that case i thought `alpine.get_packages_from_db` would be enough.

scanpipe/pipes/alpine.py
quepop: @pombredanne That was the question i forgot to ask in the PR message. What about logging and error handling?

scanpipe/pipes/alpine.py
quepop: So adding more explanations in doc strings will do?

scanpipe/pipes/alpine.py
quepop: I think split is better here because we can't be sure that it will always be in  `major.minor` format.

scanpipe/pipes/alpine.py
quepop: It could be indicated by an empty scan result file.

quepop: How do you think it should be handled? Simple stdout logging?

scanpipe/pipes/alpine.py
quepop: So do you want me to split it? Given my reasons above I don't think we should.

scanpipe/pipelines/alpine_packages.py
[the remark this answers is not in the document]
quepop: We discussed this on our meeting.  We came to a conclusion that these fields are guaranteed by scancode.io so I'm marking this as resolved.

quepop:
@pombredanne @aalexanderr I cannot decide on how to test the entire pipeline. It looks like alpine pipe tests which i commited already test the majority of the AlpinePackages pipeline. Integration tests are the only viable option in my opinion but they are slow and problematic (due to using fetchcode etc.). Either way they won't be big, so tell me what you think and i will be quick to write them.

scanpipe/pipelines/alpine_packages.py
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+    def download_aports_repo(self):
+        """
+        Set pipeline's `aports_dir_path` variable to it's project temporary path.
+        Iterate over every alpine version associated with this project.
+        Download corresponding aports repository branches (alpine versions).
+        """
+        self.aports_dir_path = self.project.tmp_path
```
tdruez, line 65: Is the `self.aports_dir_path` variable really needed?

scanpipe/pipelines/alpine_packages.py
```diff
@@ -0,0 +1,107 @@
...
+        for image_id, alpine_version in self.alpine_versions.items():
```
tdruez, line 66: Since `image_id` is not used, I would suggest:
`for alpine_version in self.alpine_versions.values()`

scanpipe/pipelines/alpine_packages.py
```diff
@@ -0,0 +1,107 @@
...
+            download_or_checkout_aports(
+                aports_dir_path=self.project.tmp_path, alpine_version=alpine_version
+            )
+
+    def complement_missing_package_data(self):
```
tdruez, line 71: The following code should be made more digest and readable.

scanpipe/pipelines/alpine_packages.py
```diff
@@ -0,0 +1,107 @@
...
+        """
+        for (
+            alpine_version,
+            commit_id,
+            scan_target_path,
+            scan_result_path,
+            package,
+        ) in get_unscanned_packages_from_db(
+            project=self.project, alpine_versions=self.alpine_versions
+        ):
```
tdruez, line 87: In general, when the name of the keyword argument and the provided variable is the same, it's explicit enough to only keep the variable.

For example:

```
get_unscanned_packages_from_db(project=self.project, alpine_versions=self.alpine_versions)
```

I think the following is as explicit and more readable:
```
get_unscanned_packages_from_db(self.project, self.alpine_versions)
```

It make sense to keep the keyword agrs in the following example though:
```
run_scancode(
    location=str(scan_target_path),
    output_file=str(scan_result_path),
    options=self.scancode_options,
)
```

scanpipe/pipelines/alpine_packages.py
```diff
@@ -0,0 +1,107 @@
...
+            if not download_or_checkout_aports(
+                aports_dir_path=self.aports_dir_path,
+                alpine_version=alpine_version,
+                commit_id=commit_id,
+            ) or not prepare_scan_dir(
+                package_name=package.name, scan_target_path=scan_target_path
+            ):
+                continue
+            run_extractcode(location=str(scan_target_path))
```
tdruez, line 96: This `run_extractcode` function does not exists in the main branch anymore since https://github.com/nexB/scancode.io/commit/b035f00b60f72a15d5aba49f3f3b03ef192d252c. You need to migrate to the new `scancode.extract_archives` API.
See https://github.com/nexB/scancode.io/blob/main/scanpipe/pipelines/scan_codebase.py#L73 for an example.

scanpipe/pipelines/alpine_packages.py
```diff
@@ -0,0 +1,107 @@
...
+            run_scancode(
```
tdruez, line 97: I would suggest to call directly the ScanCode `scancode.api.get_copyrights` function instead of starting a full scancode subprocess. 
This will be more efficient and will remove the need for `extract_summary_fields`.

tdruez (changes requested):
> @quepop  Integration tests are the only viable option in my opinion but they are slow and problematic (due to using fetchcode etc.)

What about mocking the fetchcode calls?

scanpipe/pipelines/alpine_packages.py
quepop: I used unnamed positional arguments before and @pombredanne commented that i should use named positionals everywhere.
> Here and in general, do you mind to use named keyword arguments rather than un-named positional arguments? This makes reading much easier and is more resistant to refactorings that adds or reorders arguments

scanpipe/pipelines/alpine_packages.py
quepop: Using `items() ` was @pombredanne 's commit suggestion.

scanpipe/pipelines/alpine_packages.py
tdruez: I disagree with the "makes reading much easier" in the cases mentioned above but "more resistant to refactorings" may be a fair point.
You can leave it as-is then ;)

tdruez: You could store that call in a `unscanned_packages` to help with the for loop layout.

scanpipe/pipelines/alpine_packages.py
tdruez: @pombredanne why using items() here instead of values()?

tdruez: @quepop values() would be better since you do not use the `image_id` variable.

pombredanne:
I saved a clone of this repo and branches so we can revisit this when we have this but this will happen in PurlDB using package sets rather than in ScanCode.io proper. We have implemented this for some package types already, and this is where the feature would be best homed.

pombredanne:
See https://github.com/nexB/purldb/issues/307 for the follow up.

Issue #288, opened 2021-08-17
Get scancode.io version and scanpipe run errors in API


JonoYang:
I am updating the readme at https://github.com/nexB/scancode.io-reference-scans where I am placing scanpipe project statistics in a table. I would like to show the scancode.io version used and errors encountered for each project. However, I am unable to get the scancode.io version or number of errors encountered in a project from the `api/projects/<project uuid>` endpoint. It might make more sense to add the scancode.io version and scanpipe run errors count in the `api/runs/<run uuid>` endpoint, as scancode.io could be updated between different runs.

tdruez:
@JonoYang did you have the chance to look at the PR https://github.com/nexB/scancode.io/pull/290?

JonoYang:
@tdruez Everything looks good! I'm able to get the scancode.io version and error count from the API.

PR #290, opened 2021-08-17
Add a scancodeio_version field to the Run model

tdruez:
Also add errors, resources, and packages count in the Project details API view


tdruez:
@JonoYang this PR should address your suggestions in https://github.com/nexB/scancode.io/issues/288
Please have a look and let me know ;)

aalexanderr (changes requested):
I think this line should be removed in this PR: 
https://github.com/nexB/scancode.io/blob/main/scanpipe/pipes/output.py#L158
as version will get serialized with each Run

tdruez:
@aalexanderr we still want to keep track of the version that was used to generate the output though.

aalexanderr:
@tdruez makes sense, thanks for the reply! I'm unsure how to dismiss my review?

aalexanderr (approved):

JonoYang:
@tdruez This is perfect. Thanks!

JonoYang (approved):

PR #336, opened 2021-10-04
[WIP] Add OpenIDConnect Authentication

aalexanderr:
This PR aims to add OIDC integration for scancode.io.

Why:
We want to run scancode.io on public address and would prefer to have the ability to give access to it only to the ONAP community.

I've asked question on weekly meeting if you guys are interested in OIDC integration for scancode.io (as we were adding it for our instance) and got response that yes, you are. I probably should've created an issue for it.

Commits currently pushed were authored by Mateusz Sobocinski (@matteuszs @msobocinsk) who just left our company thus I'll  try to adjust this to be mergable. I'll squash the commits, add Mateusz's SoB & add docs regarding usage.

We have tested it with keycloak (there is new target in the makefile adding keycloak instance in docker but needs manual configuration for now, should I describe it here/automate it so it can be checked easily?)

The main question I have before removing WIP label is - should it be enabled by default?


Makefile
```diff
@@ -108,7 +116,7 @@ sqlite:
 	@$(MAKE) migrate
 
 run:
-	${MANAGE} runserver 8001
+	${MANAGE} runserver 0.0.0.0:8001
```
aalexanderr, line 119: should probably be reverted, i see no need to expose on all addresses

Makefile
```diff
@@ -129,4 +137,21 @@ docs:
...
 	@${ACTIVATE} sphinx-build docs/ docs/_build/
 
+keycloak:
+	@echo "-> Run Keycloak Identity provider"
+	docker run -dit --name scancodeio-keycloak \
+                    --env KEYCLOAK_USER=${KEYCLOAK_USER} \
+                    --env KEYCLOAK_PASSWORD=${KEYCLOAK_PASSWORD} \
+                    --publish ${KEYCLOAK_HTTPS_BIND_PORT}:8443 \
+                    jboss/keycloak
+	@echo "-> Keycloak launched. Please visit https://localhost:${KEYCLOAK_HTTPS_BIND_PORT} and configure realm"
```
aalexanderr, line 147: add script to configure using kcadm.sh
https://github.com/keycloak/keycloak-documentation/blob/master/server_admin/topics/admin-cli.adoc

Makefile
```diff
@@ -129,4 +137,21 @@ docs:
 	rm -rf docs/_build/
 	@${ACTIVATE} sphinx-build docs/ docs/_build/
 
+keycloak:
+	@echo "-> Run Keycloak Identity provider"
+	docker run -dit --name scancodeio-keycloak \
```
aalexanderr, line 142: My guess is that -it flags are not needed here

Makefile
```diff
@@ -129,4 +137,21 @@ docs:
...
+                    --env KEYCLOAK_USER=${KEYCLOAK_USER} \
+                    --env KEYCLOAK_PASSWORD=${KEYCLOAK_PASSWORD} \
+                    --publish ${KEYCLOAK_HTTPS_BIND_PORT}:8443 \
+                    jboss/keycloak
+	@echo "-> Keycloak launched. Please visit https://localhost:${KEYCLOAK_HTTPS_BIND_PORT} and configure realm"
+
+oidc:
+	@echo "-> Configure OIDC settings"
+	@echo "" >> ${ENV_FILE}
+	@echo OIDC_OP_REALM_ENDPOINT=\"https://keycloak:${KEYCLOAK_HTTPS_BIND_PORT}/auth/realms/${KEYCLOAK_REALM_NAME}/\" >> ${ENV_FILE}
```
aalexanderr, line 152: add env for `https://keycloak` so it can be easily configured when having keycloak running on different domain

tdruez:
Could you provide some context and explanation about those changes?

aalexanderr:
> Could you provide some context and explanation about those changes?

Added some context to the PR msg, let me know if I missed something and I'll be happy to write it down :)

Issue #63, opened 2020-12-07
Create a Homepage for presenting Scancode.io


tdruez:
Available at https://scancode.io/

See the https://github.com/nexB/scancode.io/tree/pages branch for progress using GitHub Pages.

theCodeKeep:
Sir, I want to contribute in this project. 
I am new in Open Source, it would be really helpful if you guide me through.

Thanks and Regards

tdruez:
The goal a to create a simple presentation homepage of the ScanCode.io project.

The code should be in HTML/CSS.

The page should list the key features and benefits of the app and also provide all the relevant links to:
- The documentation
- The GitHub repo
- The source
- The releases

@theCodeKeep Sir, feel free to submit a prototype in a branch and we can iterate from there.

ThatOneTallKid:
I also want to contribute to this task

tdruez:
@ThatOneTallKid please do ;)

ThatOneTallKid:
Any specific color schema you want me to follow?

tdruez:
@ThatOneTallKid not in particular, make sure the code is structured in a way to make it easy to make changes to the color scheme.

ThatOneTallKid:
Ok sure sir!

gauravgulati15:
@tdruez Can I contribute to this task if it's not complete?

tdruez:
@gauravgulati15 Yes, see https://github.com/nexB/scancode.io/issues/63#issuecomment-754794624 for layout suggestions.

gauravgulati15:
Okay, will get back to you asap!

gauravgulati15:
@tdruez is there any logo for ScanCode.io or should I just use the nexB logo?

gauravgulati15:
@tdruez I have made the draft home page for ScanCode.io. The GitHub repo link -[HomePage](https://github.com/gauravgulati15/ScanCode.io-HomePage)

tdruez:
@gauravgulati15 There's no logo yet, feel free to put simply ScanCode.io in text as a placeholder for now.
Your link of the draft is not working by the way.

gauravgulati15:
@tdruez  Oh I am sorry for that! I made the repository private by mistake, now its public you can see it!

divya814:
@tdruez Sir, I wish to work for the enhancement of the homepage.

tdruez:
@divya814 sounds good 👍

tdruez:
[...]

Issue #339, opened 2021-10-08
Add content to scancode.io


pombredanne:
Here is a first shot of what we could put on the scancode.io homepage proper.

ScanCode.io is a open source server to scan and analyze your code for origin, license (and soon vulnerabilities) using scripted pipelines. In particular, it integrates ScanCode toolkit and many libraries to provides advanced pipelines for license and package detection. 

Because not one scan fits all the codebases, you can select a specific scanning pipeline for each project. For instance, ScanCode.io provides a best-in-class pipeline to analyse in details a Docker container in a way that is deeply aware of the specific structure of a container image with its layered file systems and combination of system and application packages. And this will handle equally well Debian, RPM and Alpine Linux distributions, each with their own unique characteristics.
Or you can analyze a complete virtual machine image, or specifically a single application package from Pypi, each with specialized pipelines.

Scaning pipelines are designed to be easily extended and remixed; they can integrate all and any tools in an elegant unified framework.

Backed by a database, ScanCode.io will persist and organize all your scanning projects and make them accessible through a web UI, a JSON REST API or a command line interface.

ScanCode.io is Apache-licensed and written in Python with Django and DRF. It has an extensive [documentation TODO link] to help get you started and is supported by an active [community of users TODO Link to gitter] and adopters. 

[add link to repo and download and doc for installation]
[add screen shot here]

pombredanne:
@hyounes4560 ping ^ do you think this would be enough to get you started? I think we should also update the README.rst file to match something like that BTW.

PR #345, opened 2021-10-12
Add ScanCode.io homepage #63

tdruez:


tdruez:
@hyounes4560 I've setup a new homepage based on the https://startbootstrap.com/template/modern-business template and the content provided by @pombredanne in https://github.com/nexB/scancode.io/issues/339

This should be enough to get you going. You can take over this branch and refine the content and the layout.
We could move to another template if needed.

The result is visible at https://nexb.github.io/scancode.io/index.html

PR #371, opened 2021-11-29
Add new plugins/ with templates and experiments

pombredanne:
This PR adds a new plugins directory with:
- a plugin template
- two experimental plugins

Each of these plugins is designed as a standalone Python package that can be wheel and installed.
They are not installed by default


setup.cfg
```diff
@@ -51,7 +51,10 @@ include_package_data = true
 zip_safe = false
 install_requires =
     # Warning: the importlib_metadata 4.6.x series breaks the documentation builds
-    importlib_metadata==3.10.0; python_version < "3.8"
+    # we still want a non-built-in one as it has the features we need for Python
+    # projects
+    importlib_metadata==4.8.2
```
tdruez, line 56: @pombredanne Why this change?
Make sure to test this on Python3.6 and 3.7 as we were keeping and older version as mention in the "Warning" for compatibility with those Python versions.

pombredanne: @tdruez SCTK uses `importlib_metadata` as a library and we need the latest and greatest AFAIK. It could be a case where we want to vendor the library rather than use it as an external dependency ... let me think a bit about it

PR #378, opened 2021-12-08
Add argument to `make_codebase_resource` to make the created CodebaseResource path relative to the project work path rather than the project codebase path

JonoYang:
I have a pipeline that performs something similar to a Merkle tree calculation on a codebase. In order for this to work, I need a root directory to roll my results up to. scanpipe currently does not have the concept of a root, so my current workaround is to use the project's codebase directory as the root directory. 

I create a new pipeline that creates a `CodebaseResource` for the project codebase directory, but when I call `scanpipe.pipes.scancode.make_codebase_resource`, the created `CodebaseResources` do not have `codebase` as the prefix.

I've modified `scanpipe.pipes.scancode.make_codebase_resource` to have a new argument `relative_to_work_path` such that the path created for the CodebaseResource is related to the project work path, rather than the project codebase path. This means that the created CodebaseResource paths will be prefixed with `codebase` e.g. `codebase/notice.NOTICE`.


tdruez:
@JonoYang I'm not confortable with the idea of storing the `codebase/` prefix **as a choice**.
This lack of consistency in the way CodebaseResource path would be treated will likely be a pain in the future as features will now have to deal with both possibility (prefix or no-prefix).

> In order for this to work, I need a root directory to roll my results up to

What about a special "root" CodebaseResource instance instead?
Maybe a simple ``.`` or ``./`` coule be good enough?

Note that we used to always store the `codebase/` prefix in the database but this had a real estate cost in the UI and storage cost in the database without much benefits (at least at the time).
We could bring back the prefix is we do not find a better approach, but this would not be a choice and should be enforced for all CodebaseResource instances.

JonoYang:
@tdruez Thanks for the review! 

> >In order for this to work, I need a root directory to roll my results up to
>
>What about a special "root" CodebaseResource instance instead?
>Maybe a simple . or ./ coule be good enough?

This would be acceptable too. The thing I am interested in is the ability to use the [`ProjectCodebase`](https://github.com/nexB/scancode.io/blob/e574fa9e815c763ac5025f540dfc05d61122d23e/scanpipe/pipes/codebase.py#L56) class to iterate through the `CodebaseResource`s using code that was originally written for use with scancode-toolkit's `Codebase` and `VirtualCodebase` classes. In particular, the code I have uses `Codebase.walk()` and `Resource.walk()`. If we are to create a special root `CodebaseResource` named `.` or `./`, I would assume that all of the other `CodebaseResource`s in our codebase would need to have `.` as a prefix since `CodebaseResource.walk()` in scancode.io uses the `CodebaseResource` path to determine its children. 

For example:
```
./
./foo/bar.c
./README.txt
```

I'll modify my pipeline to create a `CodebaseResource` with the path `./` and then modify `make_codebase_resource` to allow me to create `CodebaseResource`s with `.` as a path prefix.

tdruez:
@JonoYang my point is that we do not want to store any prefix in the database.
I would suggest to refine the `walk()` functions logic instead to support a special case for a root.

scanpipe/models.py
```diff
@@ -657,9 +658,13 @@ def get_latest_output(self, filename):
 
     def walk_codebase_path(self):
         """
-        Returns all files and directories path of the codebase/ directory recursively.
+        Returns all files and directories path of the codebase/ directory recursively,
+        including the codebase/ directory itself.
         """
-        return self.codebase_path.rglob("*")
+        return itertools.chain(
```
JonoYang, line 664: @tdruez I am trying to get `walk_codebase_path()` to return the `codebase/` directory in addition to the subdirectories and files of `codebase/`. Is there a better way to do this?

scanpipe/models.py
```diff
@@ -1567,7 +1575,7 @@ def children(self, codebase=None):
         https://github.com/nexB/commoncode/blob/main/src/commoncode/resource.py
 
         `codebase` is not used in this context but required for compatibility
-        with the commoncode.resource.VirtualCodebase class API.
+        with the commoncode.resource.Codebase class API.
         """
         exactly_one_sub_directory = "[^/]+$"
```
JonoYang, line 1580: @tdruez There is an issue when we cannot visit the children of the "." CodebaseResource since the logic of `CodebaseResource.walk()` depends on the parent and child CodebaseResources having the same path prefixes. I'm not sure on how to write the regex to walk from "." to a subdirectory or file.

JonoYang (commented):
@tdruez I've made some changes where we now create the "." CodebaseResource when running the `scan_codebase` pipeline. I've done this by modifying `Project.walk_codebase_path()` to return the project codebase/ directory and by modifying `make_codebase_resource` to handle the cases where it is making a CodebaseResource for the project codebase/ directory.

scanpipe/models.py
JonoYang: I've fixed this by using the regex "^[^/]+$" to capture anything without a "/" in the path, e.g.

```
foobar.txt
foobar
```

scanpipe/models.py
```diff
@@ -1580,7 +1588,12 @@ def descendants(self):
         database query on the current CodebaseResource `path`.
         The current CodebaseResource is not included.
         """
-        return self.project.codebaseresources.filter(path__startswith=f"{self.path}/")
+        if self.path == ".":
+            return self.project.codebaseresources.exclude(path=".")
```
tdruez, line 1592: This condition does not seem to be tested. Also, could you document in the dostring why the root need to be treated differently?

scanpipe/models.py
```diff
@@ -1592,10 +1605,14 @@ def children(self, codebase=None):
         https://github.com/nexB/commoncode/blob/main/src/commoncode/resource.py
 
         `codebase` is not used in this context but required for compatibility
-        with the commoncode.resource.VirtualCodebase class API.
+        with the commoncode.resource.Codebase class API.
         """
-        exactly_one_sub_directory = "[^/]+$"
-        children_regex = rf"^{self.path}/{exactly_one_sub_directory}"
+        if self.path == ".":
```
tdruez, line 1610: Same as above, it's not clear nor documented why we have to treat the root differently.

tdruez (changes requested):
> I have not implemented a way to create "." when creating CodebaseResources for the docker pipeline. An immediate solution would be to check to see if a "." exists before creating the CodebaseResources for a layer (https://github.com/nexB/scancode.io/blob/main/scanpipe/pipes/docker.py#L134) 

> Maybe a better idea would be to create "." when the project is created?

No, let's not create resources at project creation time, resources should be created during pipelines execution.

tdruez:
@JonoYang now that we merged the toolkit upgrade, would it be a good time to rebase et complete this PR?

JonoYang:
@tdruez 

I'm merging in the changes from main into this branch. I'm having trouble with `replace_root_path_and_name` (https://github.com/nexB/scancode.io/blob/180-projectcodebase-root-workaround/scanpipe/pipes/scancode.py#L504)  that I made. My function renames the `root` resource path and name to ".", and removes the root path from the remaining resource paths.  It's conflicting with the changes made to how Resource in a Codebase in scancode are related to one another. The `walk` method on `commoncode.Resource` uses the path of a Resource to find its children. If we walk from the root Resource of a Codebase, the walk fails because it cannot find the rest of the Resources in the codebase because no other Resource paths start with ".".

I think what I am trying to do in this function is best implemented somewhere in `commoncode.Codebase`, where I can pass some arguments to the `Codebase` to do this path transformations and modify the logic of `walk` to find the child of a root resource whose path is ".".

JonoYang:
Closing this in favor of #624. Instead of creating a single root CodebaseResource, ProjectCodebase has been updated to walk, not from a single root, but from every CodebaseResource in the root of a Project. This effectively will get the same tree traversal done without large changes to the Project/CodebaseResource models.

PR #391, opened 2022-01-24
Fixing the extensibility of the text box under project data

prathamaror:
Basically the overflow of the x and y were set to scroll so thats why the text was uncanny under the final commit i tend to resolve the issue


prathamaror:
basically the overflow of the x and y were set to scroll so thats why the text was uncanny under the final commit i tend to resolve the issue

PR #398, opened 2022-02-14
Add ability to set Redis instance password

aalexanderr:
Title says it all :)


pombredanne (approved):
This is simple and neat. LGTM, Thank you ++

Issue #335, opened 2021-10-03
Kubernetes support and a few feature ideas


Atharex:
Hi!

I like the direction scancode.io is evolving to and would love to use it myself as part of my DevOps stack in kubernetes :)

Through my own deconstruction of your docker-compose, I've made this k8s manifests file, which should be able to deploy scancode.io as a kubernetes deployment. The input is of course the image built from your Dockerfile. Give it a try and let me know if it works correctly already  (had to rename the file to a `.txt`, because `.yaml` is not supported as an attachment)    
[k8s.txt](https://github.com/nexB/scancode.io/files/7274206/k8s.txt)

Also I wanted to briefly mention a few ideas...

1) In particular I'd love to get scancode.io running as a persistent service in kubernetes, where I can centrally schedule
scanning runs on my docker images in my private registry and also be able to draw up reports to stakeholders in the company.

2) Another really great feature would be to be able to point scancode to a github repository (e.g. https://github.com/nexB/scancode.io) and have it automatically download the main branch/latest release and perform a license scan on it, without manually uploading the zip/tar files of the releases

3) Ideally both of the previous scenarios would support on-demand triggers and scheduled runs

4) Also it would be great to have some sort of policy checks (e.g. like vulnerability policies in container vuln. scanners) where for each project (either a repository or docker image/VM) we could define a set of policies to have a project checked against. The result of a check would be either safe (passed), potentially safe (warning) or unsafe (failed), where each state would hold a custom defined set of licenses (e.g. permissive licenses for safe and any copyleft as unsafe)  and they would be shown as such on the dashboard

tdruez:
Hi @Atharex thanks for your input!

Here's some answers:

> Give it a try and let me know if it works correctly already

I'm actually curious to know if it works on your side first :)

> Another really great feature would be to be able to point scancode to a github repository (e.g. https://github.com/nexB/scancode.io) and have it automatically download the main branch/latest release and perform a license scan on it, without manually uploading the zip/tar files of the releases

You can paste URLs such as "https://github.com/nexB/scancode.io/archive/refs/heads/main.zip" or "https://github.com/nexB/scancode.io/archive/refs/tags/v21.9.6.zip" in the "Download URLs" field.

> Ideally both of the previous scenarios would support on-demand triggers and scheduled runs

One way to trigger Pipeline on demand is to use the REST API https://scancodeio.readthedocs.io/en/latest/rest-api.html#create-a-project

[...]

PR #401, opened 2022-02-15
Add django-probes for wait_for_database

aalexanderr:
This feature is nice-to-have for Kubernetes/OpenShift deployments making
start up of app more streamlined by giving ability for K8s/OS to hold off running
 app until database is up.

As per it's docs it supports all db engines that Django supports:
https://pypi.org/project/django-probes/
License: BSD-3-Clause

Related: #335

### DIFF
--- a/scancodeio/settings.py
+++ b/scancodeio/settings.py
@@ -92,6 +92,7 @@ INSTALLED_APPS = (
     "rest_framework",
     "rest_framework.authtoken",
     "django_rq",
+    "django_probes",
 )
 
 MIDDLEWARE = (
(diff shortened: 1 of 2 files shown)


pombredanne:
This is sleek. Thanks!

pombredanne:
This is sleek. Thanks!

pombredanne:
I guess there is no easy way to have a test for this, right?

pombredanne:
@cco3 would this be something that could make sense for you too?
@Atharex ping too, you were on some issues related to this

cco3:
I'm not sure what our use for it would be.  We have our instance set up with a remote DB.

aalexanderr:
OK so when deploying the app, the DB initialization is done via Django, yet it doesn't provide us with a nice interface for checking if DB is actually up. 
It allows to use the same container with same settings as for runtime to wait with initialization before the database is ready for it.
Another nice-to-have thing would be to do the same for Redis before workers are spawned.
I'll check in a bit how docker-compose up brings everything up but it might be worth making use of there too in this PR.

PR #420, opened 2022-04-07
More rows in the table

lf32:
* increased number of rows from 10 to 25 in the projects table


lf32:
What would be an appropriate number 20 or 25?

PR #422, opened 2022-04-07
Close btn works as expected without archiving the project

lf32:
The close btn whose purpose is to cancel any action taken to archive the project didn't work as expected.

So, had to modify the btn functionality to prevent archiving on clicking the close div btn


tdruez:
@lf32 Thanks for the fix!

lf32:
Thanks for the merge!

PR #425, opened 2022-04-09
license/ web page renders without 500 error

### LINKED ISSUE #424: license/ web page fails with 500 error
Accessing the license/ web page fails with a "500" error

lf32:
Closes #424

https://github.com/nexB/scancode.io/blob/16644a60c6dbe81fa078fc6941fbd8be817fe935/scancodeio/urls.py#L53

include function is updated

### DIFF
--- a/scancodeio/urls.py
+++ b/scancodeio/urls.py
@@ -47,11 +47,11 @@ auth_urlpatterns = [
     path("accounts/profile/", AccountProfileView.as_view(), name="account_profile"),
 ]
 
+
 urlpatterns = auth_urlpatterns + [
     path("admin/", admin.site.urls),
     path("api/", include(api_router.urls)),
-    path("licenses/", licenses.license_list_view, name="license_list"),
-    path("licenses/<path:key>/", licenses.license_details_view, name="license_details"),
+    path("licenses/", include(licenses.urls)),
     path("", include("scanpipe.urls")),
     path("", RedirectView.as_view(url="project/")),
 ]
--- a/scancodeio/urls.py
+++ b/scancodeio/urls.py
@@ -51,7 +51,7 @@ auth_urlpatterns = [
 urlpatterns = auth_urlpatterns + [
     path("admin/", admin.site.urls),
     path("api/", include(api_router.urls)),
-    path("licenses/", include(licenses.urls)),
+    path("license/", include(licenses.urls)),
     path("", include("scanpipe.urls")),
     path("", RedirectView.as_view(url="project/")),
 ]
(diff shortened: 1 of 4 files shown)


pombredanne (commented):
Thanks! This is looking great. Can you add a test too?

lf32:
wrote test cases and moved all urls to main urls.py

tdruez:
@lf32 what's the reason to move the URL patterns outside their related module?

lf32:
using licenses as namespace in includes() was throwing errors

lf32:
if possible license urls can be seperated by placing in `license_urlpatterns` list 
and then `urlpatterns += license_urlpatterns` just like `auth_urlpatterns`
```
license_urlpatterns = [
    ...
]

urlpatterns = auth_urlpatterns + license_urlpatterns + [
   ...
]
```
would it be ok?

tdruez:
@lf32 a better solution is to simply remove the "license_app" namespace during the registration at https://github.com/nexB/scancode.io/blob/main/scancodeio/urls.py#L53

Also, you new TestCase does not follow existing naming convention for tests:

Suggestions:
- get_licenses_test -> test_license_list_view
- get_license_details_test -> test_license_details_view

Lastly, the code is not properly formatted, make sure to run make valid before committing changes.

scancodeio/urls.py
```diff
@@ -47,10 +47,11 @@
     path("accounts/profile/", AccountProfileView.as_view(), name="account_profile"),
 ]
 
+
 urlpatterns = auth_urlpatterns + [
     path("admin/", admin.site.urls),
     path("api/", include(api_router.urls)),
-    path("license/", include((licenses.urls, "license_app"))),
```
tdruez, line 53: No reasons to rename the initial "license".

scanpipe/tests/test_licenses.py
```diff
@@ -0,0 +1,46 @@
...
+# Visit https://github.com/nexB/scancode.io for support and download.
+
+from unittest import mock
+
+from django.test import TestCase
+from django.urls import reverse
+
+
+class LicensesTest(TestCase):
+    """docstring for LicensesTest"""
```
tdruez, line 30: This docstring adds no value, please remove.

tdruez (changes requested):
@lf32 Both tests are failing on the CI. Please fix so we can proceed with the merge ;)

scancodeio/urls.py
lf32: ok

tdruez:
@lf32 Looking good! Thanks for taking care of this ⭐

lf32:
@tdruez Thanks for merging this

PR #431, opened 2022-04-21
Display summary data

JonoYang:
This PR modifies the project detail view to the information from the updated summary plugin for projects that run the `scan_package` pipeline. This is still a rough work in progress.
image


JonoYang:
> @JonoYang could you rebase this one on the latest main? Also, fontawesome is already available in the statics as js lib https://github.com/nexB/scancode.io/blob/main/scancodeio/static/fontawesome-5.15.1.all.min.js Why adding the css version?

I'm not to familiar with front-end development on Django and ended up cargo culting code.

tdruez:
> I'm not familiar with front-end development on Django and ended up cargo culting code.

No worries, let's simply remove the additional fontawesome library files.

JonoYang:
@tdruez That would be most optimal to get this feature out quickly, though I would also be interested to learn more about how to make congruent UI elements that fit the style of existing elements.

tdruez:
> @JonoYang I would also be interested to learn more about how to make congruent UI elements that fit the style of existing elements.

Check the 2 latest commit to get an idea on how I would refine the templates.
Currently (still room for improvements):
Screenshot 2022-07-14 at 22 46 41

PR #435, opened 2022-05-06
411 toolkit upgrade 2

JonoYang:
This PR updates scancode-toolkit in scancode.io, expected test results, as well as updating code that uses deprecated scancode-toolkit functions.


tdruez:
@JonoYang can we make the new dependency system as a separate feature from the toolkit upgrade?

JonoYang:
@tdruez No problem, I'll revert the model changes.

JonoYang:
@tdruez 

I've regenerated the test data using the new command you created and I run into some problems where I would like your opinion on what should be done.

1. ProjectCodebase and the lack of a root codebase directory

I run `regen_test_data.py` and the tests for `ProjectCodebase` all fail now because the `ProjectCodebase` code has no good way of determining the root of a scan. Previous scancode.io Resource creation behavior was to create the `codebase` directory Resource, and then every Resource under `codebase`. The paths of all Resources created starts with `codebase`. The ProjectCodebase object considers the root directory
to be the `codebase` directory (https://github.com/nexB/scancode.io/blob/main/scanpipe/pipes/codebase.py#L71) Since I regenerated the test fixtures, the `codebase` directory Resource is no longer created and Resource paths no longer start with `codebase`

I have some previous attempts where I worked on different solutions. In the branch `starting-paths`, we had the idea of considering the directories under `codebase` directory as individual roots. (https://github.com/nexB/scancode.io/blob/fa6cd0c597bcb53d449c0c119208f0f314955ea2/scanpipe/pipes/codebase.py#L81) Should I expand on the work I did there or should I remove `ProjectCodebase` entirely until we come up with something satisfactory?


2. 

In a scancode JSON output, all Packages detected in a scan are present in a top-level attribute named `packages`. Likewise, all detected Dependencies are placed in the `Dependencies` attribute. Multiple copies of the same package can be present in the `packages` field, if a particular package was detected multiple times in the same codebase. Each copy of this package will have a different `package_uid`. A `package_uid` is the purl of that package with a qualifier named `uuid` that is specific to the scancode run.  e.g. `pkg:pypi/django-audit-tools@0.4.0?uuid=9c19275c-c3fe-43dd-b6ec-a4f2bf65810f`

For each Resource that is for a Package, the `for_packages` for those Resources will be populated with the `package_uid` of the Package they are for.

When I first modified the pipes code to handle the new top-level `packages` field from a scan, I just put the `package_uid` in the `extra_data` field. When I wanted to assign a Resource to a Package, I would then query for that `package_uid` in the `DiscoveredPackage` table (https://github.com/nexB/scancode.io/blob/411-toolkit-upgrade-2/scanpipe/pipes/scancode.py#L392).

I started testing the pipeline using the package `django-audit-tools-0.4.0.tar.gz` and I was getting errors when I tried to look up a `package_uid` from a Resource's `for_packages` field in the `DiscoveredPackage` table. The issue was because the same package was detected multiple times throughout the scan, so there were multiple instances of that package (with different `package_uid`s), but only 1 version of that package was saved to the database (since I am assuming that we can only have unique instances of packages in the DiscoveredPackage table).

I decided to update how I stored the `package_uid` in `extra_data` by creating the field `package_uids` whose value is a list of `package_uid`s. (https://github.com/nexB/scancode.io/blob/774c5b727b52f7d93601e016a9c3a05941ec3d98/scanpipe/pipes/__init__.py#L95)

The `extra_data` probably isn't the best place to keep the `package_uid`s, so I'd like to know how we should handle `package_uid`s in `DiscoveredPackages`. What is the best way to store `package_uid`s on `DiscoveredPackage`s?


3.

The scan output `asgiref-3.3.0_scan.json` created by `regen_test_data.py` is slightly different than the scan output generated by scancode-toolkit:

- No top-level `dependencies` field
- the `for_packages` field on Resources is populated with `purl`s instead of `package_uid`s

Properly handling `package_uid`s in (2) will probably help with this.

tdruez:
> ProjectCodebase and the lack of a root codebase directory

The test data was generated from a time where the `codebase/` prefix was stored in the DB. The test data should have been updated at the time of the code change.

> Should I expand on the work I did there or should I remove ProjectCodebase entirely until we come up with something satisfactory?

Why not completing the https://github.com/nexB/scancode.io/pull/378#pullrequestreview-835085594 PR instead? It seems that my review points were never addressed but it looks like we were close to a merge. Will this PR still relevant to solve this?

> The extra_data probably isn't the best place to keep the package_uids

Probably not the best place, but I think the goal of this first PR is to make upgrade the toolkit with the least amount of models changes to get it out there asap. We should then have tickets and more PRs to work on the optimizations, such as new models and fields for Packages and Dependencies

> The scan output asgiref-3.3.0_scan.json created by regen_test_data.py is slightly different than the scan output generated by scancode-toolkit:

We probably have to update the `JSONResultsGenerator` to include the new format and data.

JonoYang:
@tdruez 

I tagged the `ProjectCodebase` tests with `expectedFailure` and regenerated the `asgiref-3.3.0` test files using `regen_test_data.py`

I've added `scanpipe/tests/data/asgiref-3.3.0_scancode_scan.json` for use in `test_scanpipe_pipes_scancode_create_codebase_resources_inject_policy` (https://github.com/nexB/scancode.io/blob/411-toolkit-upgrade-2/scanpipe/tests/test_pipes.py#L553) because `scanpipe/tests/data/asgiref-3.3.0_scan.json` (generated by `regen_test_data.py`) is not exactly in the same format as what scancode-toolkit outputs.

tdruez:
While running pipelines on this branch:

```
module 'packagedcode.debian' has no attribute 'get_installed_packages'

Traceback:
  File "/app/scanpipe/pipelines/__init__.py", line 115, in execute
    step(self)
  File "/app/scanpipe/pipelines/docker.py", line 93, in collect_and_create_system_packages
    docker.scan_image_for_system_packages(self.project, image)
  File "/app/scanpipe/pipes/docker.py", line 166, in scan_image_for_system_packages
    for i, (purl, package, layer) in enumerate(installed_packages):
  File "/usr/local/lib/python3.9/site-packages/container_inspector/image.py", line 446, in get_installed_packages
    for purl, package in layer.get_installed_packages(packages_getter):
  File "/app/scanpipe/pipes/debian.py", line 31, in package_getter
    packages = debian.get_installed_packages(
```

1. our test suite did not catch this.
2. The packager getter get_installed_packages system was changed on the toolkit side, this needs to be upgraded on the ScanCode.io too

See https://github.com/nexB/scancode-toolkit/commit/fbdba2ec53d4c147c7170150c94e32609783c6cf#diff-29c4686d506744c5a6ae71725c1fa392956829a0c6009c53d26536ef54e01858L57

The following pipe modules are impacted:
- alpine
- debian
- rpm

tdruez:
Scanning https://github.com/bastikr/boolean.py/archive/refs/heads/master.zip with `scan_package` create 1 discovered package in the DB but the summary returns 4 times the same entry in the summary `key_files_packages`. 

https://staging.scancode.io/api/projects/ac26f606-4a7b-4359-b585-72a9616a921c/summary/
https://staging.scancode.io/api/projects/ac26f606-4a7b-4359-b585-72a9616a921c/results/

JonoYang:
> While running pipelines on this branch:
> 
> ```
> module 'packagedcode.debian' has no attribute 'get_installed_packages'
> 
> Traceback:
>   File "/app/scanpipe/pipelines/__init__.py", line 115, in execute
>     step(self)
>   File "/app/scanpipe/pipelines/docker.py", line 93, in collect_and_create_system_packages
>     docker.scan_image_for_system_packages(self.project, image)
>   File "/app/scanpipe/pipes/docker.py", line 166, in scan_image_for_system_packages
>     for i, (purl, package, layer) in enumerate(installed_packages):
>   File "/usr/local/lib/python3.9/site-packages/container_inspector/image.py", line 446, in get_installed_packages
>     for purl, package in layer.get_installed_packages(packages_getter):
>   File "/app/scanpipe/pipes/debian.py", line 31, in package_getter
>     packages = debian.get_installed_packages(
> ```
> 
> 1. our test suite did not catch this.
> 2. The packager getter get_installed_packages system was changed on the toolkit side, this needs to be upgraded on the ScanCode.io too
> 
> See [nexB/scancode-toolkit@fbdba2e#diff-29c4686d506744c5a6ae71725c1fa392956829a0c6009c53d26536ef54e01858L57](https://github.com/nexB/scancode-toolkit/commit/fbdba2ec53d4c147c7170150c94e32609783c6cf#diff-29c4686d506744c5a6ae71725c1fa392956829a0c6009c53d26536ef54e01858L57)
> 
> The following pipe modules are impacted:
> 
> * alpine
> * debian
> * rpm

The `get_installed_packages` function has been removed from the alpine, debian/ubuntu, and rpm package handlers from `packagedcode`. There are now multiple "datafile handlers" that handle the different OS files, like the debian status files. The datafile handler then assemble all of the detected packages, dependencies, etc it finds. For example, this is the `assemble` method for the datafile handler that deals with installed Debian packages. (https://github.com/nexB/scancode-toolkit/blob/develop/src/packagedcode/debian.py#L223) 

There doesn't seem to be an easy way to get the previous `get_installed_packages` functionality back from the new way we process package data. I think the logic from the `assemble` method I linked earlier is what we want, but we'll have to modify it to fit our model in scancode.io.

JonoYang:
@tdruez 

I've brought in the old code that handled processing installed Debian package files and updated enough of it to get the pipeline to work. I had some issues with license detection, but it was resolved when I ran `scancode --reindex-licenses`.

tdruez:
> I've brought in the old code that handled processing installed Debian package files and updated enough of it to get the pipeline to work.

I don't think we want to bring old toolkit code over in ScanCode.io but we should rather make use of the new code.

scanpipe/tests/test_pipelines.py
```diff
@@ -287,6 +287,14 @@ def test_scanpipe_scan_package_pipeline_integration_test(self):
         expected_file = self.data_location / "is-npm-1.0.0_scan_package_summary.json"
         self.assertPipelineResultEqual(expected_file, summary_file, regen=False)
 
+        # Ensure that we only have one instance of is-npm in `key_files_packages`
+        summary_data = json.loads(Path(summary_file).read_text())
+        key_files_packages = summary_data.get("key_files_packages", [])
+        self.assertEqual(1, len(key_files_packages))
+        key_file_package = key_files_packages[0]
+        key_file_package_purl = key_file_package.get("purl", "")
```
tdruez, line 295: The `, "")` is not necessary in this context of testing.

scanpipe/pipes/scancode.py
```diff
@@ -481,10 +481,10 @@ def make_results_summary(project, scan_results_location):
         key_files.append(resource_data)
 
         packages = resource.discovered_packages.all()
-        packages_data = [
-            DiscoveredPackageSerializer(package).data for package in packages
-        ]
-        key_files_packages.extend(packages_data)
+        for package in packages:
+            package_data = DiscoveredPackageSerializer(package).data
+            if package_data not in key_files_packages:
```
tdruez, line 486: It seems a bit much to test on a whole data structure such as `package_data`.
Could we use a more specific id instead?

scanpipe/pipelines/docker.py
```diff
@@ -92,6 +99,32 @@ def collect_and_create_system_packages(self):
...
+
+    def build_inventory_from_scan(self):
+        """
+        Processes the JSON scan results to determine resources and packages.
+        """
+        project = self.project
+        scanned_codebase = scancode.get_virtual_codebase(project, str(self.scan_output))
+        scancode.create_discovered_packages(project, scanned_codebase)
+        scancode.create_codebase_resources(project, scanned_codebase)
+
```
tdruez, line 127: This is a copy paste and it does not respect the DRY principle.
We need a better way to add those common steps to Pipeline subclasses.

tdruez:
@JonoYang I've run the latest Docker pipeline on Staging:

Before the changes https://staging.scancode.io/project/065cf751-16da-4443-ae6c-7f8a8f7bef01/
- 6 packages
- 1,605 resources with proper scan data (license expressions etc..)

After the changes https://staging.scancode.io/project/0efbadd7-2a76-41cd-ab92-6419dbe5ddf8/
- 0 packages
- 1,761 resources without any scan data

tdruez:
@JonoYang I also refined the `key_files_packages` using a QuerySet in place of the `unique_seen`
https://github.com/nexB/scancode.io/pull/435/commits/9757841522be8b04772e513e3c1c576696615a14

Could you confirm that this approach works?

----

It's time to update `test_scanpipe_pipes_scancode_make_results_summary` to test for the full results with an input that include the multiple package instances case.

JonoYang:
@tdruez 

> @JonoYang I've run the latest version of the `docker` pipeline on Staging:
> 
> Before the changes https://staging.scancode.io/project/065cf751-16da-4443-ae6c-7f8a8f7bef01/
> 
> * 6 packages
> * 1,605 resources with proper scan data (license expressions etc..)
> 
> After the changes https://staging.scancode.io/project/0efbadd7-2a76-41cd-ab92-6419dbe5ddf8/
> 
> * 0 packages
> * 1,761 resources without any scan data

I think scancode has an issue getting system installed packages from distroless images. I've downloaded the distroless docker image you used, extracted the layers, and then run scancode with the `--package` and new `--system-package` option and scancode was unable to find any system packages. However, scancode is able to detect system packages from extracted debian/red hat/alpine images.

> @JonoYang I also refined the key_files_packages using a QuerySet in place of the unique_seen
> https://github.com/nexB/scancode.io/commit/9757841522be8b04772e513e3c1c576696615a14
> 
> Could you confirm that this approach works?
>
> It's time to update test_scanpipe_pipes_scancode_make_results_summary to test for the full results with an input that include the multiple package instances case.

I've tested this on https://github.com/bastikr/boolean.py/archive/refs/heads/master.zip and I get the same results as I did before.  I'll create a test for multiple package instances.

scanpipe/pipelines/docker.py
JonoYang: I removed the code I copied into the `Docker` pipeline and instead, I've added `scan_package.ScanPackage` as a superclass to the `Docker` pipeline.

https://github.com/nexB/scancode.io/blob/d867c52fc941414662191458944f03536f155416/scanpipe/pipelines/docker.py#L29

Is this an acceptable solution or should we create a more generic scan pipeline that can be used as the base for other scan-related pipelines?

tdruez:
> I think scancode has an issue getting system installed packages from distroless images. I've downloaded the distroless docker image you used, extracted the layers, and then run scancode with the --package and new --system-package option and scancode was unable to find any system packages. However, scancode is able to detect system packages from extracted debian/red hat/alpine images.

@JonoYang sounds like a regression.

What about the fact that the codebase resources have no scan values?

Before toolkit upgrade:
Screenshot 2022-05-19 at 11 26 24

Now:
Screenshot 2022-05-19 at 11 26 37

scanpipe/pipelines/docker.py
```diff
@@ -77,21 +83,6 @@ def collect_images_information(self):
...
-            docker.create_codebase_resources(self.project, image)
-
-    def collect_and_create_system_packages(self):
-        """
-        Collects installed system packages for each layer based on the distro.
-        """
-        with self.save_errors(rootfs.DistroNotFound, rootfs.DistroNotSupported):
-            for image in self.images:
-                docker.scan_image_for_system_packages(self.project, image)
-
```
pombredanne, line 94: IMHO you cannot get rid of `collect_and_create_codebase_resources()` as a first step and and of `collect_and_create_system_packages()` when scanning a docker image.
Instead we need to find a way to continue to scan by image and in each image by layer otherwise things will be incorrect.

pombredanne (commented):
IMHO we need to revisit the proposed approach in this PR for docker images.

scanpipe/pipes/__init__.py
```diff
@@ -73,15 +73,27 @@ def update_or_create_package(project, package_data, codebase_resource=None):
...
+        package = DiscoveredPackage.objects.get(
+            project=project,
+            **purl_data_and_package_uid
+        )
     except DiscoveredPackage.DoesNotExist:
         package = None
 
+    package_data = {
+        **package_data,
+        "extra_data": {"package_uid": package_uid},
```
tdruez, line 95: > The extra_data probably isn't the best place to keep the package_uids, so I'd like to know how we should handle package_uids in DiscoveredPackages. What is the best way to store package_uids on DiscoveredPackages?

@JonoYang Revisit this whole PR, you were right, using extra_data is not great for storing package_uid and it adds some complexity on the query side of things.
This value seems important enough to be its own field.
Now that the code is more stable, it's probably a good time to add a new field on the DiscoveredPackage model.
Would a simple `package_uid = models.UUIDField()` be good enough as a start?

tdruez (changes requested):
@JonoYang The code looks pretty good.
Do you have a list of what's left to do to complete this PR?

scanpipe/pipes/__init__.py
pombredanne: @tdruez 

> Would a simple package_uid = models.UUIDField() be good enough as a start?

the package_uid is a plain string not an actual bare UUID
It is made from a purl with a uuid as a qualifier, I would use something like 500 chars to account for other possible long qualifiers, or as long as a what we use for URL and purl fields

tdruez: Base on PackageURLMixin model, the purl value can be up to 1695 chars.
I would suggested 1024 for the package_uid to be a the safe side.

pombredanne: perfect.

JonoYang:
@tdruez 

The last few things I finished up were adding the `package_uid` field on the `DiscoveredPackage` model and replacing the ubuntu and redhat test Docker images with smaller images. Other than that, I need to merge https://github.com/nexB/scancode-toolkit/pull/2974 and https://github.com/nexB/scancode-toolkit/pull/2988 into the `develop` branch and then create a new release of scancode so we can test this branch out on staging.

scanpipe/pipelines/scan_package.py
```diff
@@ -121,8 +118,8 @@ def build_inventory_from_scan(self):
         """
         project = self.project
         scanned_codebase = scancode.get_virtual_codebase(project, str(self.scan_output))
-        scancode.create_codebase_resources(project, scanned_codebase)
```
tdruez, line 124: @JonoYang should the `LoadInventory.build_inventory_from_scan` gets the same change?

tdruez, line 124: @JonoYang should the `LoadInventory.build_inventory_from_scan` gets the same change?

JonoYang: I've made the change to `LoadInventory.build_inventory_from_scan`

tdruez (approved):

PR #446, opened 2022-06-12
Enable sorting project list by name and creation date #413

### LINKED ISSUE #413: Enhance Project list view page
When you have start to build up a library of Scans, it can be difficult to find a particular Scan or a set of Scans of interest on the Project summary page /scancode.io/project/. Search works pretty well if you know the name of the Scan or the Pipeline, but we can do better.

Some simple ideas to improve the usability are:

- Increase the default number of displayed rows from 10 to 20 or 25
- Add the Date Completed to the display 
- Enable column sort by Name, Date or Pipeline

lf32:
- [x] Increase table rows to 20
- [X] Added created date column in project table
- [x] Support sorting by name and creation date of the project

### DIFF
--- a/scanpipe/templates/scanpipe/includes/project_list_table.html
+++ b/scanpipe/templates/scanpipe/includes/project_list_table.html
@@ -15,6 +15,7 @@
     <tr>
       <th class="break-all">
         <a href="{{ project.get_absolute_url }}">{{ project }}</a>
+        <span class="has-text-grey is-size-7 is-block" title="{{ project.created_date|date:'N j, Y, P T' }}">Created {{ project.created_date|naturaltime }}</span>
       </th>
       <td>
         {% if project.package_count %}
--- a/scanpipe/templates/scanpipe/project_list.html
+++ b/scanpipe/templates/scanpipe/project_list.html
@@ -21,7 +21,27 @@
             </a>
           {% endif %}
         </div>
-        <a href="{% url 'project_add' %}" class="button is-link">New Project</a>
+        <div>
+          <div class="dropdown is-right is-hoverable">
+            <div class="dropdown-trigger">
+              <button class="button" aria-haspopup="true" aria-controls="dropdown-menu">
+                <span>Sort</span>
+                <span class="icon is-small">
+                  <i class="fas fa-angle-down" aria-hidden="true"></i>
+                </span>
+              </button>
+            </div>
+            <div class="dropdown-menu" id="dropdown-sort-filters" role="menu">
+              <div class="dropdown-content">
+                <a href="?sort=name" class="dropdown-item">Name (A-Z)</a>
+                <a href="?sort=-name" class="dropdown-item">Name (Z-A)</a>
+                <a href="?sort=created_date" class="dropdown-item">Oldest</a>
+                <a href="?sort=-created_date" class="dropdown-item">Newest</a>
+              </div>
+            </div>
+          </div>
+          <a href="{% url 'project_add' %}" class="button is-link">New Project</a>
+        </div>
       </div>
 
       {% include 'scanpipe/includes/search_field.html' %}
--- a/scanpipe/views.py
+++ b/scanpipe/views.py
@@ -102,7 +102,7 @@ class ProjectListView(
     filterset_class = ProjectFilterSet
     template_name = "scanpipe/project_list.html"
     prefetch_related = ["runs"]
-    paginate_by = 10
+    paginate_by = 20
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
--- a/scanpipe/filters.py
+++ b/scanpipe/filters.py
@@ -88,6 +88,9 @@ class FilterSetUtilsMixin:
 
 class ProjectFilterSet(FilterSetUtilsMixin, django_filters.FilterSet):
     search = django_filters.CharFilter(field_name="name", lookup_expr="icontains")
+    sort = django_filters.OrderingFilter(
+        fields=(("name", "name"), ("created_date", "created_date")),
+    )
 
     class Meta:
         model = Project


scanpipe/models.py
```diff
@@ -416,9 +416,6 @@ class Project(UUIDPKModel, ExtraDataFieldMixin, models.Model):
         ),
     )
 
-    class Meta:
-        ordering = ["-created_date"]
```
tdruez, line 420: What's the reason for changing the model?

scanpipe/templates/scanpipe/includes/project_list_table.html
```diff
@@ -16,6 +17,9 @@
       <th class="break-all">
         <a href="{{ project.get_absolute_url }}">{{ project }}</a>
       </th>
+      <td>
+        {{ project.created_date| date:"d-m-Y" }}
```
tdruez, line 21: What about using the natural style as we do on the details view?
https://github.com/nexB/scancode.io/blob/main/scanpipe/templates/scanpipe/project_detail.html#L18

Also, what about displaying this info directly under the project name, with the following style `has-text-grey is-size-7` instead of a column.

Screenshot 2022-06-13 at 10 29 41

scanpipe/templates/scanpipe/project_list.html
```diff
@@ -21,7 +21,27 @@
             </a>
           {% endif %}
         </div>
-        <a href="{% url 'project_add' %}" class="button is-link">New Project</a>
+        <div>
+          <div class="dropdown is-right is-hoverable">
+            <div class="dropdown-trigger">
+              <button class="button" aria-haspopup="true" aria-controls="dropdown-menu">
+                <span>Sort By</span>
```
tdruez, line 28: "Sort" would be enough.

scanpipe/templates/scanpipe/project_list.html
```diff
@@ -21,7 +21,27 @@
...
+                <span class="icon is-small">
+                  <i class="fas fa-angle-down" aria-hidden="true"></i>
+                </span>
+              </button>
+            </div>
+            <div class="dropdown-menu" id="dropdown-menu" role="menu">
```
tdruez, line 34: `dropdown-menu` is too generic for a unique `id`.

scanpipe/templates/scanpipe/project_list.html
```diff
@@ -21,7 +21,27 @@
...
+              <div class="dropdown-content">
+                <a href="?sort=name" class="dropdown-item">Name (A-Z)</a>
+                <a href="?sort=-name" class="dropdown-item">Name (Z-A)</a>
+                <a href="?sort=created_date" class="dropdown-item">Date Created (0-9)</a>
```
tdruez, line 38: Let's use "Newest" and "Oldest" as label.
Also, those should be first in the list as "Newest" is the default ordering.

scanpipe/views.py
```diff
@@ -102,14 +102,22 @@ class ProjectListView(
...
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         context["active_count"] = Project.objects.filter(is_archived=False).count()
         context["archived_count"] = Project.objects.filter(is_archived=True).count()
         return context
 
+    def get_ordering(self):
+        ordering = self.request.GET.get("sort")
+        return (
+            ordering
```
tdruez, line 116: This needs to be implemented on the `ProjectFilterSet` instead.
See https://django-filter.readthedocs.io/en/main/ref/filters.html?highlight=OrderingFilter#orderingfilter

tdruez (changes requested):

scanpipe/templates/scanpipe/project_list.html
lf32: ok

scanpipe/templates/scanpipe/project_list.html
lf32: what would be better?

`project-list-dropdown`

scanpipe/templates/scanpipe/project_list.html
lf32: so dropdown will replace `Date created` with `Oldest`, `Newest`. 
what about sorting based on Names?

scanpipe/templates/scanpipe/project_list.html
tdruez: `project-list-dropdown` does not describe what the dropdown is about.
`dropdown-filters` would be better.

scanpipe/templates/scanpipe/includes/project_list_table.html
lf32: like this?

```
<th class="break-all">
    <a href="{{ project.get_absolute_url }}">{{ project }}</a><br>
    <span class="has-text-grey is-size-7" title="{{ project.created_date|date:'N j, Y, P T' }}">Created {{ project.created_date|naturaltime }}</span>
</th>
```

tdruez: Sure, but let's use the `is-block` class on the span instead of using `<br>` tag for layout.

scanpipe/templates/scanpipe/project_list.html
tdruez: Name options are fine.

scanpipe/templates/scanpipe/includes/project_list_table.html
lf32: there is a text overflow for 97+ characters in the
```
/project/uuid/
/project/uuid/resources/
/project/uuid/resources/number/
```

what about using `is-widescreen` over `is-max-desktop`

scanpipe/views.py
lf32: Is this enough for ProjectFilterSet?

```
sort = django_filters.OrderingFilter(
   fields=(
       ("name", "name"),
       ("created_date", "created_date")
    ),
)
```

scanpipe/models.py
lf32: I thought to change it in the views, will put it back

scanpipe/views.py
lf32: Yes it, works

lf32:
@tdruez it appears that I have lost the review after pushing the changes into a new commit, would you please review it?

scanpipe/views.py
lf32: sure, this part of the code works and I have added it to this [commit](https://github.com/nexB/scancode.io/pull/446/commits/042c70d20f8791d7d8ad8851445311b5c65805c2)

lf32:
There is an issue I came up with after adding this feature

sort option only works for the projects that aren't archived i.e, if user's path is `/project/?is_archived=true` then the sort option from the dropdown redirects user to `/project/?sort=` but not `/project/?is_archived=true&sort=`

lf32:
There is an issue I came up with after adding this feature

sort option only works for the projects that aren't archived i.e,
let's say the user is at `/project/` and clicks at sort option, it just works but
if user's path is `/project/?is_archived=true` then the sort option from the dropdown redirects user to `/project/?sort=` but not `/project/?is_archived=true&sort=`

lf32:
Thanks for the merge! :)

tdruez:
@lf32 Thanks for your contribution!

> There is an issue I came up with after adding this feature sort option only works for the projects that aren't archived i.e,
let's say the user is at /project/ and clicks at sort option, it just works but
if user's path is /project/?is_archived=true then the sort option from the dropdown redirects user to /project/?sort= but not /project/?is_archived=true&sort=

See https://github.com/nexB/scancode.io/pull/465 for the proper implementation of the filters.

lf32:
> See https://github.com/nexB/scancode.io/pull/465 for the proper implementation of the filters.

This is much better, thanks.

Issue #444, opened 2022-06-08
Add support for package instances


JonoYang:
In scancode-toolkit, we report each specific instance with a package_uid to differentiate multiple detection of the same package in a codebase. We should implement this idea in scancode.io when we discover packages in pipelines.

tdruez:
@JonoYang is the recent merged changes address this? Or is there more we can do?

JonoYang:
@tdruez The recent changes adds the `package_uid` field to `DiscoveredPackage` model, which is populated if we use the `scan_package` pipeline. However, we use the function `scan_for_application_packages` (https://github.com/nexB/scancode.io/blob/2d342fa1a1b4f06fea9be73298a00e627cbb6a46/scanpipe/pipes/scancode.py#L311) during the `scan_for_application_packages` step on the `docker`, `rootfs`, and `scan_codebase` pipelines. This function doesn't assign a `package_uid` for discovered application packages, which we would need to add to bring scancode.io application package scanning to be more in line with scancode-toolkit.

The package scanning function of scancode is now two steps, where we first scan the individual Resources for package data, and then afterwards we assemble the package data from multiple resources and assign package instances. We would need a way to do the package assembly step in scancode.io.

JonoYang:
After talking to @pombredanne, it may be necessary to create a new model that for Package and Dependency instances. The current charts in the project detail view show the total number of Packages or Dependencies discovered in a Project. With the introduction of Package and Dependency instances, the Project chart stats are thrown off because we could have multiple copies of the same packages and dependencies. 

1. Using an instance for packages and deps as an intermediate model
Pkg -> PkgInstance -> Res
Pkg -> PkgInstance -> Dep -> DepInstance - >(and when we resolve -> Pkg or PkgInstance)

Here, we can show each unique Package and Dependency, then show the individual instances of the packages, and then show which resources correspond to which Package instance.

pombredanne:
Right now the latest thinking would be to treat the current DiscoveredPackage record as being a package instance. We would add a UID and we would display each copy of a package as it's own. Duplication in the UI and reports is something that is not exceptional but not the main case either and we can solve this in the UI in future.

pombredanne:
Should `datafile_paths` be an actual relationship to a CodebaseResource? ... IMHO yes and we are missing it https://github.com/nexB/scancode-toolkit/blob/bfa6d9632c9ef63250267c8ce3be2d5ddae1a9fc/src/packagedcode/models.py#L1224

[...]

PR #482, opened 2022-08-02
Package scanning update

JonoYang:
This PR updates application package scanning in scancode.io to reflect the updates made to package scanning in scancode-toolkit. Most notably, the DiscoveredDependency model has been added to scancode.io (as the analogue to the Dependency model from scancode-toolkit), and the `scan_for_application_packages` scancode pipe function has been updated to run the `assemble()` methods from scancode-toolkit Package handlers. This allows us to create DiscoveredDependencies for DiscoveredPackages and to properly associate application packages to CodebaseResources.

The Project API has been updated with the following:
  - ``dependency_count``
    - The number of DiscoveredDependencies associated with the project.
  - ``discovered_dependency_summary``
    - A mapping that contains following fields:
      - ``total``
        - The number of DiscoveredDependencies associated with the project.
      - ``is_runtime``
        - The number of runtime dependencies.
      - ``is_optional``
        - The number of optional dependencies.
      - ``is_resolved``
        - The number of resolved dependencies.

The Project detail view has been updated to show the DiscoveredDependencies stats and a DiscoveredDependency view, but more work on the UI will be needed. In particular, the field labels (I'm not sure what the best labels are) and colors used for boolean values should be updated.
image

The DiscoveredDependency view field labels, column widths, and search box (doesn't filter) need to be updated as well.
image

This feature requires https://github.com/nexB/scancode-toolkit/pull/3035, which will be available in scancode-toolkit 31.0.0rc4


JonoYang:
Some more to do's:

- ManyToMany field on DiscoveredPackage to DiscoveredDependencies
- Implement dependency detail view
    - Have hyperlink to the Package a dependency is from
- Resources tab in package detail view does not work

tdruez:
> ManyToMany field on DiscoveredPackage to DiscoveredDependencies

Let's try to make it part of the same migration to limit the number of migration files.

> I can click on the Resources tab in firefox, but not in chrome

Can you try to refresh/reset your cache? The tab system is working fine on my side in chrome.

---

On a more general note, I think there's a bit too much within this one PR and it makes it harder to review and test.
It would have been better to split the resource name, the package assemble, and the dependencies (model and UI) in separate PRs. Not a big deal, but let's try to keep 1 PR per issue in the future.

scanpipe/models.py
```diff
@@ -1582,6 +1608,56 @@ def unique_license_expressions(self):
...
+        Return a sequence of sibling Resource objects for this Resource
+        or an empty sequence.
+
+        `codebase` is not used in this context but required for compatibility
+        with the commoncode.resource.Codebase class API.
+        """
+        if self.has_parent():
+            return self.parent(codebase).children(codebase)
+        return []
+
```
tdruez, line 1660: Are those new method related to this PR?

scanpipe/pipes/scancode.py
```diff
@@ -310,18 +311,108 @@ def scan_for_files(project):
...
+
+def assemble_packages(project):
+    """
+    Create instances of DiscoveredPackage and DiscoveredDependency for `project`
+    from the parsed package data present in the CodebaseResources of `project`.
+    """
+    seen_resource_paths = set()
+    package_data_resources_qs = project.codebaseresources.filter(
+        package_data__isnull=False
+    )
```
tdruez, line 358: Let's add a new method `has_package_data` on the QuerySet.

scanpipe/pipes/scancode.py
```diff
@@ -310,18 +311,108 @@ def scan_for_files(project):
...
+    # detected from
     _scan_and_save(
         resource_qs=resource_qs,
         scan_func=scan_for_package_data,
         save_func=save_scan_package_results,
     )
 
+    # Iterate through CodebaseResources with Package data and handle them using
+    # the proper Package handler from packagedcode
+    assemble_packages(project=project)
```
tdruez, line 333: How are the performances of the new assemble? Do we have some stats about the time spent to assemble compared to the scan_and_save processing?

scanpipe/pipes/scancode.py
```diff
@@ -310,18 +311,108 @@ def scan_for_files(project):
...
+        if resource.path in seen_resource_paths:
+            continue
+
+        logger.info(
+            f"project: {project}:\n"
+            "function: assemble_packages\n"
+            f"Processing: CodebaseResource {resource.path}\n"
+        )
+
+        for package_data in resource.package_data:
```
tdruez, line 369: This wording is a bit confusing, we are using the same `package_data` but on the left side it's a single entry and on the right side it's a list.

scanpipe/pipes/scancode.py
```diff
@@ -310,18 +311,108 @@ def scan_for_files(project):
...
+    package_data_resources_qs = project.codebaseresources.filter(
+        package_data__isnull=False
+    )
+    for resource in package_data_resources_qs:
+        if resource.path in seen_resource_paths:
+            continue
+
+        logger.info(
+            f"project: {project}:\n"
+            "function: assemble_packages\n"
```
tdruez, line 365: Not sure we need to repeat those 2 lines in each log info.

scanpipe/templates/scanpipe/package_list.html
```diff
@@ -27,7 +27,7 @@
         <tbody>
           {% for package in object_list %}
             <tr class="break-word">
-              <td style="min-width: 500px;" title="{{ package.package_uid }}">
+              <td style="min-width: 500px;" title="{{ package.package_url }}">
```
tdruez, line 30: Not sure why we would want to `package_url` as title since we already display it a the text of the <a>

scanpipe/templates/scanpipe/project_detail.html
```diff
@@ -244,6 +271,10 @@ <h3 class="title is-4 has-text-centered mb-3">
     // Packages
     makeChart("package_types", "#package_type_chart", "Package\nType");
     makeChart("package_licenses", "#package_license_chart", "Package\nLicense\nExpression");
+    // Dependencies
+    makeChart("dependency_is_runtime", "#dependency_is_runtime_chart", "Dependency\nis_runtime");
```
tdruez, line 275: What about "Runtime Dependencies", "Optional Dependencies", and "Resolved Dependencies" as labels?

tdruez (changes requested):
Code quality is great! I think we are also missing a few tests for the new functions and methods.

scanpipe/models.py
JonoYang: These methods are called in the `.assembly()` methods of different Package handlers in packagedcode. Some of these methods has logic that determine the Resources of a Package based on the files around a Package manifest. e.g. https://github.com/nexB/scancode-toolkit/blob/develop/src/packagedcode/pypi.py#L185

scanpipe/pipes/scancode.py
JonoYang: `has_package_data()` has been added to `CodebaseResourceQuerySet`

scanpipe/pipes/scancode.py
JonoYang: I've renamed the variables to avoid overloading `package_data`

scanpipe/templates/scanpipe/package_list.html
JonoYang: I've reverted this change.

scanpipe/templates/scanpipe/project_detail.html
JonoYang: I've updated the dependency donut chart labels

scanpipe/pipes/scancode.py
JonoYang: The function now displays these two lines at the beginning of the function, and I've removed them from the subsequent logger messages.

scanpipe/pipes/scancode.py
JonoYang: I don't have any metrics on the performance impact that `assemble_packages` has on application package scanning. What is the process you do to collect performance metrics on different scancode.io parts?

scanpipe/models.py
JonoYang: I've also updated the CodebaseResource walk test to also test for these methods

JonoYang:
@tdruez Thanks! I've responded to your comments. Now, I am wondering about the best approach for #444, where we want to represent individual instances of Packages and Dependencies. Right now, in the case where we have detected the same package multiple times, we create unique packages for each instance even though it is essentially the same package.

scanpipe/models.py
tdruez: Ok let's make sure we have full test coverage for those.

scanpipe/pipes/scancode.py
tdruez: We run the pipeline using various inputs pre-changes then again post-changes and we can compare the execution time of the relevant steps.

Also, a profiling utility is available to go deeper into optimizations:  https://github.com/nexB/scancode.io/blob/main/scanpipe/pipelines/__init__.py#L155

pombredanne: > How are the performances of the new assemble? Do we have some stats about the time spent to assemble compared to the scan_and_save processing?

The new assemble is a must have functional change that we should get working and tested before we can optimize if we ever need to optimize? ... @tdruez did you experience an issue?

JonoYang:
I am closing this PR and splitting it into two: 

- One where we implement the package assembly step
- One where we create the DiscoveredDependency model

PR #485, opened 2022-08-04
Implement package assembly in scancode.io

JonoYang:
This PR updates the application package scanning step in scancode.io to reflect the updates made to package scanning in scancode-toolkit. Package scanning is now a two step process, where we first detect Package data from CodebaseResources, then we iterate through all the Resources with Package data and process them using the new Package handlers from scancode-toolkit.

Issue #445, opened 2022-06-11
Diagnose issues with Resources not associated with a Package


mjherzog:
We have some major gaps in Package Detection where Resources are not associated with a Package or a Package is not detected.
The task here is to run SCIO Scans on a selection of popular public Docker images to identity a representative sample of the Package detection gaps. A suggested list is:

- grafana/grafana:7.2.0
- influxdb:1.8.10
- logstash:7.17.1
- nginx:1.21.6-alpine
- postgres:11.15-bullseye
- python:3.7-slim-bullseye
- rabbitmq:3.9.13-management

pombredanne:
The latest SCTK 31.xx is designed to solve these issues. These images will be an excellent test suite . We should just wait a couple days before starting this to ensure this is tested with a branch that has the latest SCTK with its advanced package files handling capabilities.

JonoYang:
Some of the missing resources are caused by this issue: https://github.com/nexB/extractcode/issues/41

A fix would be to have an option in the commoncode function where we do the portable filename conversion to allow colons in file names.

tdruez:
@JonoYang Is this one ready to be closed since https://github.com/nexB/scancode.io/pull/485 is merged?

JonoYang:
@tdruez This looks to be fixed in the current version of scancode.io lists file with `:` in the filename as Package Resources:
![image](https://github.com/nexB/scancode.io/assets/6044800/ae4aacd3-6299-446e-ac70-f7f7d73442b0)

PR #486, opened 2022-08-04
Implement DiscoveredDependency model

JonoYang:
This PR adds the DiscoveredDependency model to scancode.io. New views, based off the existing DiscoveredPackage views, have been added for DiscoveredDependency. 

This is the second PR split from #482 and is built on top of #485


JonoYang:
@tdruez What would be a better way to write this migration query? https://github.com/nexB/scancode.io/blob/e2c9bcdae3530c379f17478f40303a94a9939659/scanpipe/migrations/0023_migrate_dependencies.py#L8

I am trying to populate the DiscoveredDependency table from the dependency data stored in the JSON fields on DiscoveredPackage before I remove the field.

tdruez:
> What would be a better way to write this migration query?

@JonoYang unfortunatly, I don't think there's an easy way here to use `update()` since the old `dependencies` field uses a complex data structure (JSON field as list of dict).
Since the amount of Packages is much less than Resources, I think we can go ahead with your approach.
Also, `dependencies__isnull=False` is not the proper lookup as it will return `[]` empty list value as well, and `dependencies_data` is not a field on the model, `dependencies` is.

I would refactor the code into something along:

```
def migrate_dependencies_to_discovereddependencies(apps, schema_editor):
    DiscoveredPackage = apps.get_model('scanpipe', 'DiscoveredPackage')
    DiscoveredDependency = apps.get_model('scanpipe', 'DiscoveredDependency')

    package_with_dependencies = DiscoveredPackage.objects.exclude(dependencies=[])
    
    for package in package_with_dependencies:
        for dependencies_data in package.dependencies:
            # Remove non-supported fields from the data dict
            dependencies_data.pop("extra_data", None)
            dependencies_data.pop("resolved_package", None)
            DiscoveredDependency.objects.create(project=package.project, **dependencies_data)
```

scanpipe/migrations/asd/0022_alter_discovereddependency_options_and_more.py
```diff
@@ -0,0 +1,47 @@
+# Generated by Django 4.0.6 on 2022-08-11 00:56
```
tdruez, line 1: I think those file were committed by mistake.

scanpipe/models.py
```diff
@@ -1465,6 +1486,17 @@ class Compliance(models.TextChoices):
...
+        default=list,
+        blank=True,
+        help_text=_("List of Package data detected from this CodebaseResource"),
+    )
+
+    package_data = models.JSONField(
+        default=list,
+        blank=True,
+        help_text=_("List of Package data detected from this CodebaseResource"),
+    )
```
tdruez, line 1499: Double data? ;)

scanpipe/models.py
```diff
@@ -1857,6 +1931,193 @@ def update_from_data(self, package_data, override=False):
...
+        return self.dependency_uid
+
+    def get_absolute_url(self):
+        return reverse("dependency_detail", args=[self.project_id, self.pk])
+
+    @cached_property
+    def for_package_uid(self):
+        if self.for_package:
+            return self.for_package.package_uid
+        else:
```
tdruez, line 2018: `else` not needed.

scanpipe/models.py
```diff
@@ -1857,6 +1931,193 @@ def update_from_data(self, package_data, override=False):
...
+            return ""
+
+    @cached_property
+    def datafile_path(self):
+        if self.datafile_resource:
+            return self.datafile_resource.path
+        else:
```
tdruez, line 2025: `else` not needed.

scanpipe/models.py
```diff
@@ -1857,6 +1931,193 @@ def update_from_data(self, package_data, override=False):
...
+            project=project,
+            for_package=for_package,
+            datafile_resource=datafile_resource,
+            **cleaned_dependency_data,
+        )
+        discovered_dependency.save()
+
+        return discovered_dependency
+
+    def update_from_data(self, dependency_data):
```
tdruez, line 2094: Seems like duplicated code, time for refactoring into a Mixin class. The base code seems generic enough.

scanpipe/migrations/0023_migrate_dependencies.py
```diff
@@ -0,0 +1,52 @@
...
+            dependency_data.pop("extra_data", None)
+            dependency_data.pop("resolved_package", None)
+
+            for_package = None
+            for_package_uid = dependency_data.get("for_package_uid")
+            p = project.discoveredpackages.filter(package_uid=for_package_uid)
+            if p.exists():
+                for_package = project.discoveredpackages.get(
+                    package_uid=for_package_uid
+                )
```
tdruez, line 28: This approach works but it triggers 2 queries.
Another possibility is to directly get your object and catch if it is not available:
```
try:
    for_package = project.discoveredpackages.get(package_uid=for_package_uid)
except (ObjectDoesNotExist, MultipleObjectsReturned):
    for_package = None
```

scanpipe/migrations/0023_migrate_dependencies.py
```diff
@@ -0,0 +1,52 @@
...
+
+            datafile_resource = None
+            datafile_path = dependency_data.get("datafile_path")
+            datafile = project.codebaseresources.filter(path=datafile_path)
+            if datafile.exists():
+                datafile_resource = project.codebaseresources.get(path=datafile_path)
```
tdruez, line 34: Same as https://github.com/nexB/scancode.io/pull/486/files#r955608871

scanpipe/migrations/0025_discovereddependency_name_and_more.py
```diff
@@ -0,0 +1,43 @@
+# Generated by Django 4.0.6 on 2022-08-15 22:25
```
tdruez, line 1: Can this one be merged with 0022?

scanpipe/migrations/0028_alter_discovereddependency_options.py
```diff
@@ -0,0 +1,28 @@
+# Generated by Django 4.1 on 2022-08-26 01:36
```
tdruez, line 1: Seems like those should be in the 0022 migration.

scanpipe/migrations/0024_remove_discoveredpackage_dependencies_data.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('scanpipe', '0023_migrate_dependencies'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
```
tdruez, line 13: This could be done along in 0027, to limit a bit the number of migration files.

tdruez:
@JonoYang What's the status on your side? Could you provide a list of what's left todo on this one?

Also, I've made some changes to have a consistent "discovereddependencies" related name, this solves issues with the filter system, see https://github.com/nexB/scancode.io/pull/486/commits/5d8cd436265ef2566eab2039887da266e8318b85#diff-d71c8a8d98cec2c888bbf875d6a34a481041ba8a56656ec516aa53b0b7f6f194R1978

Lastly, I've refactored the `update_from_data` method into a Mixin to remove the duplicated code, see https://github.com/nexB/scancode.io/pull/486/commits/dba4891709751ddcba75f0e01ded57d6b49e191d#diff-d71c8a8d98cec2c888bbf875d6a34a481041ba8a56656ec516aa53b0b7f6f194R1013
The only difference with the `DiscoveredDependency.update_from_data` is the following skip exception:
`field_name in PURL_FIELDS`. Is it something we do not want to check on the dependency side?
If that's the case, we can add a `skip_purl_fields` kwarg to the `update_from_data` method to make it usable for both models.

JonoYang:
@tdruez 

> What's the status on your side? Could you provide a list of what's left todo on this one?

I think we're ready to go for now as the important stuff is here: 
- model created
- pipelines updated to create DiscoveredDependencies and relate them to the Package and Resource they're from
- view updated to show DiscoveredDependency results and their relations

I was taking a look at #445 with this branch and I run into an issue with rubygem packages and their dependencies being reported multiple times. This is an issue with scancode-toolkit, and not scancode.io (https://github.com/nexB/scancode-toolkit/issues/3072). I've also run into the license scan errors that you reported earlier.

> I've made some changes to have a consistent "discovereddependencies" related name, this solves issues with the filter system, see https://github.com/nexB/scancode.io/commit/5d8cd436265ef2566eab2039887da266e8318b85#diff-d71c8a8d98cec2c888bbf875d6a34a481041ba8a56656ec516aa53b0b7f6f194R1978

Thanks! I've updated the `verbose_name` and the `verbose_plural_name` and I was wondering why the related name didn't change.

> The only difference with the DiscoveredDependency.update_from_data is the following skip exception:
field_name in PURL_FIELDS. Is it something we do not want to check on the dependency side?

I skipped this initially because we did not have purl fields in the DiscoveredDependency model. We can have this check again now that we're using the PURL field mixin in DiscoveredDependencies.

Issue #467, opened 2022-07-21
Codebase resource "Name" is not the filename 


pombredanne:
I found highly surprising that the name field of the CodebaseResource model (and therefore also in the UI) is NOT the actual file name or directory name, but rather has been stripped from it's extension, which is stored in a separate field. What's the rationale for this?

JonoYang:
This inconsistency is causing problems when updating the application package scanning step with the new Package assembly methods from packagedcode. The assembly methods on different Package handlers look for particular Package manifest files by their filename with extension (e.g. https://github.com/nexB/scancode-toolkit/blob/develop/src/packagedcode/npm.py#L72). Because scancode.io stores the filename without the extension in the `name` field of `CodebaseResource`, these manifest files are never found and the detected Package data cannot be processed properly.

tdruez:
@JonoYang is https://github.com/nexB/scancode.io/compare/package-scanning-update ready for review?
I don't see any PR yet ;)

PR #493, opened 2022-08-09
Store extension in CodebaseResource.name #467

JonoYang:
* Create migration on CodebaseResource to append extension to name, if name does not end with extension
    * Update exptected test results

Signed-off-by: Jono Yang <jyang@nexb.com>


tdruez:
@JonoYang LGTM. Thanks for the fix 👍

Issue #499, opened 2022-08-17
Compressed docker image not recognized in docker pipeline


pombredanne:
I tried to scan this as a docker image: https://github.com/nyph-infosec/daggerboard/releases/download/v1.0.1/daggerboard_docker_image_v1-0-1.tgz and nothing happened. This is a docker "saved" image that's been gzipped and renamed to .tgz

tdruez:
Looking at https://github.com/nexB/scancode.io/blob/main/scanpipe/pipes/docker.py#L48 and prior changes, it seems that we never supported the `.tgz` extension.

It should be fairly easy to update the glob pattern to capture `.tgz` along tars, assuming the `container_inspector.utils.extract_tar` supports `tgz` as well.

PR #504, opened 2022-08-18
Add support for .tgz extension for input files in Docker pipeline #499

tdruez:


pombredanne (approved):
Thanks! we could need something a bit more flexible than two extension patterns, but that's plenty good enough for now.

Issue #518, opened 2022-09-02
Error in running docker image on windows 10.


35C4n0r:
ERROR 1 (Solved):

  ![image](https://user-images.githubusercontent.com/70096901/188196880-c2ccf3b2-53b1-407c-bafb-9bceb12a2bd4.png)
  While running docker-compose up on windows
  I'm getting the error:
  
  ```
  /usr/bin/env: ‘python\r’: No such file or directory
  ```
  
  ```host not found in upstream "web:8000" in /etc/nginx/conf.d/default.conf:2
  nginx_1   | nginx: [emerg] host not found in upstream "web:8000" in /etc/nginx/conf.d/default.conf:2
  scancodeio_nginx_1 exited with code 1
  worker_1  | wait-for-it: timeout occurred after waiting 15 seconds for web:8000
  worker_1  | /usr/bin/env: ‘python\r’: No such file or directory
  ```
  
  Tried the Solution provided in #412 , but that didn't work.

ERROR 2:

```
nginx_1   | 2022/09/04 07:08:32 [notice] 1#1: start worker process 33
worker_1  | wait-for-it: timeout occurred after waiting 15 seconds for web:8000
db_1      | 2022-09-04 07:09:26.348 UTC [27] FATAL:  the database system is starting up
web_1     | Traceback (most recent call last):
web_1     |   File "/usr/local/lib/python3.9/site-packages/django/db/backends/base/base.py", line 282, in ensure_connection
web_1     |     self.connect()
web_1     |   File "/usr/local/lib/python3.9/site-packages/django/utils/asyncio.py", line 26, in inner
web_1     |     return func(*args, **kwargs)
web_1     |   File "/usr/local/lib/python3.9/site-packages/django/db/backends/base/base.py", line 263, in connect
web_1     |     self.connection = self.get_new_connection(conn_params)
web_1     |   File "/usr/local/lib/python3.9/site-packages/django/utils/asyncio.py", line 26, in inner
web_1     |     return func(*args, **kwargs)
web_1     |   File "/usr/local/lib/python3.9/site-packages/django/db/backends/postgresql/base.py", line 215, in get_new_connection
web_1     |     connection = Database.connect(**conn_params)
web_1     |   File "/usr/local/lib/python3.9/site-packages/psycopg2/__init__.py", line 122, in connect
web_1     |     conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
web_1     | psycopg2.OperationalError: FATAL:  the database system is starting up
[...]

PR #536, opened 2022-09-30
A Note for Windows users to set git autocrlf = false before cloning the repo to avoid issues like #518 .

35C4n0r:
Git automatically converts LF to CR LF while cloning a repository. This was leading to CR LF errors in windows such as issue https://github.com/nexB/scancode.io/issues/518.
This error can be avoided altogether by just setting git's autocrlf to false.

Signed-off-by: 35C4n0r <jaykumar20march@gmail.com>

### DIFF
--- a/scanpipe/models.py
+++ b/scanpipe/models.py
@@ -649,7 +649,7 @@ class Project(UUIDPKModel, ExtraDataFieldMixin, models.Model):
         pipeline Run.
         """
         shutil.rmtree(self.tmp_path, ignore_errors=True)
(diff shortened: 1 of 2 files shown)


pombredanne:
Hey Jay, 
Regarding autocrlf this would still make sense to have as a note in the development docs

tdruez:
Added note in https://github.com/nexB/scancode.io/pull/553/commits/47f09589c2673d80727fa4c52c5208f728b94bc8

PR #585, opened 2023-01-13
added docker-compose requirement to [Run With Docker]

### LINKED ISSUE #584: docker-compose does not come pre-installed with Linux
Addition of installation guide to include **docker-compose** as it is not pre-installed with Linux distributions.

OctoPie23:
This PR fixes #584 by adding:
 - docker-compose to the **installation.rst** installation file guide.
 - change all the `docker compose` to `docker-compose` 

Signed-off-by: OctoPie23 <coveringguy0123@gmail.com>

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -45,7 +45,8 @@ create an **environment file**, and **build the Docker image**::
 
     git clone https://github.com/nexB/scancode.io.git && cd scancode.io
     make envfile
-    docker compose build
+    apt install docker-compose
+    docker-compose build
 
 .. note::
     You need to rebuild the image whenever ScanCode.io's source code has been
@@ -56,7 +57,7 @@ Run the App
 
 **Run your image** as a container::
 
-    docker compose up
+    docker-compose up
 
 At this point, the ScanCode.io app should be running at port 80 on your Docker host.
 Go to http://localhost/ on a web browser to **access the web UI**.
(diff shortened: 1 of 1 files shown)


OctoPie23:
@pombredanne Can you look into this small documentation change?

docs/installation.rst
```diff
@@ -45,7 +45,8 @@ create an **environment file**, and **build the Docker image**::
 
     git clone https://github.com/nexB/scancode.io.git && cd scancode.io
     make envfile
-    docker compose build
+    apt install docker-compose
```
pombredanne, line 48: We are not supporting only Linux and only Debian-based distros. We have two ways to deal with this. We can use docker-compose 1.x which is a Python package or the docker-compose 2.x which is a single executable is Go.

docs/installation.rst
```diff
@@ -56,7 +57,7 @@ Run the App
 
 **Run your image** as a container::
 
-    docker compose up
+    docker-compose up
```
pombredanne, line 60: You are likely assuming this is using docker-compose 2.x?

pombredanne (changes requested):
Thank you ... but there are some issues with this seemingly tiny doc change ... It would need to be improved.

OctoPie23:
Oh, I'm sorry🤦‍♂️. I got confused.
I had downloaded the docker-compose 1.x version, therefore the command `docker compose` was returning the error `compose: command not found`.
Would it be possible to refer to the [Manual Installation](https://docs.docker.com/compose/install/linux/#install-the-plugin-manually) in case someone encounters a problem?
Else
We could probably just close this PR.

Issue #210, opened 2021-06-19
 Project data text box should be extensible


pombredanne:
The Project data text box size is fixed and IMHO it would help if it were:
1. extensible both in length and width
2. to help with looking at the data and possibly copy/paste or saving it, it would be great to have a way to open that text (may be raw) in a separate window

ayushk26:
Hey, can I try this issue?

pombredanne:
@ayushk26 It would be great. What's your approach for this?

ayushk26:
I went through the code within the template folder, and I am not sure about where the input regarding project data was given. Could you please help me with it.

prathamaror:
I was thinking about using tkinter to get the job done I have a pretty good idea how things are gonna transcend should I start working on this? I have some questions prior to that

pombredanne:
@prathamaror we do not use tkinter .... this is a Django web page for the project details from https://github.com/nexB/scancode.io/blob/20e8057b46ee62066b940acc1753a49aa803b2e3/scanpipe/views.py#L150 and using this template https://github.com/nexB/scancode.io/blob/20e8057b46ee62066b940acc1753a49aa803b2e3/scanpipe/templates/scanpipe/project_detail.html#L84

prathamaror:
[https://github.com/nexB/scancode.io/pull/392](url)

prathamaror:
#392

prathamaror:
#392 do review and let me know if there are any further changes needed

prathamaror:
https://github.com/nexB/scancode.io/pull/392#issue-1113181502

RishiPratap:
Hi there can you assign me this issue

AyanSinhaMahapatra:
You don't need to be `assigned` for creating PRs for an issue, just us know how you're solving the issue and go for it.

AkshayMuthal:
Hello, is this issue still open?

AkshayMuthal:
Hi @pombredanne, I have raised a MR for this issue. Please do let me know if any changes/suggestions are there from your end.
https://github.com/nexB/scancode.io/pull/667

tdruez:
Implemented in https://github.com/nexB/scancode.io/pull/777

PR #596, opened 2023-02-03
made Text box  extensible #210

2pushkaraj3:
issue: Project data text box should be extensible #210
As specified earlier [scancode.io/scanpipe/templates/scanpipe/project_detail.html] consisted of projects-extra-data and base.html had project-extra--data with overflow in y was set to scroll in the vertical direction with 300px maximum height. So basically the overflow-y: auto and overflow-x: auto properties will allow the scrollbar to appear only when necessary in both directions. Tried with different maximum heights and I think it is working fine.


tdruez:
Implemented in https://github.com/nexB/scancode.io/pull/777

PR #624, opened 2023-03-01
Update ProjectCodebase

JonoYang:
* Remove root property and create root_resources
    * Update walk to walk from root resources
    * Update tests


JonoYang:
@tdruez Is this something we want to keep? https://github.com/nexB/scancode.io/blob/update-ProjectCodebase/scanpipe/templates/scanpipe/project_tree.html

tdruez:
> @tdruez Is this something we want to keep?

Nop, removed in https://github.com/nexB/scancode.io/commit/3cf6394b8d4e08803df74aa44ae15435bbf86c88

tdruez:
@JonoYang  This seems related to https://github.com/nexB/scancode.io/pull/378
What's the status on #378? Is it still relevant following this PR?

JonoYang:
@tdruez 

#378 addresses the same issue I had before, where I wanted a root CodebaseResource created whenever we create a Project such that I can do depth-first traversal of a Project codebase. I made changes to create a root CodebaseResource with the name and path of `.`, but there were issues with creating a VirtualCodebase object from scancode.io JSON results where the we could not visit any other Resource than the root Resource in the VirtualCodebase because no other Resource starts with `.` (VirtualCodebase/Resource walk logic uses the parent path to find its children). The changes I made in this PR work around this issue by just walking from each Resource in the root of the project, rather than from a single root resource. This way, we do not need to perform major changes to how Project/CodebaseResource and commoncode Codebase/VirtualCodebase/Resource works.

tdruez:
@JonoYang could you add an entry with those implementation details in the CHANGELOG about the changes?

PR #626, opened 2023-03-01
Add instructions for vulnerablecode integration

TG1999:


tdruez:
@TG1999 Could you add an introduction paragraph about the context of this integration.
You added the "how" but we also need the "why".

Also, a section about the usage of the `find_vulnerabilities` pipeline would be useful, following the configuration parts.

TG1999:
@tdruez done!

tdruez:
It would be more interesting to display screenshots about the actual vulnerability results of running the pipeline instead of showing how to add and run a pipeline, which is already cover in other tutorial.

TG1999:
@tdruez done!

PR #714, opened 2023-05-04
Add d2d step for JavaScript/TypeScript match

keshav-space:


scanpipe/pipelines/deploy_to_develop.py
```diff
@@ -49,6 +49,7 @@ def steps(cls):
             cls.checksum_map,
             cls.java_to_class_map,
             cls.jar_to_source_map,
+            cls.javascript_match,
```
pombredanne, line 52: this would be a map now, not a match... we renamed these. Matching is reserved for matching to the purldb, mapping is for d2d

scanpipe/pipelines/deploy_to_develop.py
```diff
@@ -106,6 +107,10 @@ def jar_to_source_map(self):
         """Map a .class compiled file to its .java source."""
         d2d.jar_to_source_map(project=self.project, logger=self.log)
 
+    def javascript_match(self):
+        """Match JavaScript and TypeScript files."""
```
pombredanne, line 111:
```suggestion
        """Map a packed or minified JavaScript, TypeScript, CSS and SCSS to its source."""
```

scanpipe/pipes/d2d.py
```diff
@@ -38,6 +41,37 @@
 IGNORE_EXTENSIONS = ()
 IGNORE_PATHS = ("gradleTest/",)
 
+PROSPECTIVE_JAVASCRIPT_MATCH = {
+    ".scss": [
+        {
```
pombredanne, line 46: Could you add some comments to describe what this is? Or add a field in the dictionary with some help text?

scanpipe/pipes/d2d.py
```diff
@@ -387,3 +421,192 @@ def purldb_match(project, extensions, logger=None):
             start_time=start_time,
         )
         _resource_purldb_match(project, to_resource)
+
+
+def javascript_match(project, logger=None):
```
pombredanne, line 426: match->map

scanpipe/pipes/d2d.py
```diff
@@ -387,3 +421,192 @@ def purldb_match(project, extensions, logger=None):
...
+
+        if any_match:
+            break
+
+
+def _is_minified_and_map_compiled_from_source(
+    to_resources, from_source, minified_extension
+):
+    """Check if a minified file and its map were compiled from a source file."""
+    if not minified_extension:
```
pombredanne, line 529: What is this `minified_extension`?

scanpipe/pipes/d2d.py
```diff
@@ -387,3 +421,192 @@ def purldb_match(project, extensions, logger=None):
...
+                or _source_in_map(minified_map_file, source_file_name)
+            ):
+                return True
+            break
+
+    return False
+
+
+def _source_mapping_in_minified(resource, source_mapping):
+    """Check if a string contains a specific string in its last 5 lines."""
```
pombredanne, line 557:
```suggestion
    """Return True if a string contains a specific string in its last 5 lines."""
```

scanpipe/pipes/d2d.py
```diff
@@ -387,3 +421,192 @@ def purldb_match(project, extensions, logger=None):
...
+                _source_content_in_map(minified_map_file, from_source)
+                or _source_in_map(minified_map_file, source_file_name)
+            ):
+                return True
+            break
+
+    return False
+
+
+def _source_mapping_in_minified(resource, source_mapping):
```
pombredanne, line 556: Why mark it as "private" with underscore?

pombredanne (changes requested):
Thanks! LGTM overall but I did not review it all! Some nits for you

scanpipe/pipes/d2d.py
keshav-space: File extension of minified files like .js, .css

scanpipe/pipes/d2d.py
keshav-space: not private just an indication that it's helper function

scanpipe/pipes/d2d.py
```diff
@@ -387,3 +440,189 @@ def purldb_match(project, extensions, logger=None):
...
+            resource_index,
+            resource_count,
+            last_percent,
+            increment_percent=10,
+            start_time=start_time,
+        )
+        _resource_javascript_map(from_resource, to_resources)
+
+
+def _resource_javascript_map(from_resource, to_resources):
```
tdruez, line 481: The code of this function is too complex with 3 imbricated `for` loops. 
It needs to be refactored into smaller pieces.

scanpipe/pipes/d2d.py
```diff
@@ -387,3 +440,189 @@ def purldb_match(project, extensions, logger=None):
...
+    for resource in to_resources:
+        if resource.path.endswith(minified_extension):
+            minified_file = resource
+        elif resource.path.endswith(f"{minified_extension}.map"):
+            minified_map_file = resource
+        if minified_file and minified_map_file:
+            if _source_mapping_in_minified(minified_file, source_mapping) and (
+                _source_content_in_map(minified_map_file, from_source)
+                or _source_in_map(minified_map_file, source_file_name)
+            ):
```
tdruez, line 568: Too hard to read.

scanpipe/pipes/d2d.py
```diff
@@ -387,3 +440,189 @@ def purldb_match(project, extensions, logger=None):
...
+    try:
+        with open(map_file.location) as f:
+            data = json.load(f)
+            contents = data.get("sourcesContent", [])
+            return any(origin_sha1 == _sha1(content) for content in contents)
+    except json.JSONDecodeError:
+        return False
+
+
+def _get_basename_and_extension(filename):
```
tdruez, line 616: This seems very tight to the JS world, the function name, docstring and variables are too generic and not reflecting this.

scanpipe/tests/pipes/test_d2d.py
```diff
@@ -326,3 +326,206 @@ def test_scanpipe_pipes_d2d_path_map(self):
         self.assertEqual(to1, relation.to_resource)
         self.assertEqual("path", relation.map_type)
         self.assertEqual({"path_score": "3/3"}, relation.extra_data)
+
+    def test_scanpipe_pipes_d2d_get_basename_and_extension(self):
```
tdruez, line 330: Testing on ".ts" would be nice since we have a comment about it.

> `.d.ts` should be tested first before `.ts`.

tdruez (changes requested):

scanpipe/pipes/d2d.py
```diff
@@ -500,3 +551,198 @@ def match_purldb(project, extensions, logger=None):
...
```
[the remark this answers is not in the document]
tdruez: Actually, I think it would make sense to move all those JS related function into their own `js.py` pipe module. See for example `jvm.py`.
We would keep the 2 main mapper functions `map_javascript` `_map_javascript_resource` in the d2d module.

tdruez:
@keshav-space See https://github.com/nexB/scancode.io/pull/714#discussion_r1189487264

scanpipe/pipes/d2d.py
```diff
@@ -500,3 +551,198 @@ def match_purldb(project, extensions, logger=None):
             start_time=start_time,
         )
         _match_purldb_resource(project, to_resource)
+
+
+def map_javascript(project, logger=None):
```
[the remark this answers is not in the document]
keshav-space: > IMO that won’t change the outcome and will needlessly complicate the logic. Since in almost all case we have one-to-many relation ( single source file is generating upto 7-8 different file), in such case if we start by `to` side we will need to supply another layer of logic to selectively get those specific 7-8 file that might relate to a particular file from `from` side.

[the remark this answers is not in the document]
tdruez: Fair enough, maybe add this explanation in the docstring of the map_javascript ;)

scanpipe/pipes/d2d.py
```diff
@@ -500,3 +551,198 @@ def match_purldb(project, extensions, logger=None):
...
+
+
+def _relate_js_files(matches, from_resource, map_type, path_score):
+    for match in matches:
+        relation = CodebaseRelation.objects.filter(
+            from_resource=from_resource,
+            to_resource=match,
+            map_type=map_type,
+        )
+        if not relation.exists():
```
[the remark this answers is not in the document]
keshav-space: Yes that's old code we don't need that anymore.

scanpipe/filters.py
```diff
@@ -286,6 +286,7 @@ def __init__(self, *args, **kwargs):
             ("many", "Many map"),
             ("java_to_class", "java to class"),
             ("jar_to_source", "jar to source"),
+            ("js_compiled", "js csompiled"),
```
tdruez, line 289: Typo in csompiled

keshav-space: fixed ;)

scanpipe/pipes/d2d.py
```diff
@@ -40,6 +43,55 @@
...
+# each dict specifies the file extension of the transformed minified file, and
+# the `to_related` key specifies the related file extensions that are generated
+# alongside the transformed file (inclusive of minified file extension).
+#
+# For example, the `.scss` key maps to a list of two dict. The first dict
+# specifies that `.scss` file is transformed into minified `.css` files along
+# with `.css.map` and `_rtl.css`, and the second dict specifies that `.scss`
+# is also transformed into minified `.scss.js` file along with `.scss.js.map`.
+
+PROSPECTIVE_JAVASCRIPT_MAP = {
```
tdruez, line 57: This should be relocated to the js.py module.

scanpipe/pipes/d2d.py
```diff
@@ -524,3 +576,97 @@ def match_purldb(project, extensions, logger=None):
...
         _match_purldb_resource(project, to_resource)
+
+
+def map_javascript(project, logger=None):
+    """Map a packed or minified JavaScript, TypeScript, CSS and SCSS to its source."""
+    project_files = project.codebaseresources.files().only("path")
+
+    from_resources = project_files.from_codebase().filter(
+        extension__in=PROSPECTIVE_JAVASCRIPT_MAP.keys()
+    )
```
tdruez, line 587: Suggestion to improve readability:
```
javascript_extensions = PROSPECTIVE_JAVASCRIPT_MAP.keys()
from_resources = project_files.from_codebase().filter(extension__in=javascript_extensions)
```

scanpipe/pipes/d2d.py
```diff
@@ -524,3 +576,97 @@ def match_purldb(project, extensions, logger=None):
...
+            start_time=start_time,
+        )
+        _map_javascript_resource(from_resource, to_resources)
+
+
+def _map_javascript_resource(from_resource, to_resources):
+    path = Path(from_resource.path.lstrip("/"))
+    # Skip any dot files.
+    if path.name.startswith("."):
+        return
```
tdruez, line 621: Shouldn't we exclude this case directly in the `from_resources` QuerySet?

scanpipe/pipes/js.py
```diff
@@ -0,0 +1,114 @@
...
+    """Return the basename and extension of a JavaScript/TypeScript related file."""
+    # The order of extensions in the list matters since
+    # `.d.ts` should be tested first before `.ts`.
+    js_extensions = [".d.ts", ".ts", ".js", ".jsx", ".scss"]
+    for ext in js_extensions:
+        if filename.endswith(ext):
+            extension = ext
+            break
+    else:
+        raise ValueError(f"{filename} has an invalid extension")
```
tdruez, line 112: `invalid extension` is unclear, what about `f"{filename} is not JavaScript related"`?

scanpipe/pipes/d2d.py
```diff
@@ -524,3 +576,97 @@ def match_purldb(project, extensions, logger=None):
...
+        any_map = False
+
+        for candidate in PROSPECTIVE_JAVASCRIPT_MAP.get(from_extension, []):
+            query = Q(
+                *[
+                    Q(path__endswith=f"{current_path}{extension}")
+                    for extension in candidate["to_related"]
+                ],
+                _connector=Q.OR,
+            )
```
tdruez, line 640: This QuerySet is expensive as it triggers a table scan. We should not call it multiple times within 2 for loops, but rather try to get everything at once and then iter on the values.

@pombredanne We likely want to use the new indexing system here as well, to be generated in `map_javascript` with all the candidates for mapping.

tdruez (changes requested):

scanpipe/pipes/d2d.py
keshav-space: @pombredanne we're only interested in collecting files with these ['.scss', '.js', '.jsx', '.ts', '.d.ts'] extension in `from` side.

pombredanne: and the .css may be?

pombredanne: @keshav-space my earlier point was about using `.keys()` when the base iteration  on a mapping is always on keys, so there is no need to call `.keys()`

keshav-space: > and the .css may be?

To my understanding it's not normal to generate a `.css.map` form `.css` file.
In a usual scenario `.css.map` file will generally come from`.scss`, `.sass` but never directly through `.css `

keshav-space: > @keshav-space my earlier point was about using `.keys()` when the base iteration on a mapping is always on keys, so there is no need to call `.keys()`

Won't that be too clever from the point of readability? ;)

pombredanne: that's a common idiom, but I am fine either way.

tdruez:
Moved to https://github.com/nexB/scancode.io/pull/728

Issue #569, opened 2022-12-08
Update ScanCode.io with latest ScanCode Toolkit with license detection support


pombredanne:
ScanCode Toolkit with license detection support is about to be release and we need to update SCIO accordingly.

tdruez:
@pombredanne are you referring to version [v31.2.4](https://github.com/nexB/scancode-toolkit/releases/tag/v31.2.4) or unreleased [v32.0.0](https://github.com/nexB/scancode-toolkit/releases/tag/v32.0.0rc1) ?

AyanSinhaMahapatra:
@tdruez this is the unreleased v32.0.0 for which we only have release candidates now (v32.0.0rc2 is almost there), and this is WIP now from me

PR #715, opened 2023-05-05
Upgrade ScanCode-toolkit to latest v32 #569

tdruez:


scanpipe/models.py
```diff
@@ -1471,6 +1467,41 @@ def has_value(self, field_name):
...
+        help_text=_(
+            "List of license matches that are not proper detections and potentially "
+            "just clues to licenses or likely false positives. Those are not included "
+            "in computing the detected license expression for the resource."
+        ),
+    )
+    percentage_of_license_text = models.FloatField(
+        blank=True,
+        null=True,
+        help_text=_(
```
pombredanne, line 1500: Let's use this instead (I am updating SCTK too):

```suggestion
        help_text=_("Percentage of file words detected as license text or notice."),
```

scanpipe/tests/data/is-npm-1.0.0_scan_codebase.json
```diff
@@ -68,14 +68,38 @@
...
+            }
+          ],
+          "identifier": "mit-a822f434-d61f-f2b1-c792-8b8cb9e7b9bf",
+          "license_expression": "mit"
+        }
+      ],
+      "other_license_expression": "",
+      "other_license_expression_spdx": "",
+      "other_license_detections": [],
+      "extracted_license_statement": "['MIT']",
```
pombredanne, line 98: This is a bug!! we should not have this ugly repr-like in the extracted license statement but proper YAML like this was done before with the SCTK 31.x in "declared_license" . @AyanSinhaMahapatra ping :)

pombredanne (approved):
All good, I added a few nitpickings on the doc and two follow up issues:

- https://github.com/nexB/scancode.io/issues/732
- https://github.com/nexB/scancode.io/issues/733

AyanSinhaMahapatra (approved):
@tdruez thanks! I've tested with scanning and using multiple and different pipelines, migrating old data, and the views. Everything works as expected and nicely. I'll also open follow up issues on more upgrades for later.

PR #743, opened 2023-05-22
Add pipelines for loading nuget-inspector scans and running d2d with scancode scans

TG1999:


TG1999:
@pombredanne what's your take on this, does this pipeline needs to be more generic or good to go for now?

tdruez (changes requested):
1. The first concern is that you are addressing two unrelated topics within the same PR. We want to keep the PR focused on a single issue.
2. Instead of multiplying the number of default pipelines, I would suggest that we improve the existing ones. For example, instead of adding a new `DeployToDevelopFromScancodeScans` with a minor pipe change, why not improve the support for inputs such as scan results directly in the `get_inputs` logic?
3. Same question for `LoadNugetInspectorScans`, I'm wondering why we would want a specialized pipeline for this instead of having support in the existing ones.

tdruez:
@TG1999 Any comments on the above?

pombredanne:
This is useful and needed, but we should re-design how we want to have these pipelines or features:

1. IMHO we cannot have too many of these smaller pipelines.
2. Yet using scancode scans as input instead of codebase can be extremely useful as a generic feature when practical (like when the backing code is no longer needed afterwards)... See https://github.com/nexB/scancode.io/issues/1090
3. nuget-inspector should be refactored to either produce a standard scancode-format data output OR create some SBOM to the same effect See https://github.com/nexB/scancode.io/issues/1089

I am closing this for now... but let's keep the branch open

PR #755, opened 2023-05-31
Match .class files against purldb

JonoYang:
This PR updates the d2d pipeline to match .class files against purldb. In addition to addling `.class` to `purldb_resource_extensions`, I've also removed the code that tags `.class` files with NO_JAVA_SOURCE after the java to class map step. This is done so we can collect the `.class` files for matching in the purldb match step.


scanpipe/pipes/d2d.py
```diff
@@ -235,10 +235,6 @@ def map_java_to_class(project, logger=None):
             )
         _map_java_to_class_resource(to_resource, from_resources, from_classes_index)
 
-    # Flag not mapped .class in to/ codebase
```
JonoYang, line 238: @tdruez 

Do we want to update the status of unmapped/unmatched `.class` files after the purldb match?

pombredanne: IMHO matching implies a status update of sorts.

JonoYang: @pombredanne 

The status of Resources are updated when they have a successful match against purldb. (https://github.com/nexB/scancode.io/blob/class-file-match/scanpipe/pipes/d2d.py#L490)

tdruez:
@JonoYang The PurlDB matching code looks fine but I do not see where/when the `extra_data__directory_content` value is set.

Also, you need to run `make valid` before committing to fix the code format.

JonoYang:
@tdruez 

> @JonoYang The PurlDB matching code looks fine but I do not see where/when the `extra_data__directory_content` value is set.

It's set by the function [`compute_directory_fingerprints()`](https://github.com/nexB/purldb/blob/main/matchcode-toolkit/src/matchcode_toolkit/fingerprinting.py#L72) from matchcode-toolkit. I will release this as its own library.

> Also, you need to run `make valid` before committing to fix the code format.

I've run `make valid` and fixed the code format issues.

JonoYang:
I'm closing this in favor of #782

Issue #781, opened 2023-06-20
Improve purldb match performance


JonoYang:
Currently, purldb matching is slow and many improvements can be made to speed things up:

- Send multiple sha1s off for matching 
- Reducing the number of requests sent to purldb
- Implement directory matching,

tdruez:
@JonoYang Can we close this issue following the PR merge?

PR #782, opened 2023-06-20
purldb matching updates #781

JonoYang:
This PR updates the purldb matching portion of the d2d pipeline. This is done by:

- Implementing directory matching
- Sending multiple sha1 for matching at once vs sending one sha1 at a time
- Caching package data when matching Resources


JonoYang:
Looks like I have some db ordering differences when I am regenerating tests.

tdruez:
@JonoYang is this PR complete and ready to be reviewed?
If not, please provide the latest status and a list of remaining tasks.

JonoYang:
@tdruez 

I'd appreciate your thought on my changes so far, but the remaining tasks are:
- In purldb, implement POST API endpoint on packages and resources to allow for filtering by multiple sha1
- Update scio purldb code to use the new endpoint
- Tests

tdruez:
Hi @JonoYang the code looks generally fine, but here's my concern regarding the process that seems quite un-efficient:

1. We use JSONResultsGenerator which loads the entire Project data from the database (are we even using all that data, I don't think Package, Dependencies, etc.. are part of generating directory fingerprints)
2. All that data is Serialized and written to memory
3. We create a VirtualCodebase that deserializes that data on its instance
4. Latestly we can run the fingerprinting

I'm wondering why we need to use `VirtualCodebase` at all. Cannot we achieve this fingerprinting by just getting the right data from the DB to compute and update a few records?

Also, do we have the spec of the "directory fingerprint" available somewhere?

JonoYang:
@tdruez 

> Hi @JonoYang the code looks generally fine, but here's my concern regarding the process that seems quite un-efficient:
> 
> 1. We use JSONResultsGenerator which loads the entire Project data from the database (are we even using all that data, I don't think Package, Dependencies, etc.. are part of generating directory fingerprints)
> 2. All that data is Serialized and written to memory
> 3. We create a VirtualCodebase that deserializes that data on its instance
> 4. Finally we can run the fingerprinting
> 
> I'm wondering why we need to use `VirtualCodebase` at all. Cannot we achieve this fingerprinting by just getting the right data from the DB to compute and update a few records?

There are a couple of reasons why I am using `VirtualCodebase` here:

- We use the VirtualCodebase because the code that computes the directory fingerprints was originally written to be run as a scancode-toolkit plugin, which uses the Codebase as the main data structure.

- The process fingerprinting the directories of a codebase is pretty much computing a Merkle tree using a locality sensitive hash (https://github.com/nexB/purldb/blob/main/matchcode-toolkit/src/matchcode_toolkit/halohash.py). This involves traversing the codebase depth-first, so we can compute the fingerprints of the lowest directories all the way to the root directory of the codebase. At each directory, we are putting into the hash the sha1 values of the Resources in the directory and all other Resources under that directory, recursively. This is the directory content fingerprint. The tree nature of the Codebase/VirtualCodebase model makes it easy to compute the directory fingerprints. There's no reason why you can't compute the directory fingerprints without using VirtualCodebase, but the reason I didn't do that is because I don't know how query the database in such a way that resembles depth-first tree traversal and ensuring that CodebaseResources that are returned are in the same order as if I was traversing a VirtualCodebase the same way.

- I tried using ProjectCodebase (https://github.com/nexB/scancode.io/blob/main/scanpipe/pipes/codebase.py#L99) and using that with my code (https://github.com/nexB/purldb/blob/main/matchcode-toolkit/src/matchcode_toolkit/fingerprinting.py#L108). This works, but I found that computing the directory fingerprints took very long.  I thought that computing the directory fingerprints in memory using the VirtualCodebase would be faster. I had to run `JSONResultsGenerator` to populate the `VirtualCodebase`. I understand this isn't great. I figured that running `JSONResultsGenerator` at the beginning of the pipeline, when there is very little information in the db other than CodebaseResources with basic file info, wouldn't be too bad. 

> Also, do we have the spec of the "directory fingerprint" available somewhere?

We have docs on the hash algorithm and implementation we use, but not on the process of computing directory fingerprints. I outline the directory fingerprinting computation process in the second bullet point in my response above. I will update the docs.

scanpipe/pipes/d2d.py
```diff
@@ -524,8 +590,55 @@ def match_purldb(project, extensions, matcher_func, logger=None):
...
+                        resources_by_sha1=resources_by_sha1,
+                        package_data_by_purldb_urls=package_data_by_purldb_urls,
+                    )
+                    resources_by_sha1 = defaultdict(list)
+                    last_percent = pipes.log_progress(
+                        logger,
+                        resource_index,
+                        resource_count,
+                        last_percent,
+                        increment_percent=10,
```
tdruez, line 630: This whole logic is too complex. Could it be simplify by chunking instead of depending on a harcoded `== 1000`.
See https://github.com/nexB/scancode.io/blob/main/scanpipe/pipes/vulnerablecode.py#L75 and https://github.com/nexB/scancode.io/blob/main/scanpipe/pipes/vulnerablecode.py#L205
for some examples of splitting the batch of item on several requests.
(We can move the `chunked` function out of the vulnerable code pipes module for the generic one if needed)

tdruez (changes requested):
@JonoYang see my comments.

The new `get_basic_virtual_codebase` seems much better than relying on the results generator. 👍

scanpipe/pipes/purldb.py
```diff
@@ -77,7 +77,9 @@ def request_get(url, payload=None, timeout=None):
     if not url:
         return
 
-    params = {"format": "json"}
+    params = {}
+    if "format=json" not in url:
+        params.update({"format": "json"})
```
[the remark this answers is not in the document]
JonoYang: Before I created the POST-based API endpoint for Package filtering on purldb, I used the query params-based filtering on the `api/packages` endpoint. I would use the `next` link from the response to visit the next page for more package data. The `next` link already has `format=json` appended to it. When I would use the `next` link from the response to collect the package data, `format=json` would continuously be appended to the url. This caused issues when visiting multiple pages of results. I forget what error was returned by the server, but it was due to the number of query parameters I was passing.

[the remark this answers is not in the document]
tdruez: Thanks for the explanation, but we need to find a better way than depending on a string that may be in the URL.

What if we were to set the format as the content-type header instead of passing this as an argument? We should apply this header upfront for the whole session and do not worry about the format in the rest of the code.
```
session = requests.Session()
session.headers.update({"Content-Type": "application/json"})
```
Could you check if this would work with your code? (and the issue described above)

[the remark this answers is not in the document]
JonoYang: @tdruez 

When I added `session.headers.update({"Content-Type": "application/json"})` to `scanpipe/pipes/purldb.py`, the POST requests made to purldb would not work. I get the message `WARNING Bad Request: /api/packages/filter_by_checksums/` on purldb.

scanpipe/pipes/d2d.py
JonoYang: @tdruez I tried using `chunked`, but it relies on list slicing to divide the iterable passed to it, which doesn't work on querysets. I've adapted this code from vulnerablecode (https://github.com/nexB/scancode.io/blob/781-purldb-matching-updates/scanpipe/models.py#L1612) and added it to the `CodebaseResource` queryset to return `CodebaseResource`s in groups of arbitrary sizes.

scanpipe/pipes/purldb.py
```diff
@@ -99,24 +101,56 @@ def request_post(url, data, timeout=None):
...
+            response = request_post(url=next_page, data=data, timeout=timeout)
+            if response and response.get("count"):
+                results.extend(response["results"])
+            next_page = response.get("next")
+    return results
+
+
+def match_packages(sha1_list, timeout=None, api_url=PURLDB_API_URL):
+    """Match a list of SHA1 in the PurlDB for package-type files."""
+    data = {"sha1": sha1_list}
```
TG1999, line 120: @JonoYang please add `enhance_package_data` as well in this PR itself, if it makes sense

JonoYang: Ack, I've made the update.

scanpipe/models.py
```diff
@@ -1600,6 +1601,18 @@ def path_pattern(self, pattern):
...
+        """
+        Resources that have the key `directory_content` set in the `extra_data`
+        field.
+        """
+        return self.filter(~Q(extra_data__directory_content=""))
+
+    def paginated(self, per_page=5000):
+        """Iterate over a (large) QuerySet by chunks of ``per_page`` items."""
+        for page in Paginator(self, per_page=per_page):
+            yield page.object_list
```
tdruez, line 1614: Could you add some content in the docstring about why this is different from using the `queryset.iterator` method and about the benefit/context for using this?

scanpipe/pipes/d2d.py
```diff
@@ -496,30 +493,107 @@ def create_package_from_purldb_data(project, resource, package_data):
...
+        matched_package = create_package_from_purldb_data(
+            project=project, resource=resource, package_data=package_data
+        )
+        resource.refresh_from_db()
+        matched_to_purldb_after = resource.status == flag.MATCHED_TO_PURLDB
+        if matched_package and matched_to_purldb_before != matched_to_purldb_after:
+            # Increment matched_resources_count if we have `matched_package` and
+            # when we are able to match on a Resource for the first time.
+            match_resources_count += 1
+    return match_resources_count
```
tdruez, line 515: it seems that the `pipes.update_or_create_package` will be called for each single resource. This is not very efficient. We should call the package creation code once, then deal with the resource association with an optimized queryset.

scanpipe/pipes/d2d.py
```diff
@@ -496,30 +493,107 @@ def create_package_from_purldb_data(project, resource, package_data):
...
 
-def match_purldb_resource(project, resource):
-    """Match a single file resource in the PurlDB."""
-    sha1_list = [resource.sha1]
-    if resource.path.endswith(".map"):
-        sha1_list.extend(js.source_content_sha1_list(resource))
 
-    if results := purldb.match_resource(sha1_list=sha1_list):
+def match_purldb_resource(
+    project, resources_by_sha1, package_data_by_purldb_urls={}, **kwargs
```
tdruez, line 546: Never use a mutable for a default argument value in Python ;)

scanpipe/pipes/d2d.py
```diff
@@ -548,13 +638,51 @@ def match_purldb(project, extensions, matcher_func, logger=None):
...
-        if matched_package:
-            matched_count += 1
 
     logger(f"{matched_count:,d} resource(s) matched in PurlDB")
 
 
+def match_purldb_directories(project, logger=None):
+    """Match against PurlDB selecting codebase directories."""
+    to_directories = (
+        project.codebaseresources.directories().to_codebase().order_by("path")
```
tdruez, line 648: Could you add a comment about the need for the ordering in the context of the directories matching?

scanpipe/pipes/purldb.py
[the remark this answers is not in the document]
tdruez: Ok, let's keep the exception for now and we'll deal with side effects if any.

tdruez (changes requested):

JonoYang:
@tdruez I am failing this test: https://github.com/nexB/scancode.io/actions/runs/5958512864/job/16162845512#step:8:534 (https://github.com/nexB/scancode.io/blob/781-purldb-matching-updates/scanpipe/tests/pipes/test_d2d.py#L579)

Here, we are matching a JS map file that contains two JS packages within it, so one resource has been matched to two packages. Before my changes to the purldb and matching code, the purldb match function considers this to be a single resource matched. After my changes, the function considers this to be two resources matched, due to the way I process matches by [sha1](https://github.com/nexB/scancode.io/blob/781-purldb-matching-updates/scanpipe/pipes/d2d.py#L575) instead of per CodebaseResource. I'm not sure how to best get the previous behavior back with the current system I implemented.

tdruez:
@JonoYang I don't fully understand the problem here so I cannot provide you guidelines for a solution.

> I'm not sure how to best get the previous behavior back with the current system I implemented.

Do we want to previous behavior though?

My feeling is that If there are 2 sha1 matches, the log message should be explicit about it.

JonoYang:
@tdruez 

I ended up finding a way to prevent double counting of already-matched CodebaseResources, that matches with how we counted the total number of matched CodebaseResources from before and updated the log message to relate the number of SHA1s sent to PurlDB to the number of resources matched.

tdruez (changes requested):
@JonoYang See the minor suggestions.
We'll be ready to merge after those.

tdruez:
@JonoYang Also, I would suggest adding a DEFAULT_TIMEOUT variable in the header of the purldb module, and using it instead of None in the function args:
```
def request_get(url, payload=None, timeout=DEFAULT_TIMEOUT)

def request_post(url, data, timeout=DEFAULT_TIMEOUT)
```

JonoYang:
@tdruez 

I've set `DEFAULT_TIMEOUT` in `scanpipe/pipes/purldb.py` to be 30 seconds (https://github.com/nexB/scancode.io/blob/781-purldb-matching-updates/scanpipe/pipes/purldb.py#L52). Is this a reasonable timeout length?

Issue #809, opened 2023-07-14
Support ignoring resources from `ignored_resources` in ABOUT file


AyanSinhaMahapatra:
Reference: https://github.com/nexB/aboutcode-toolkit/issues/527

> We have the about_resource filepaths, which are there to keep track of filepaths which are summarized by the ABOUT file, but often there are folders in a package/product which are patched/added and needs more detailed curation/summarization, or needs to be ignored from the overall package summarization.
> It is more concise and easier to list the paths to ignore, than to list all the paths which are under the ABOUT file

This upgrades the functionality of ABOUT files used to document code/resources which are patched/vendored/added: https://github.com/nexB/scancode.io/issues/740

PR #810, opened 2023-07-14
Support ignored_resources attribute in ABOUT files

AyanSinhaMahapatra:
Reference: https://github.com/nexB/scancode.io/issues/809


scanpipe/pipes/resolve.py
```diff
@@ -89,6 +89,10 @@ def resolve_about_package(input_location):
     if about_resource := about_data.get("about_resource"):
         package_data["filename"] = list(about_resource.keys())[0]
 
+    if ignored_resources := about_data.get("ignored_resources"):
+        package_data["extra_data"] = {}
+        package_data["extra_data"]["ignored_resources"] = list(ignored_resources.keys())
+
```
tdruez, line 95:
```suggestion
extra_data = {"ignored_resources": list(ignored_resources.keys())}
package_data["extra_data"] = extra_data
```

scanpipe/pipes/d2d.py
```diff
@@ -614,12 +614,23 @@ def _map_about_file_resource(project, about_file_resource, to_resources):
...
     if not codebase_resources:
         # If there's nothing to map on the ``to/`` do not create the package.
         return
 
+    # Ignore resources for paths in `ignored_resources` attribute
+    if ignored_resources:
+        for resource_path in ignored_resources:
+            codebase_resources = codebase_resources.exclude(
+                path__contains=f"/{resource_path.lstrip('/')}"
+            )
```
tdruez, line 632: A better pattern would be to craft the list of lookups and call the queryset only once.

scanpipe/pipes/d2d.py
```diff
@@ -614,12 +614,23 @@ def _map_about_file_resource(project, about_file_resource, to_resources):
         # Cannot map anything without the about_resource value.
         return
 
+    ignored_resources = None
```
tdruez, line 617:
```suggestion
    ignored_resources = []
```

tdruez (changes requested):

scanpipe/pipes/d2d.py
AyanSinhaMahapatra: Added this, thanks!

scanpipe/pipes/resolve.py
AyanSinhaMahapatra: Added this, thanks!

scanpipe/pipes/d2d.py
AyanSinhaMahapatra: This would be much better, implemented this, thanks!

AyanSinhaMahapatra:
@tdruez thanks for the feedback, I have addressed these now. All green and ready for review again!

scanpipe/pipes/d2d.py
```diff
@@ -614,12 +616,23 @@ def _map_about_file_resource(project, about_file_resource, to_resources):
...
     if not codebase_resources:
         # If there's nothing to map on the ``to/`` do not create the package.
         return
 
+    # Ignore resources for paths in `ignored_resources` attribute
+    if ignored_resources:
+        lookups = Q()
+        for resource_path in ignored_resources:
+            lookups |= Q(**{"path__contains": resource_path})
+        codebase_resources = codebase_resources.filter(~lookups)
```
tdruez, line 634: @AyanSinhaMahapatra one clarification: `ignored_resources` is a list of exact paths right?
This code looks like it deals with patterns (ie: `contains`) more than exact resource paths.
I want to make sure this is the intended behavior.
Could you point me to the spec of the AboutCode `ignored_resources` field?

AyanSinhaMahapatra: @tdruez it is a list of paths which should be ignored, but this also means ignoring anything under that path, i.e. if the path is a directory, then all the files/sub-folders are to be ignored too. So it's not just list of exact paths which should be ignored.
See the aboutcode spec at:
https://aboutcode-toolkit.readthedocs.io/en/latest/specification.html#referencing-the-file-or-directory-documented-by-an-about-file say:
```
In this example, the ABOUT file documents a whole sub-directory, with some sub-paths under the directory ignored:

    about_resource: linux-kernel-2.6.23
    ignored_resources: linux-kernel-2.6.23/Documentation

```
Also at https://aboutcode-toolkit.readthedocs.io/en/latest/specification.html#optional-information-fields:
```
ignored_resources: A list of paths under the about_resource path, which are not documented in the ABOUT file, and the information in the ABOUT file does not apply to these subpaths.
```

tdruez: @AyanSinhaMahapatra thanks for the precisions.
Is that behavior "files/sub-folders" covered by the unit tests?

AyanSinhaMahapatra: > Is that behavior "files/sub-folders" covered by the unit tests?

Not entirely, as the package used for the tests did not have that much sub-folders to be able to do this properly.
But since `contains` is used I didn't think it would require to be added as a test. I have tested with other bigger packages locally and this works well with a list of paths which are directories, and everything below them are ignored from being assigned to the package.

Neither the function `map_about_files` nor `_map_about_file_resource` has unit tests before, so I didn't add more unit tests there too. So we just have a full pipeline run test instead, and I modified the same.

I've added this to the about file:
```
about_resource: flume-ng-node-1.9.0.jar
ignored_resources:
  - flume-ng-node-1.9.0.jar-extract/org/apache/flume/node/ConfigurationProvider.class
```

And this results in excluding the file from being assigned to the package: https://github.com/nexB/scancode.io/blob/f6981cf029de6276e8ec2cc5240ab3642785c2c8/scanpipe/tests/data/d2d/about_files/expected.json#L1258

If you think that's better, I can revert the existing test case to whatever it was before, and add a new test case with more files/sub-folders too. Please advise on the same.

tdruez (approved):

Issue #825, opened 2023-07-27
ABOUT files should be mapped when their documented resource is mapped


pombredanne:
ABOUT files should be mapped when their documented resource is mapped. They should be mapped to the same origin and license (e.g., the package documented in the ABOUT file) AND the .LICENSE and .NOTICE if any should also be mapped. All these assigned a status and should not be part of further scans

AyanSinhaMahapatra:
@pombredanne  I've implemented this in the commit above, but just to confirm:
> All these assigned a status and should be part of further scans
Did you mean to say they should not be scanned further?

pombredanne:
> Did you mean to say they should not be scanned further?

Yes... I fat fingered it!

> Also just detailing the changes to confirm this is what we want:
> 
>  - Files mapped by ABOUT file are now tagged with status: about-mapped
>  - Companion files for the ABOUT files (License/Notice etc) and the ABOUT file itself are also tagged with status: about-mapped
>  - Files with status about-mapped are skipped from the file scanning step at last

Actually all files mapped through an ABOUT file should be mapped super early in the pipeline and tagged such that there is NOT ANY other step that processes them further be it for mapping, matching, scanning or anything else. Their analysis is essentially "done".

tdruez:
@AyanSinhaMahapatra ready to be closed?

AyanSinhaMahapatra:
> is NOT ANY other step that processes them further be it for mapping, matching, scanning or anything else.

@pombredanne 
Most other pipes in the d2d pipeline already was doing this using `.no_status()`. For now only `map_jar_to_source` and `map_javascript` does not have filters by status, but they have filters by extension which selects files.

I've only modified the `scan_mapped_from_for_files` pipe to ignore only the files which are summarized by the ABOUT files, since that was the main issue. 

@tdruez everything mentioned in https://github.com/nexB/scancode.io/issues/825#issue-1824598825 is implemented, but should we do more based on the above comments? ^

pombredanne:
@AyanSinhaMahapatra I am letting find the best solution with @keshav-space and @tdruez if need be :)
And my point stands: things mapped with an ABOUT file **SHOULD NOT** be further analyzed... if this means changing other steps, then so be it.

keshav-space:
Ack @pombredanne , will update the `map_javascript` to ignore the files with `about file` status.

AyanSinhaMahapatra:
Closing this issue as this is completed.

PR #837, opened 2023-08-01
Tag about files and companions correctly #825

AyanSinhaMahapatra:
Tag about files and LICENSE/NOTICE files accompanying the ABOUT file as "about-mapped", and ignore these files from the scanning step.


scanpipe/models.py
```diff
@@ -1943,6 +1943,13 @@ def location_path(self):
         path = Path(str(self.path).strip("/"))
         return self.project.codebase_path / path
 
+    @property
+    def name_without_extension(self):
+        """Return the name of the resource without it's extension."""
+        if self.extension:
+            return self.name.replace(self.extension, "")
```
tdruez, line 1950: Using replace in that context is not ideal.

scanpipe/pipes/d2d.py
```diff
@@ -664,14 +673,19 @@ def map_about_files(project, logger=None):
     for about_file_resource in from_about_files:
         _map_about_file_resource(project, about_file_resource, to_resources)
 
+        about_file_companions = about_file_resource.siblings().filter(
+            name__contains=about_file_resource.name_without_extension
```
tdruez, line 677: `__startswith` may be more precise. To be confirmed.

scanpipe/pipes/d2d.py
```diff
@@ -664,14 +673,19 @@ def map_about_files(project, logger=None):
     for about_file_resource in from_about_files:
         _map_about_file_resource(project, about_file_resource, to_resources)
 
+        about_file_companions = about_file_resource.siblings().filter(
```
tdruez, line 676: Shouldn't we limit the QS to specific ABOUT files known extensions?

It seems that there's a risk of selecting unwanted files.

tdruez (changes requested):

scanpipe/models.py
AyanSinhaMahapatra: Using `.rstrip()` instead of `replace` now, that should be more precise?
Thanks!

scanpipe/pipes/d2d.py
AyanSinhaMahapatra: Yes, this would be more precise, adding this. Thanks!

scanpipe/pipes/d2d.py
AyanSinhaMahapatra: Yes, added another filter with known extensions. Thanks!

scanpipe/models.py
tdruez: `.rstrip` is dangerous:
```
>>> "file.x.ext".rstrip(".ext")
'fil'
```

AyanSinhaMahapatra:
Oh I'm sorry, I thought the commits in SCIO was always one per PR, so I was doing that. I should instead preserve commit history and after approving are the commits merged into one commit?

tdruez:
@AyanSinhaMahapatra we always want to preserve full commit history. It's not possible to review incremental changes without it.
All the branch commits are squashed and merged as 1 commit once the code is approved.

scanpipe/models.py
AyanSinhaMahapatra: Yes, using `.rpartition()` instead now. This should be safe.

AyanSinhaMahapatra:
> we always want to preserve full commit history. It's not possible to review incremental changes without it.
All the branch commits are squashed and merged as 1 commit once the code is approved.

I've restored the old commit and added the changes in a separate commit. Will do the same in future. Thanks for the heads-up

scanpipe/models.py
```diff
@@ -1943,6 +1943,14 @@ def location_path(self):
         path = Path(str(self.path).strip("/"))
         return self.project.codebase_path / path
 
+    @property
+    def name_without_extension(self):
+        """Return the name of the resource without it's extension."""
+        if self.extension:
+            name_wout_ext, ext, _ = self.name.rpartition(self.extension)
+            return name_wout_ext
```
tdruez, line 1951:
```suggestion
            return self.name.rpartition(self.extension)[0]
```

tdruez (changes requested):

scanpipe/models.py
AyanSinhaMahapatra: Added this, much cleaner. Thanks!

tdruez:
"AyanSinhaMahapatra force-pushed the update-about-mapping branch from c1c2d1d to dae972c 1 minute ago"

Is the force-push still needed?

AyanSinhaMahapatra:
> Is the force-push still needed?

It was a really minor change, and the last commit that I force-pushed is still really small to be able to review changes without any issues IMHO. The main commit is still intact, I didn't force push both.
I thought the whole point about not force-pushing commits with larger set of changes was to aid review, if instead the issue is to never force push, that's ok too. I can do new commits for every tiny change/test-fix too if that's preferred by you.

tdruez:
@AyanSinhaMahapatra Let's avoid rewriting history as a general rule.
Thanks for the PR 🚀

AyanSinhaMahapatra:
> Let's avoid rewriting history as a general rule.

That makes sense. Will adhere to that in future PRs.
Thanks for the improvement suggestions. :smile:

PR #840, opened 2023-08-01
fix: update rootfs to scan resource path

philcali:
When uploading an exported Linux rootfs archive, the current rootfs included the extracted root directory, which broke linking files from installed packages, (ie: rpms.sqlite). This change fixes that.


pombredanne:
Thanks! 
Do you mind you amend your commit with a DCS signoff?

philcali:
Will do.

pombredanne (changes requested):
This is looking great!
Thanks.
I submitted https://github.com/philcali/scancode.io/pull/1 so you get the latest main and we would need a simple CHANGELOG entry here to merge this PR

philcali:
Thank you! Changelog entry was added.

Issue #827, opened 2023-07-27
Map archives when their extracted directory mapped/processed


pombredanne:
In a deploy_to_devel pipeline, when I have an archive like "foo.zip", there will be a directory "foo.zip.extract" with the extracted content.

- If "foo.zip" is matched to the purlDB then "foo.zip.extract" should be treated as matched too. This is already the case.

- Otherwise, if not matched to the devel side and not matched to the purldb, "foo.zip" should be assigned an "extracted" status right at the end of the step that is matching archives to the purldb and this archive should not be further processed. This works because we have its extracted content that is processed otherwise.

We need to validate that:
- all the "is_archive" files are extracted. These files not extracted are possible errors.
- we are trying to match every "is_archive" file. This need changing and is tracked

tdruez:
@pombredanne Some notes looking at the data on a ScanCode.io instance:
1. All extracted resources are properly flagged as `is_archive`.
2. **NOT** all resources flagged as `is_archive` are extracted.

Resources with the following extensions are flagged as `is_archive` but not extracted by `extractcode`:
```
- .tar
- .a
- .img
- .aifc
- .lz4
- .xz
- .exe
- .deb
- .beam
- .sym
- .bz2
- .swf
- .egg
- .svgz
- .res
```

------------

Keeping the scripts to revisit this later:

```
from scanpipe.models import *

# We want to exclude filed run and run without actual local extraction)
run_qs = Run.objects.succeed().filter(pipeline_name__in=["deploy_to_develop", "docker", "scan_codebase"])
project_qs = Project.objects.filter(runs__in=run_qs)
resource_qs = CodebaseResource.objects.filter(project__in=project_qs)

extracted_root_dirs = resource_qs.filter(name__endswith="-extract")
extracted_root_dirs.count()
# 9,273
is_archive_resources = resource_qs.filter(is_archive=True)
is_archive_resources.count()
# 19,185

archives_paths = [
    extracted_dir_resource.path.removesuffix("-extract")
    for extracted_dir_resource in extracted_root_dirs
]

archives = resource_qs.filter(path__in=archives_paths)
print([resource.path for resource in archives if not resource.is_archive])
# [] -> 1. All extracted resources are properly flagged as `is_archive`

# -> 2. Let's see if all resources flagged as `is_archive` are extracted?

unextracted_extensions = set()
for resource in is_archive_resources:
    try:
        _ = resource_qs.get(
            project=resource.project,
            path=f"{resource.path}-extract",
[...]

PR #861, opened 2023-08-15
Assign status to processed archives

AyanSinhaMahapatra:
Reference: https://github.com/nexB/scancode.io/issues/827

### DIFF
--- a/scanpipe/pipes/d2d.py
+++ b/scanpipe/pipes/d2d.py
@@ -953,3 +953,32 @@ def _map_javascript_colocation_resource(
         from_neighboring_resources,
         map_type="js_colocation",
     )
+
+
+def flag_processed_archives(project):
+    """
+    Resources without an assigned status which are package archives, and all
+    resources inside the archive has a status, should also be considered as
+    processed.
+    """
+    to_resources = project.codebaseresources.files().to_codebase()
+    to_resources_archives = to_resources.no_status().filter(is_archive=True)
+
+    for to_archive in to_resources_archives:
+        archive_extract_path = to_archive.path + "-extract"
+        archive_extract_resource = to_resources.filter(path=archive_extract_path)
+
+        # TODO:There are archives which are not extracted by default, these need to
+        # be accounted for. See https://github.com/nexB/scancode.io/issues/827
+        if not archive_extract_resource:
+            continue
+
+        archive_resources_unmapped = to_resources.no_status().filter(
+            path__startswith=archive_extract_path
+        )
+        # If there are resources in the archives which are unmapped,
+        # they are not considered as processed
+        if archive_resources_unmapped:
+            continue
+
+        to_archive.update(status=flag.ARCHIVE_PROCESSED)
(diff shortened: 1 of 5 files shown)


scanpipe/pipes/d2d.py
```diff
@@ -953,3 +953,32 @@ def _map_javascript_colocation_resource(
...
+    to_resources_archives = to_resources.no_status().filter(is_archive=True)
+
+    for to_archive in to_resources_archives:
+        archive_extract_path = to_archive.path + "-extract"
+        archive_extract_resource = to_resources.filter(path=archive_extract_path)
+
+        # TODO:There are archives which are not extracted by default, these need to
+        # be accounted for. See https://github.com/nexB/scancode.io/issues/827
+        if not archive_extract_resource:
+            continue
```
tdruez, line 974: Let's not leave TODOs in the code, those never get addressed. We want issues instead, but here it seems that we have one.

tdruez: Using `archive_extract_resource.exists()` would result in a more efficient query.

scanpipe/pipes/d2d.py
```diff
@@ -953,3 +953,32 @@ def _map_javascript_colocation_resource(
...
+
+        archive_resources_unmapped = to_resources.no_status().filter(
+            path__startswith=archive_extract_path
+        )
+        # If there are resources in the archives which are unmapped,
+        # they are not considered as processed
+        if archive_resources_unmapped:
+            continue
```
tdruez, line 982: I think this is actually not needed and triggers an extra query, ie it evaluates the `archive_resources_unmapped` queryset with a SELECT query then does the UPDATE query.
Where doing directly the UPDATE (that may update nothing) would always ensure only 1 query is executed.
To be confirmed though.

tdruez (changes requested):

tdruez:
@AyanSinhaMahapatra Could you also add an entry in the changelog?

scanpipe/pipes/d2d.py
AyanSinhaMahapatra: Thanks, removing the TODO there and using the exists() function.

scanpipe/pipes/d2d.py
AyanSinhaMahapatra: I'm not sure I understand what you mean correctly.
The SELECT query and the UPDATE query are on two different set of files, one on `archive_resources_unmapped` and the other on a singular `to_archive`.
The SELECT query is done on all the files in the extracted package, to check if any of them still have no status.
Then if and only there are no files which have status empty in those extracted files, we do an UPDATE on only one resource, which is the resource for the archive in question.
But I can use the `.exists()` function here too as this is same as the case mentioned in https://github.com/nexB/scancode.io/pull/861#discussion_r1295798243 to make this more efficient?

AyanSinhaMahapatra:
@tdruez thanks for the improvement suggesttions!
I've added the updates and a CHANGELOG entry, except this point: https://github.com/nexB/scancode.io/pull/861#discussion_r1295861240
Ready for your review again! :smile:

scanpipe/pipes/d2d.py
tdruez: Nevermind, I've read too fast thinking the update was applying to the previous queryset.
All good ;)

PR #876, opened 2023-08-17
Move purldb package match ahead

pombredanne:
BY @AyanSinhaMahapatra 
> Run the purldb matching step to match package archives before the java/javascript mapping steps in the d2d pipeline.


pombredanne:
@AyanSinhaMahapatra we need rebased so it can be merge in the integration branch for testing?

tdruez:
@AyanSinhaMahapatra could you look into what needs to be done and merge on this one?

AyanSinhaMahapatra:
I think the https://github.com/nexB/scancode.io/pull/876/commits/7f4b1cf75d462d494be96374b25809058f2ea598 moving purldb package matching step is the only extra commit here, and rest of the commits in the integration branch has been merged seperately. And this commit was being tested for the jar matching bug, but it turned out to be an issue in the pagination, so I don't think this is required anymore. So we can safely close this PR IMHO

Issue #872, opened 2023-08-17
too-many-maps problem


pombredanne:
We should not ever consider a single matched path segment/file name as the source for a "too-many-maps" status.

keshav-space:
fixed in https://github.com/nexB/scancode.io/pull/882

PR #882, opened 2023-08-21
Avoid path map solely based on file name

keshav-space:
- fixes https://github.com/nexB/scancode.io/issues/872

### DIFF
--- a/scanpipe/pipes/d2d.py
+++ b/scanpipe/pipes/d2d.py
@@ -418,13 +418,14 @@ def _map_path_resource(
     if not match:
         return
 
+    # Don't path map resource solely based on the file name.
+    if match.matched_path_length < 2:
+        return
+
     # Only create relations when the number of matches if inferior or equal to
     # the current number of path segment matched.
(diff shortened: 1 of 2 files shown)


scanpipe/pipes/d2d.py
```diff
@@ -421,7 +421,10 @@ def _map_path_resource(
     # Only create relations when the number of matches if inferior or equal to
     # the current number of path segment matched.
     if len(match.resource_ids) > match.matched_path_length:
-        to_resource.update(status=flag.TOO_MANY_MAPS)
+        # Single matched path segment/file name don't
```
pombredanne, line 424: IMHO we should go further and not store any mapping at all in this case

keshav-space: Ack, So we don't want to path map any files solely based on the file name.

tdruez:
@keshav-space is this ready to be merged?

keshav-space:
> @keshav-space is this ready to be merged?

@tdruez this is ready for merge!

PR #891, opened 2023-08-22
Add errors if ABOUT file resource not found

AyanSinhaMahapatra:


tdruez (changes requested):
@AyanSinhaMahapatra Since the errors are localized on specific objects and not global for the pipeline, I would suggest using the `project.add_error(self, error, model, details=None)` API when the error is encountered instead of propagating back to the pipeline.
This would allow us to set the `details` field with the relevant data to ensure we can act on the error later. 

Also, instead of "ERROR: at map_about_files: ", I would remove the "ERROR" as its redundant and I would set the `Error.model` field to `map_about_files` so it can be use to group those errors and filter by this value.

PR #894, opened 2023-08-25
Document rosetta support

Hritik14:
Some of us are using mac and would benefit from a faster development environment setup.


tdruez:
@Hritik14 Thanks!

PR #904, opened 2023-09-01
Modify pipelines to get purls from package_data

AyanSinhaMahapatra:
The `scan_codebase_packages` pipeline currently takes a lot of time for large projects with ~700k resources, and it is stuck in the `scan_for_application_packages` step, and so presumably it's the package assembly in there that takes the longest.
The main objective of this pipeline was as a precursor to the `populate_purldb` pipeline to get purls from the codebase and submit them to purldb for indexing. So this PR does the following:

1. Disables the package assembly for the `scan_codebase_packages` pipeline. So we only have the codebase scanned for package_data and these are added to the specific resources without creating DiscoveredPackage and DiscoveredDependency objects out of these package data, by assembling.
2. If there are no packages/dependencies added to the project (otherwise we will add purls from these packages/deps twice), then further look for package data detected in the resources and add purls from those.

### DIFF
--- a/scanpipe/pipelines/populate_purldb.py
+++ b/scanpipe/pipelines/populate_purldb.py
@@ -52,10 +52,13 @@ class PopulatePurlDB(Pipeline):
 
     def populate_purldb_with_detected_purls(self):
         """Add DiscoveredPackage to PurlDB."""
-        if (
-            not self.project.discoveredpackages.exists()
-            and not self.project.discovereddependencies.exists()
-        ):
+        no_packages_and_no_dependencies = all(
+            [
+                not self.project.discoveredpackages.exists(),
+                not self.project.discovereddependencies.exists(),
+            ]
+        )
+        if no_packages_and_no_dependencies:
             packages = scancode.get_packages_with_purl_from_resources(self.project)
             self.feed_purldb(
                 packages=list(packages),
--- a/scanpipe/pipelines/populate_purldb.py
+++ b/scanpipe/pipelines/populate_purldb.py
@@ -52,11 +52,15 @@ class PopulatePurlDB(Pipeline):
 
     def populate_purldb_with_detected_purls(self):
         """Add DiscoveredPackage to PurlDB."""
-        packages = scancode.get_packages_with_purl_from_resources(self.project)
-        self.feed_purldb(
-            packages=list(packages),
-            package_type="DiscoveredPackage",
-        )
+        if (
+            not self.project.discoveredpackages.exists()
+            and not self.project.discovereddependencies.exists()
+        ):
+            packages = scancode.get_packages_with_purl_from_resources(self.project)
+            self.feed_purldb(
+                packages=list(packages),
+                package_type="DiscoveredPackage",
+            )
 
     def feed_purldb(self, packages, package_type):
         """Feed PurlDB with list of PURLs for indexing."""
--- a/scanpipe/pipelines/populate_purldb.py
+++ b/scanpipe/pipelines/populate_purldb.py
@@ -58,6 +58,10 @@ class PopulatePurlDB(Pipeline):
                 not self.project.discovereddependencies.exists(),
             ]
         )
+        # Even when there are no packages/dependencies, resource level
+        # package data could be detected (i.e. when we detect packages,
+        # but skip the assembly step that creates
+        # package/dependency instances)
         if no_packages_and_no_dependencies:
             packages = scancode.get_packages_with_purl_from_resources(self.project)
             self.feed_purldb(
--- a/scanpipe/pipelines/populate_purldb.py
+++ b/scanpipe/pipelines/populate_purldb.py
@@ -22,6 +22,7 @@
 
 from scanpipe.pipelines import Pipeline
 from scanpipe.pipes import purldb
+from scanpipe.pipes import scancode
 
 
 class PopulatePurlDB(Pipeline):
@@ -32,6 +33,7 @@ class PopulatePurlDB(Pipeline):
         return (
             cls.populate_purldb_with_discovered_packages,
             cls.populate_purldb_with_discovered_dependencies,
+            cls.populate_purldb_with_detected_purls,
         )
 
     def populate_purldb_with_discovered_packages(self):
@@ -46,6 +48,14 @@ class PopulatePurlDB(Pipeline):
         self.feed_purldb(
             packages=self.project.discovereddependencies.all(),
             package_type="DiscoveredDependency",
+        )
+
+    def populate_purldb_with_detected_purls(self):
+        """Add DiscoveredPackage to PurlDB."""
+        packages = scancode.get_packages_with_purl_from_resources(self.project)
+        self.feed_purldb(
+            packages=list(packages),
+            package_type="DiscoveredPackage",
         )
 
     def feed_purldb(self, packages, package_type):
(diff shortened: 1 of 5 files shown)


scanpipe/pipes/scancode.py
```diff
@@ -430,6 +431,20 @@ def assemble_packages(project):
                     logger.info(f"Unknown Package assembly item type: {item!r}")
 
 
+def get_packages_with_purl_from_resources(project):
```
tdruez, line 434: Missing a docstring and unit tests.

scanpipe/pipelines/populate_purldb.py
```diff
@@ -48,6 +50,18 @@ def populate_purldb_with_discovered_dependencies(self):
             package_type="DiscoveredDependency",
         )
 
+    def populate_purldb_with_detected_purls(self):
+        """Add DiscoveredPackage to PurlDB."""
+        if (
+            not self.project.discoveredpackages.exists()
+            and not self.project.discovereddependencies.exists()
+        ):
```
tdruez, line 58: You can wrap the condition into a explicit variable to help the readability.
```suggestion
        no_packages_and_no_dependencies = all([
            not self.project.discoveredpackages.exists(),
            not self.project.discovereddependencies.exists(),
        ])
        
        if no_packages_and_no_dependencies:
```

scanpipe/pipelines/populate_purldb.py
```diff
@@ -48,6 +50,18 @@ def populate_purldb_with_discovered_dependencies(self):
             package_type="DiscoveredDependency",
         )
 
+    def populate_purldb_with_detected_purls(self):
```
tdruez, line 53: It's unclear why this is only called when the project has not packages/dependencies.

tdruez (changes requested):
See various comments.

Also: 
> Disables the package assembly for the scan_codebase_packages pipeline. So we only have the codebase scanned for package_data and these are added to the specific resources without creating DiscoveredPackage and DiscoveredDependency objects out of these package data, by assembling.

I'm not sure about changing the behavior of package detection in `scan_codebase_packages`. It's unclear to me looking at the code the consequence of this.

scanpipe/pipelines/scan_codebase_packages.py
```diff
@@ -36,3 +37,7 @@ def steps(cls):
             cls.flag_ignored_resources,
             cls.scan_for_application_packages,
         )
+
+    def scan_for_application_packages(self):
+        """Scan unknown resources for packages information."""
+        scancode.scan_for_application_packages(self.project, assemble=False)
```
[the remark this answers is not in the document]
AyanSinhaMahapatra: Added this, thanks!

scanpipe/pipes/scancode.py
```diff
@@ -338,7 +338,7 @@ def scan_for_files(project, resource_qs=None):
     _scan_and_save(resource_qs, scan_file, save_scan_file_results, scan_func_kwargs)
 
 
-def scan_for_application_packages(project):
+def scan_for_application_packages(project, assemble=True):
     """
     Run a package scan on files without a status for a `project`,
     then create DiscoveredPackage and DiscoveredDependency instances
```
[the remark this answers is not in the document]
AyanSinhaMahapatra: Added some details on this, please review if this makes sense now. Thanks!

scanpipe/pipes/scancode.py
AyanSinhaMahapatra: Added a docstring and a test, thanks!

scanpipe/pipelines/populate_purldb.py
AyanSinhaMahapatra: Added this, thanks!

scanpipe/pipelines/populate_purldb.py
AyanSinhaMahapatra: This step exists to get purls out of package_data saved at resources. These are still detected and saved, even though we don't have Package/Dependency instances (the assemble=False skips the package assembly, but resource level package detection is still enabled and happening).

AyanSinhaMahapatra:
> I'm not sure about changing the behavior of package detection in scan_codebase_packages. It's unclear to me looking at the code the consequence of this.

If the main objective of this pipeline is to get purls from the codebase and use this as a previous step for the populate_purldb pipeline, then we need to create Discoveredpackage/DiscoveredDependency instances, but having resource level package detections in each resource is enough (We can goa  step further by passing an argument to scancode to only detect purl related information only, and ignore everything else, which will be the next step). The package assembly step uses these resource level detections and creates instances, assigns files, but this is not required in this context. 

@tdruez thanks for the review and feedback, I've addressed these, ready for your review again :smile:

tdruez (changes requested):
@AyanSinhaMahapatra It would be nice to add some details in the docstring of ScanCodebasePackages to describe its purpose and its limitations.

scanpipe/pipelines/populate_purldb.py
AyanSinhaMahapatra: @tdruez I've added this in the code as comments, thanks!

scanpipe/pipelines/populate_purldb.py
```diff
@@ -70,7 +70,7 @@ def feed_purldb(self, packages, package_type):
         if not purldb.is_available():
             raise Exception("PurlDB is not available.")
 
-        package_urls = [package.purl for package in packages]
+        package_urls = list(set([package.purl for package in packages]))
```
AyanSinhaMahapatra, line 77: Note that when we get purls from package_data and not package/dependency instances, we might have duplicate purls (like from manifests of a same package instance and from different manifests having same dependencies), so it makes sense to send only the unique purls for indexing in purldb.

scanpipe/pipes/scancode.py
```diff
@@ -436,6 +442,25 @@ def assemble_packages(project):
...
+        for package_mapping in resource.package_data:
+            for dep in package_mapping.get("dependencies"):
+                yield packagedcode_models.Dependency.from_dependent_package(
+                    dependent_package=dep,
+                    datafile_path=resource.path,
+                    datasource_id=package_mapping.get("datasource_id"),
+                    package_uid=None,
+                )
+            pd = packagedcode_models.PackageData.from_dict(mapping=package_mapping)
+            yield pd
```
tdruez, line 461: Assigning to a var is not useful here unless the var name helps with the readability, which is not the case.
```suggestion
            yield packagedcode_models.PackageData.from_dict(mapping=package_mapping)
```

scanpipe/pipes/scancode.py
```diff
@@ -436,6 +442,25 @@ def assemble_packages(project):
...
+    """
+    Yield Dependency or PackageData objects created from detected package_data
+    in all the project resources. Both Dependency and PackageData objects have
+    the `purl` attribute with a valid purl.
+    """
+    for resource in project.codebaseresources.has_package_data():
+        for package_mapping in resource.package_data:
+            for dep in package_mapping.get("dependencies"):
+                yield packagedcode_models.Dependency.from_dependent_package(
+                    dependent_package=dep,
```
tdruez, line 455:
```suggestion
            for dependency in package_mapping.get("dependencies"):
                yield packagedcode_models.Dependency.from_dependent_package(
                    dependent_package=dependency,
```

tdruez (changes requested):
2 more minor suggestions and we'll be ready ;)

scanpipe/pipes/scancode.py
AyanSinhaMahapatra: Done, thanks!

scanpipe/pipes/scancode.py
AyanSinhaMahapatra: Done, thanks!

AyanSinhaMahapatra:
@tdruez latest comments addressed, thanks for the suggestions. :smile:

Issue #914, opened 2023-09-07
Group non-package files together for analysis and reporting purposes


DennisClark:
Scan results are currently focused on packages, but in a typical product codebase there are lots of files that are not (or no longer) associated with packages, and these files may be very important from a license compliance perspective.  Some cases include: 
* first-party files that could be reported together as belonging to the product copyright holder under the product license expression
* files that have been extracted from third-party archives and used individually, often containing copyright or license statements or both

These miscellaneous files are currently difficult to analyze in the product codebase scan results, especially when there are lots of them, and they are not always being included in attribution and SBOM outputs. 

Consider grouping these miscellaneous files into "local" packages (or "custom", "virtual", "logical", "file_set" packages), perhaps combining them by license_expression.  It may also make sense to define these using the PURL convention; consider: 

Type: "local"  (or some other descriptive label that distinguishes them from standard packages) 
Namespace: the name of the product codebase 
Name: the common license_expression for the group of files
Version: the version of the product codebase
Qualifiers: optional, most likely null
Subpath: optional, most likely null 

The primary advantage of creating "local" packages in the scan results are: 
* organize the data for review and analysis, where the assigned PURL makes it clear that these "packages" are a special case
* take advantage of the existing ScanCode.io UI 
* take advantage of existing code that generates attribution and SBOMs  

I prefer the term "local" for the PURL Type value, but this is of course open to discussion.

DennisClark:
@tdruez    I guess that we would have to use the scancode.io project name for the Name, and we don't have any value for the version field.

DennisClark:
but the best solution would be to add the concept of Product (product name + product version) to SCIO to improve integration, ultimately, with DejaCode.

mjherzog:
In the big picture we need to be sure that we fully support reporting of first-party code, especially first-party code that has a copyright or license notice.

DennisClark:
alternative value for "Name":   a UUID

DennisClark:
Let's go with this:
Type: "local-files"
Namespace: a generated UUID
Name: the common license_expression for the group of files

the rest of the PURL is probably null

DennisClark:
Let's go with this:
Type: "local-files"
Namespace: the SCIO project as a **slug**
Name: generated UUID

the rest of the PURL is probably null

tdruez:
PR https://github.com/nexB/scancode.io/pull/927 merged in main.

The "local-files" packages are now created as part of the d2d pipeline.

[...]

PR #927, opened 2023-09-15
Base implementation for local-files packages creation #914

tdruez:


scanpipe/pipes/d2d.py
```diff
@@ -24,9 +24,11 @@
 from contextlib import suppress
 from pathlib import Path
 
+from django.contrib.postgres.aggregates.general import ArrayAgg
```
pombredanne, line 27: TIL about ArrayAgg! thanks!

scanpipe/pipes/d2d.py
```diff
@@ -1109,3 +1111,45 @@ def _map_thirdparty_npm_packages(package_json, to_resources, project):
...
+    )
+
+    grouped_by_license = grouped_by_license.annotate(
+        grouped_resource_ids=ArrayAgg("id", distinct=True),
+        grouped_copyrights=ArrayAgg("copyrights", distinct=True),
+    )
+
+    for group in grouped_by_license:
+        codebase_resource_ids = set(group["grouped_resource_ids"])
+        copyrights = set(
```
pombredanne, line 1144: IMHO we should use a `Counter` instead and then report copyrights sorted in order of most frequent down. This could be a second step though

scanpipe/pipes/d2d.py
```diff
@@ -1109,3 +1111,45 @@ def _map_thirdparty_npm_packages(package_json, to_resources, project):
...
+            [
+                entry["copyright"]
+                for copyrights in group["grouped_copyrights"]
+                for entry in copyrights
+            ]
+        )
+        defaults = {
+            "declared_license_expression": group.get("detected_license_expression"),
+            "copyright": "\n\n".join(copyrights),
```
pombredanne, line 1153: At the minimum, sort your set to get stable results

pombredanne (approved):
LGTM... see a few nits inlined for possible improvements.

scanpipe/pipes/d2d.py
tdruez: TIL about TIL 😄

scanpipe/pipes/d2d.py
tdruez: Counter implemented, thanks for the suggestion.

Issue #931, opened 2023-09-21
D2D: Scaned `from` files without detected license expressions should be tagged for review


keshav-space:
- [x] Scanned media files (gif/ jpg/ jpeg/ png/ svg/ mp4) without detected license expressions should be tagged as `ignored-media-file`
- [x] Scanned doc files like pdf without detected license expressions should be tagged as `ignored-doc-file`
~~- [ ] Scanned schema files like module-log4j.xml, service.xml, module-hbm.xml, default.xml, web.xml, routes.xml without detected license expressions should be tagged as `ignored-not-interesting`~~
- [x] Rest all other scanned from files without detected license expressions should be tagged as `requires-review`

pombredanne:
>  Scanned schema files like module-log4j.xml, service.xml, module-hbm.xml, default.xml, web.xml, routes.xml without detected license expressions should be tagged as ignored-not-interesting

These should require review too. We want to ensure we have affirmative origin and license on these

PR #932, opened 2023-09-22
Update status for deployed files without license

keshav-space:
- fixes https://github.com/nexB/scancode.io/issues/931

### DIFF
--- a/scanpipe/pipes/d2d.py
+++ b/scanpipe/pipes/d2d.py
@@ -1285,9 +1285,11 @@ def scan_unmapped_to_files(project, logger=None):
     )
 
 
-def flag_deployed_resources_with_missing_license(project, doc_extensions=None):
-    """Update the status for deployed files with missing license."""
-    scan_files = (
+def flag_deployed_from_resources_with_missing_license(project, doc_extensions=None):
+    """Update the status for deployed from files with missing license."""
+    # Retrieve scanned from files with an empty ``detected_license_expression``
+    # or a ``unknown`` license expression.
+    scanned_from_files = (
         project.codebaseresources.files()
         .from_codebase()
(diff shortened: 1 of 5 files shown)


scanpipe/pipes/d2d.py
```diff
@@ -1283,3 +1283,25 @@ def scan_unmapped_to_files(project, logger=None):
     project.codebaseresources.files().to_codebase().filter(status=flag.SCANNED).update(
         status=flag.REQUIRES_REVIEW
     )
+
+
+def flag_deployed_resources_with_missing_license(project, doc_extensions=None):
+    """Update the status for deployed files with missing license."""
```
tdruez, line 1289: The docstring and function name do not reflect properly what I'm seeing in the function code.
"for deployed files" -> It's a bit confusing since the QS works on "from/" files.

scanpipe/pipes/d2d.py
```diff
@@ -1283,3 +1283,25 @@ def scan_unmapped_to_files(project, logger=None):
...
+    scan_files = (
```
tdruez, line 1290:
```suggestion
    scanned_from_files = (
```

tdruez (changes requested):

scanpipe/pipes/d2d.py
keshav-space: These are from files that are deployed on `to` side.
I guess `flag_deployed_from_resources_with_missing_license` will be more appropriate function name.

tdruez: This is not explicit looking at the code, a comment about it would help the readability ;)

keshav-space: ack

Issue #929, opened 2023-09-18
Issue with CodebaseResourceQuerySet.paginated


JonoYang:
I have found an issue with `CodebaseResourceQuerySet.paginated` when debugging an issue in the d2d pipeline at the purldb matching step, where it appears that certain JARs are not being matched. The unmatched JAR in particular is `jgroups-4.1.7.Final.jar`. I put a print statement in the nested for-loop at in `match_purldb_resources` (https://github.com/nexB/scancode.io/blob/main/scanpipe/pipes/d2d.py#L604) and I see that `jgroups-4.1.7.Final.jar` is not in any `resources_batch` yielded by `resource_iterator`. If I create a d2d project with only `jgroups-4.1.7.Final.jar`, then we are able to match the JAR against purldb. In my debug branch, I've removed the use of `CodebaseResourceQuerySet.paginated` from `match_purldb_resources` and to just iterate through the resources without batching (https://github.com/nexB/scancode.io/blob/integration-purldb-matching-js-d2d-debug/scanpipe/pipes/d2d.py#L712). I was able to send `jgroups-4.1.7.Final.jar` for matching and get a result back.

I think the issue with `CodebaseResourceQuerySet.paginated` is that it is not returning the remaining CodebaseResources that does not fit the chunk size. In the case above, we had 2300 CodebaseResources to iterate through, but the logging showed that we sent about 1800 SHA1's  to purldb. The 2300 CodebaseResources were processed in batches of 1000, where we sent off around ~850-900 SHA1s during each batch, but during the last batch, no CodebaseResources were sent for the remaining ~300.

tdruez:
Paginator is probably not the right tool for the job.

Let's take a step back, what was the problem with using [`iterator()`](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#iterator) in the first place?

> In contrast, iterator() will read results directly, without doing any caching at the QuerySet level

I would like to get a better understanding of the issues while using `iterator()`, especially since we are using it in many other places, and see how we could use it properly instead of Pagination.

JonoYang:
@tdruez I used the paginator originally as a way to address the comment at https://github.com/nexB/scancode.io/pull/782#discussion_r1286716702, where you suggested to use `chunked` for batch processing of CodebaseResources. `chunked` works on lists, and not querysets. I would have to evaluate the queryset into a list to use it, which can be a problem for large queries, so I decided to try out using the paginator.

tdruez:
[...]

PR #934, opened 2023-09-22
D2d pipeline updates

JonoYang:
This PR contains changes to the D2D pipeline and the purldb matching step.

There is workaround for #929, where we process `to_resources` using `.iterator()` instead of `.paginated()`. We now keep track of the number of to/ resources we're iterating through, so we can send off a match requests of `chunk_size`. After we iterate through `to_resources`, if `resources_by_sha1` has sha1s and Resources remaining, we send those sha1s off to be matched.

The `match_purldb` step in the D2D pipeline has now been split up into two steps, `match_archives_to_purldb` and `match_resources_to_purldb`.

### DIFF
--- a/scanpipe/models.py
+++ b/scanpipe/models.py
@@ -1674,12 +1674,15 @@ class CodebaseResourceQuerySet(ProjectRelatedQuerySet):
             return self.filter(status=status)
         return self.filter(~Q(status=""))
 
-    def not_status(self, status=None):
+    def no_status(self, status=None):
+        """
+        Filter for CodebaseResources without a status.
+
+        If `status` is provided, then we filter for CodebaseResources whose
+        status is not equal to `status`.
+        """
         if status:
             return self.filter(~Q(status=status))
-        return self.filter(status="")
-
-    def no_status(self):
         return self.filter(status="")
 
     def empty(self):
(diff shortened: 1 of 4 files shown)


scanpipe/models.py
```diff
@@ -1674,6 +1674,11 @@ def status(self, status=None):
             return self.filter(status=status)
         return self.filter(~Q(status=""))
 
+    def not_status(self, status=None):
+        if status:
+            return self.filter(~Q(status=status))
+        return self.filter(status="")
+
     def no_status(self):
         return self.filter(status="")
```
tdruez, line 1684: Could we merge both functions to avoid redundancy? Maybe add the status option to the existing `no_status`.

scanpipe/pipes/d2d.py
```diff
@@ -593,35 +613,44 @@ def match_purldb_resources(
                 f"as there are {resource_count:,d}"
             )
 
-    resource_iterator = to_resources.paginated(per_page=chunk_size)
+    resource_iterator = to_resources.iterator()
```
tdruez, line 616: We should be explicit about the `chunk_size` value.

tdruez (changes requested):
Let's also remove the `CodebaseResourceQuerySet.paginated` method as we don't want to encourage future use of it.

scanpipe/pipes/d2d.py
JonoYang: @tdruez I've set `chunk_size` argument on `to_resources.iterator()` to be the same as `chunk_size` from the function argument, which has a default value of 1000. Should `chunk_size` on the iterator be larger?

PR #952, opened 2023-09-29
Limit npm package lookup to resource without status

keshav-space:


scanpipe/pipes/d2d.py
```diff
@@ -1164,7 +1164,7 @@ def _map_thirdparty_npm_packages(package_json, to_resources, project):
 
     package_resources = to_resources.filter(path__startswith=path_parent)
 
-    if not all(
+    if not package_resources or not all(
         [package, package.type, package.name, package.version, package_resources]
     ):
```
tdruez, line 1169: It's usually a good idea to put thos condition into a variable properly named to help with readability
```
all(
        [package, package.type, package.name, package.version, package_resources]
    )
```

keshav-space: Done :)

PR #954, opened 2023-10-02
Update d2d pipeline to use specific license flags

AyanSinhaMahapatra:
We were using REQUIRES-REVIEW for both no licenses and unknown licenses in the from side, this adds already existing specific flags and this makes the review flag to be less over-populated.

### DIFF
--- a/scanpipe/pipes/d2d.py
+++ b/scanpipe/pipes/d2d.py
@@ -1338,10 +1338,8 @@ def flag_deployed_from_resources_with_missing_license(project, doc_extensions=No
             status=flag.IGNORED_DOC_FILE
         )
 
-    no_license_files = scanned_from_files.filter(Q(detected_license_expression=""))
-    unknown_license_files = scanned_from_files.filter(
-        Q(detected_license_expression__icontains="unknown")
-    )
+    no_license_files = scanned_from_files.filter(detected_license_expression="")
+    unknown_license_files = scanned_from_files.unknown_license()
 
     no_license_files.update(status=flag.NO_LICENSES)
     unknown_license_files.update(status=flag.UNKNOWN_LICENSE)


pombredanne (approved):
LGTM... in the future we may wan to have a common prefix for the statuses like in "requires-review-no-license"

scanpipe/pipes/d2d.py
```diff
@@ -1344,4 +1338,10 @@ def flag_deployed_from_resources_with_missing_license(project, doc_extensions=No
             status=flag.IGNORED_DOC_FILE
         )
 
-    scanned_from_files.update(status=flag.REQUIRES_REVIEW)
+    no_license_files = scanned_from_files.filter(Q(detected_license_expression=""))
+    unknown_license_files = scanned_from_files.filter(
+        Q(detected_license_expression__icontains="unknown")
+    )
```
tdruez, line 1344: We should use the available QuerySet  `unknown_license()` methods.

scanpipe/pipes/d2d.py
```diff
@@ -1344,4 +1338,10 @@ def flag_deployed_from_resources_with_missing_license(project, doc_extensions=No
             status=flag.IGNORED_DOC_FILE
         )
 
-    scanned_from_files.update(status=flag.REQUIRES_REVIEW)
+    no_license_files = scanned_from_files.filter(Q(detected_license_expression=""))
```
tdruez, line 1341: The `Q` object is not needed here.

tdruez (changes requested):

scanpipe/pipes/d2d.py
AyanSinhaMahapatra: Added, thanks!

scanpipe/pipes/d2d.py
AyanSinhaMahapatra: Used this now, thanks!

Issue #941, opened 2023-09-26
Do not list all the packages in resource list view


keshav-space:
A resource can belong to multiple packages so we should truncate the list of packages in the resource list view, something similar to what we do in package list view.

PR #955, opened 2023-10-03
Display only 5 packages per resource in list view

keshav-space:
- fixes https://github.com/nexB/scancode.io/issues/941

### DIFF
--- a/scanpipe/templates/scanpipe/resource_list.html
+++ b/scanpipe/templates/scanpipe/resource_list.html
@@ -72,9 +72,9 @@
                   </li>
                 {% endfor %}
               </ul>
-              {% if resource.discovered_packages.all.count > 5 %}
+              {% if resource.discovered_packages.all|length > 5 %}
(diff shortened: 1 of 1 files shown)


scanpipe/templates/scanpipe/resource_list.html
```diff
@@ -65,12 +66,17 @@
...
                   <li>
                     <a href="{% url 'project_packages' project.slug %}?purl={{ package.package_url }}">{{ package }}</a>
                   </li>
                 {% endfor %}
               </ul>
+              {% if resource.discovered_packages.all.count > 5 %}
+                <a href="{% url 'resource_detail' project.slug resource.path %}#packages" class="button is-small is-fullwidth">
+                    View {{ resource.discovered_packages.all.count|intcomma }} packages
+                </a>
+              {% endif %}
```
tdruez, line 79: Using `count` triggers another SQL query. Since we already evaluated the initial QS, using the length defaultfilter would be better.

tdruez (changes requested):

PR #975, opened 2023-10-19
Choose best package for PurlDB matched resources

keshav-space:
In PurlDB resource matching step we may match single resource to multiple package, this post process step insures that we keep only the best matched package for every resource and remove the rest. ATM we visit archives's extract directory and rank all the packages based on the higest number of resource matched, and only keep those resource-package relationship where package has highest resource count.


tdruez:
@keshav-space could you fix the conflicting files?

keshav-space:
> @keshav-space could you fix the conflicting files?

Done!

Issue #969, opened 2023-10-12
Considering a  file with a single LF is not useful


pombredanne:
```
$ echo "" > c
$ sha1sum c
adc83b19e793491b1c6ea0fd8b46cd9f32e592fc  c
```
there are many files with just that single LF alright and they should not even be treated seriously in a pipeline and just reported as empty. This could be based on the size being 1 byte for a start.
Otherwise we end up sending things to be matched to the PurlDB for no good reason.
See https://github.com/nexB/purldb/issues/198

pombredanne:
For the deploy_to_devel pipeline we should have a step to ignore empty files from matching.

tdruez:
Instead of looking at the content, I would suggest that we ignore files of size under a few (TBD) Bytes.

PR #976, opened 2023-10-19
Ignore whitespace files

keshav-space:
- fixes https://github.com/nexB/scancode.io/issues/969

### DIFF
--- a/scanpipe/pipelines/deploy_to_develop.py
+++ b/scanpipe/pipelines/deploy_to_develop.py
@@ -48,6 +48,7 @@ class DeployToDevelop(Pipeline):
             cls.collect_and_create_codebase_resources,
             cls.fingerprint_codebase_directories,
             cls.flag_empty_files,
+            cls.flag_whitespace_files,
             cls.flag_ignored_resources,
             cls.map_about_files,
             cls.map_checksum,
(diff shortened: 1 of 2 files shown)


scanpipe/pipelines/deploy_to_develop.py
```diff
@@ -134,6 +135,14 @@ def fingerprint_codebase_directories(self):
         """Compute directory fingerprints for matching"""
         matchcode.fingerprint_codebase_directories(self.project, to_codebase_only=True)
 
+    def flag_whitespace_files(self):
+        """
+        Flag whitespace files with size less than or equal
+        to 1 byte as ignored.
+        """
+        qs = self.project.codebaseresources.files().filter(size__lte=1)
```
tdruez, line 143: How can you guarantee the content is whitespace when the size is 1 or less?

Also, what's your take on https://github.com/nexB/scancode.io/issues/969#issuecomment-1760828076 ?
What would be a minimum size to not be flagged as ignored?

tdruez (changes requested):

scanpipe/pipelines/deploy_to_develop.py
keshav-space: > How can you guarantee the content is whitespace when the size is 1 or less?

In our use case, we found that files with whitespace characters were 1 bytes files, and it may not be true all the time.

> Also, what's your take on [#969 (comment)](https://github.com/nexB/scancode.io/issues/969#issuecomment-1760828076) ?

It would be impractical to look inside each and every file content. IMO we can filter out the files which are 1 byte or less in size and then use regex to confirm that indeed the content inside the files is just whitespace characters.


> What would be a minimum size to not be flagged as ignored?

I don't have numbers on this ATM, but I guess we are good with not ignoring anything above 1 byte, at least for now.

tdruez: > It would be impractical to look inside each and every file content.

I think you are missing the point. I'm just wondering how valuable is to scan 2 bytes files, 3 bytes files, etc...
Is there any value in analyzing files that small, are there any licenses/copyrights to be detected in 2-byte files?

>  I guess we are good with not ignoring anything above 1 byte

Alright then.

Issue #834, opened 2023-07-31
Distinguish between ABOUT files found in the codebase vs. curations


pombredanne:
In the d2d pipeline, the "curations" are ABOUT files designed to document code on the To side and are stored in the "From" side

There could also be "regular" ABOUT files in any of the codebases that would have been created by the maintainers of the repo to document code that lives side-by-side to these ABOUT files, like is done in SCTK or SCIO.

These are different!!!

pombredanne:
@AyanSinhaMahapatra ping... this needs design.

tdruez:
> These are different!!!

How do we know what type (curations/regular) we are dealing with?

AyanSinhaMahapatra:
> How do we know what type (curations/regular) we are dealing with?

From @pombredanne this could be either:

1. some different extension from .ABOUT which has the curations
2. or we introduce a new attribute in the ABOUT spec to distinguish the two

I think the later is much easier and cleaner, maybe something like an `is_curated` attribute could work. Then we can distinguish these cases.

pombredanne:
Another consideration is that a curation may be for the current directory or it could be for a target deployed file. We should have a way to track these.
An idea: 
- we use a flag `is_curation` set to yes or no
- the about_resource: should be for the current local resource (e.g. on the development From/ Side)
- we add a new path that would be the "deployed_resource" or something along these lines

pombredanne:
Here are a few more thoughts on this issue:

- https://github.com/nexB/aboutcode-toolkit/pull/539/ is a pending PR on AboutCode toolkit. It proposes to add `deployed_resource` and `is_curated` attributes to an ABOUT file
- The experimental ABOUT file format in the PR above support patterns like in this example:
```
about_resource: '*/glowroot/*'
ignored_resources:
  - '*/glowroot/plugins/*'
name: glowroot
version: 0.13.6
```

We can now include and exclude patterns using `about_resource` and `ignored_resources`.  This is not yet in AboutCode Toolkit BUT this is already used in ScanCode.io 


An ABOUT file is like a package manifest. Or it is like a mini PurlDB.

The cases we have are:

1. an ABOUT file present in the codebase designed to document the files present there. The `about_resource` field point to these existing files. This is the easy case

2. OR, an ABOUT file present in the codebase designed to document files not yet present, such as files that will be fetched or compiled or created as a result of the build. This is the case we need and use in d2d pipelines today and the more complex case.

[...]

PR #982, opened 2023-10-24
Support patterns in ABOUT resource paths

AyanSinhaMahapatra:
ABOUT files resource path fields `about_resource` and `ignored_resources` have paths which often have versions names in it as they are part of a package archive/directory and when these packages are updated to a new version, the ABOUT files don't match to these packages anymore, and the version has to be updated in the ABOUT file which is an inconvenience.
Instead it's better to have regex expressions for these paths so they are matched more broadly without needing frequent updates.


scanpipe/models.py
```diff
@@ -1779,6 +1779,15 @@ def path_pattern(self, pattern):
...
         return self.filter(path__regex=posix_regex_to_django_regex_lookup(pattern))
 
+    def path_patterns(self, patterns):
+        """Resources with a path that match the provided ``pattern``."""
+        lookups = Q()
+        for resource_pattern in patterns:
+            lookups |= Q(
+                **{"path__regex": posix_regex_to_django_regex_lookup(resource_pattern)}
+            )
+        return self.filter(~lookups)
```
pombredanne, line 1789: Is this tilde a negation?

scanpipe/pipes/d2d.py
```diff
@@ -774,94 +777,163 @@ def _map_javascript_resource(
...
+        project_resources.files().from_codebase().filter(extension=".ABOUT")
+    )
+    if not from_about_files.exists():
+        return
+
+    indexes = AboutFileIndexes.create_indexes(
+        project=project, from_about_files=from_about_files
+    )
+
+    to_resources = project_resources.to_codebase().no_status()
```
pombredanne, line 927: I am not sure we should exclude things based on  `no status` as this makes this super specific to the step ordering.... at least we should add a comment here.

scanpipe/pipes/d2d.py
```diff
@@ -774,94 +777,163 @@ def _map_javascript_resource(
...
-    for about_file_resource in from_about_files:
-        _map_about_file_resource(project, about_file_resource, to_resources)
-
-        about_file_companions = (
-            about_file_resource.siblings()
-            .filter(name__startswith=about_file_resource.name_without_extension)
-            .filter(extension__in=[".LICENSE", ".NOTICE"])
-        )
-        about_file_companions.update(status=flag.ABOUT_MAPPED)
+    indexes.match_to_resources(to_resources)
```
pombredanne, line 935: Rename this to mapping, not matching and be explicit.... what about this?
```suggestion
    indexes.map_deployed_to_devel_using_about(to_resources)
```
Or something along the lines

pombredanne (changes requested):
See some nits and also https://github.com/nexB/scancode.io/issues/834#issuecomment-1892012130

We may want some more tests too as the path matching with globs is not entirely trivial

scanpipe/pipes/d2d.py
AyanSinhaMahapatra: added this, thanks.

scanpipe/pipes/d2d.py
```diff
@@ -774,94 +777,163 @@ def _map_javascript_resource(
...
+            for mapped_resource in mapped_resources:
+                pipes.make_relation(
+                    from_resource=about_file_resource,
+                    to_resource=mapped_resource,
+                    map_type="about_file",
+                )
+
+            about_file_resource.update(status=flag.ABOUT_MAPPED)
+
+        for about_file_resource in self.about_resources_by_path.values():
```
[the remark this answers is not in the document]
AyanSinhaMahapatra: We have to update aboutcode-toolkit to include this info as extra-data to be able to do this. So keeping the issue https://github.com/nexB/scancode.io/issues/1004 open which tracks this is part 2.

scanpipe/pipes/d2d.py
```diff
@@ -774,94 +777,163 @@ def _map_javascript_resource(
...
+                    ),
+                    model="map_about_files",
+                    details=error_message_details,
+                )
+                continue
 
-    codebase_resources.update(status=flag.ABOUT_MAPPED)
-    about_file_resource.update(status=flag.ABOUT_MAPPED)
+            # Create the Package using .ABOUT data and assign related codebase_resources
+            pipes.update_or_create_package(project, package_data, mapped_resources)
```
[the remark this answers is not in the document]
AyanSinhaMahapatra: added, thanks

scanpipe/pipes/d2d.py
```diff
@@ -774,94 +777,163 @@ def _map_javascript_resource(
...
-            from_resource=about_file_resource,
-            to_resource=to_resource,
-            map_type="about_file",
-        )
+            mapped_resources_about = self.mapped_resources_by_aboutpath.get(about_path)
+            if mapped_resources_about:
+                mapped_resources_about.append(to_resource)
+            else:
+                self.mapped_resources_by_aboutpath[about_path] = [to_resource]
+            to_resource.update(status=flag.ABOUT_MAPPED)
```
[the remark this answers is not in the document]
AyanSinhaMahapatra: This is used to update resources elsewhere in d2d and other pipelines too. See https://github.com/nexB/scancode.io/blob/main/scanpipe/pipes/d2d.py#L284 for example

scanpipe/pipes/d2d.py
```diff
@@ -774,94 +777,163 @@ def _map_javascript_resource(
...
-        # Cannot map anything without the about_resource value.
-        project.add_error(
-            description="ABOUT file does not have about_resource",
-            model="map_about_files",
-            details=error_message_details,
-        )
-        return
+    def match_to_resources(self, to_resources):
+        for to_resource in to_resources:
+            resource_matched = False
```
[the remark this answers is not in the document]
AyanSinhaMahapatra: created a seperate function for this, thanks

scanpipe/pipes/d2d.py
```diff
@@ -774,94 +777,163 @@ def _map_javascript_resource(
...
-    if not filename:
-        # Cannot map anything without the about_resource value.
-        project.add_error(
-            description="ABOUT file does not have about_resource",
-            model="map_about_files",
-            details=error_message_details,
-        )
-        return
+    def match_to_resources(self, to_resources):
+        for to_resource in to_resources:
```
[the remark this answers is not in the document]
AyanSinhaMahapatra: now returning for loggers, thanks

scanpipe/pipes/d2d.py
```diff
@@ -774,94 +777,163 @@ def _map_javascript_resource(
             resource.update(status=flag.MAPPED)
 
 
-def _map_about_file_resource(project, about_file_resource, to_resources):
-    about_file_location = str(about_file_resource.location_path)
-    package_data = resolve.resolve_about_package(about_file_location)
+class AboutFileIndexes(NamedTuple):
+    regex_by_about_path: dict
```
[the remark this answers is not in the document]
AyanSinhaMahapatra: added, thanks!

scanpipe/models.py
```diff
@@ -1909,6 +1909,18 @@ def path_pattern(self, pattern):
         """Resources with a path that match the provided ``pattern``."""
         return self.filter(path__regex=posix_regex_to_django_regex_lookup(pattern))
 
+    def path_patterns(self, patterns, ignore=False):
+        """Resources with a path that match the provided ``pattern``."""
```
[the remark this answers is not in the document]
AyanSinhaMahapatra: this is not used anymore (was being used in another implementation which was improved upon later), so dropped, and modified the other docstring.

scanpipe/pipes/d2d.py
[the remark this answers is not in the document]
AyanSinhaMahapatra: Actually we have the UpdateMixin at https://github.com/nexB/scancode.io/blob/main/scanpipe/models.py#L429 which defines an update() method to use everywhere, so instead of doing a
```
to_resource.status = flag.ABOUT_MAPPED
to_resource.save(update_fields=["status"])
```
we now do
```
to_resource.update(status=flag.ABOUT_MAPPED)
```
which takes less space, is cleaner and better. But both essentially do the same thing as per the update() method created. The CodebaseResource class inherits from the ProjectRelatedModel class, which inherits the Updatemixin class with this method, so this is indeed what is being used here.
So I think we are good here.

scanpipe/models.py
```diff
@@ -1793,12 +1793,12 @@ def profile(self, print_results=False):
                 print(output_str)
 
 
-def posix_regex_to_django_regex_lookup(regex_pattern):
+def convert_glob_to_django_regex(glob_pattern):
     """
     Convert a POSIX-style regex pattern to an equivalent pattern compatible with the
     Django regex lookup.
```
tdruez, line 1799: We may need to update the docstring as well.

scanpipe/pipes/d2d.py
```diff
@@ -774,94 +777,240 @@ def _map_javascript_resource(
             resource.update(status=flag.MAPPED)
 
 
-def _map_about_file_resource(project, about_file_resource, to_resources):
-    about_file_location = str(about_file_resource.location_path)
-    package_data = resolve.resolve_about_package(about_file_location)
+class AboutFileIndexes(NamedTuple):
```
tdruez, line 780: What are the benefits of subclassing NamedTuple in place of a regular class or dataclass?

Also, a docstring for this class would be nice.

tdruez (changes requested):
Some minor suggestions.

scanpipe/pipes/d2d.py
AyanSinhaMahapatra: yeah, dataclass would be better here indeed. Using that. Added docstring too, thanks!

scanpipe/models.py
AyanSinhaMahapatra: Updated, thanks!

AyanSinhaMahapatra:
@pombredanne @tdruez thanks for all the suggestions, I've addressed them, ready for review again.
This now has everything mentioned in https://github.com/nexB/scancode.io/issues/1004

scanpipe/pipes/d2d.py
```diff
@@ -774,94 +777,262 @@ def _map_javascript_resource(
...
 
-def _map_about_file_resource(project, about_file_resource, to_resources):
-    about_file_location = str(about_file_resource.location_path)
-    package_data = resolve.resolve_about_package(about_file_location)
+@dataclass
+class AboutFileIndexes:
+    """
+    Stores the regex path patterns, and ignore patterns,
+    package data and about file/mapped resources to create
+    packages from About files and map to deployed resources.
```
tdruez, line 785: This is a bit confusing, the summary should simply specify what that class is used for.

tdruez (changes requested):
It seems that we are missing unit tests for all the new classes, methods, and functions introduced in those changes.
We are also missing a changelog entry.

scanpipe/pipes/d2d.py
AyanSinhaMahapatra: Updated the docstring, thanks!

AyanSinhaMahapatra:
Thanks @tdruez I've made the requested changes, ready for review again.

Issue #993, opened 2023-11-01
License policy alert sould be shown on empty license


AyanSinhaMahapatra:
Currently we don't show a license policy alert if a resource or a package has empty license, and this should show a warning instead.

tdruez:
@AyanSinhaMahapatra I don't think it's a good idea to break the previous behavior like this.
Instead, the empty license compliance alert should be driven by the license policy conf.

DennisClark:
@AyanSinhaMahapatra   To add to the comments from @tdruez   I think that an empty license condition should normally detect the `no-license` license 

https://scancode-licensedb.aboutcode.org/no-license.html

and ideally that "license" should be configured with the compliance alert in the license policy.

pombredanne:
We should design this differently the lack of license is more than a policy problem... this is rather a problem of todo/review instead.

PR #994, opened 2023-11-01
Return policy warning if no license

AyanSinhaMahapatra:
Reference: https://github.com/nexB/scancode.io/issues/993


tdruez (changes requested):
I don't think this is the proper way to handle this, see my comment at https://github.com/nexB/scancode.io/issues/993#issuecomment-1792627603

pombredanne:
Agreed this should be designed differently.  Closing per https://github.com/nexB/scancode.io/issues/993#issuecomment-1964268900

PR #1014, opened 2023-11-14
Display the entire content of map files

### LINKED ISSUE #1013: SCIO not displaying the entire content of map files
While viewing the content of a `map` file in SCIO, it only displays the content inside the `sourcesContent` key, which is confusing and misleading. Instead, it should display the entire content of the map file.
The problem stems from the fact that we're using Textcode to parse the content of files, and in the case of map files, Textcode only returns the content inside the sourcesContent key.

keshav-space:
- fixes #1013


keshav-space:
> This seems a bit hackich, shouldn't we use a better global way to display source with the best integrity possible?
> 
@tdruez  We are using textcode as a global way for displaying source, and it's working well, except for the map files.
I don't believe handling this use case in Textcode is what we want, as the current behavior of textcode is desirable in the SCTK use case. Additionally, the idea of returning `numbered_text_lines` won't make much sense for the entire map file, given that there are nested files that will have their own numbered lines.

scanpipe/models.py
```diff
@@ -2292,6 +2292,15 @@ def file_content(self):
...
         """
         from textcode.analysis import numbered_text_lines
+        from typecode import get_type
+
+        # When reading a map file, Textcode only provides the content inside
+        # `sourcesContent`, which can be misleading during any kind of review.
+        # This workaround ensures that the entire content of map files is displayed.
+        file_type = get_type(self.location)
+        if file_type.is_js_map:
+            f = open(self.location, "r")
```
tdruez, line 2302: Use a context manager when opening files to ensure those are properly closed.

scanpipe/tests/test_models.py
```diff
@@ -1142,6 +1143,17 @@ def test_scanpipe_codebase_resource_model_file_content(self):
         line_count = len(resource.file_content.split("\n"))
         self.assertEqual(101, line_count)
 
+    def test_scanpipe_codebase_resource_model_file_content_for_map(self):
+        map_file_path = self.data_location / "d2d-javascript/to/main.js.map"
+        copy_input(map_file_path, self.project1.codebase_path)
+        resource = self.project1.codebaseresources.create(path="main.js.map")
+
+        f = open(map_file_path, "r")
```
tdruez, line 1151: Use a context manager when opening files to ensure those are properly closed.

tdruez (changes requested):

scanpipe/models.py
keshav-space: Done :)

Issue #991, opened 2023-11-01
Inputs added with file picker are lost when creating a new project


pombredanne:
I create a project and I add a few inputs with a click to open a file picker.
Then again I click to open a file picker again and select more inputs.
The first inputs are lost and replaced by the second inputs.
NB: this does not happen with drag and drop.

PR #1036, opened 2023-12-20
ui/project_form: Improve file handling in UI

jayanth-kumar-morem:
- Enhance file input handling in the UI to avoid losing selected files when adding new inputs.
- Introduce a `selectedFiles` array to store and manage selected files.
- Display selected files in the UI with additional features, including a delete button.
- Update the `removeFile` function to correctly remove files from both the UI and the `selectedFiles` array.

Fixes: #991


tdruez:
@jayanth-kumar-morem I've noticed an issue while reviewing this PR, to reproduce:
- Drag and drop a file into the form, then drag and drop it again. The file is duplicated in the list. Note that this does not happen with the "Click" to select file.

Issue #1047, opened 2024-01-11
URL schema when written in wrong case fails (not gracefully)


silverhook:
**Describe the bug**
When adding a Docker source, the URL schema seems to be case sensitive, and if it is not exactly as `docker://`, it fails, but does not explain why.

**System configuration**
- Which version of ScanCode.io are you running? v32.7.0 
- Are you running the app using Docker? yes
- On which OS? Linux
- What inputs are you using? docker
- Which pipeline are you running? docker

**To Reproduce**
1. start a new scan
2. add a Docker resource and do a case typo by writing e.g. `Docker://postgres:12.17-alpine3.19`
3. fill in everything else correctly and set the pipeline to `docker`
4. start the scan
5. see error

**Expected behavior**
Either:
a) not have the schema part of the URL case sensitive, and just finish the scan; or
b) return an error, where it explains the issue was the case

jayanthmahadesh:
when URL is Docker:// instead of docker it raises value error

`    whitelist = r"^docker://[a-zA-Z0-9_.:/@-]+$" `
`    if not re.match(whitelist, docker_reference): `
`    raise ValueError("Invalid Docker reference.")`

should i try
whenever there is case mistake in docker:// - like Docker:// 
change it to docker:// automatically

jayanthmahadesh:
@silverhook

silverhook:
> whenever there is case mistake in docker:// - like Docker://
> change it to docker:// automatically

That sounds like a great solution, yes. :+1:

jayanth-kumar-morem:
Docker image names, including the part specifying the repository, image name, and tag, are case-insensitive. Therefore, whether you use uppercase or lowercase letters in the image name, repository, or tag, Docker treats them as the same.

For example, both `docker://postgres:12.17-alpine3.19` and `docker://Postgres:12.17-alpine3.19` refer to the same Docker image.

tdruez:
Schema validation implemented in https://github.com/nexB/scancode.io/pull/1085

An explicit form error is now displayed for in cases of unsupported scheme:

![Screenshot 2024-02-22 at 13 54 51](https://github.com/nexB/scancode.io/assets/489057/e3d5e525-a148-47f4-8e66-ab8a250a6ad0)

PR #1059, opened 2024-01-26
forms, pipes: Make docker urls case-insensitive

jayanth-kumar-morem:
Fixes: #1047

### DIFF
--- a/scanpipe/forms.py
+++ b/scanpipe/forms.py
@@ -80,7 +80,7 @@ class InputsBaseForm(forms.Form):
         Fetch the `input_urls` and sets the `downloads` objects in the cleaned_data.
         A validation error is raised if at least one URL can't be fetched.
         """
-        input_urls_str = self.cleaned_data.get("input_urls", "")
+        input_urls_str = self.cleaned_data.get("input_urls", "").lower()
         input_urls = input_urls_str.split()
 
         errors = check_urls_availability(input_urls)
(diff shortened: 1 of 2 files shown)


scanpipe/forms.py
```diff
@@ -80,7 +80,7 @@ def clean_input_urls(self):
         Fetch the `input_urls` and sets the `downloads` objects in the cleaned_data.
         A validation error is raised if at least one URL can't be fetched.
         """
-        input_urls_str = self.cleaned_data.get("input_urls", "")
+        input_urls_str = self.cleaned_data.get("input_urls", "").lower()
```
tdruez, line 83: We should not tamper with the user input.
While trying to fix the user errors, you may be breaking valid inputs.
Here, the path component of a URL may be case-sensitive, depending on the web server and file system used.

I would suggest that we never modify the user-provided content but instead raise the proper error including a suggestion when possible.

See the behavior of Python exceptions for example:
```
>>> var = 1
>>> print(vr)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
NameError: name 'vr' is not defined. Did you mean: 'var'?
```

I think we should simply validate the schema of the input_urls and raise an error if not in the supported schema list. 
As a second improvement, we could check if the `lower()` value of the schema is in the list and improve the error with a suggestion.
Something along: "Docker:// schema is not supported. Did you mean: 'docker://'?"

tdruez (changes requested):

tdruez:
Implemented in https://github.com/nexB/scancode.io/pull/1085

PR #1060, opened 2024-01-26
Process deleting projects as background task

### LINKED ISSUE #1002: Failure to delete large project
I have a project with about 1M files and 2GB compressed inputs (~10G uncompressed)
Deleting failed now and then with a "Gateway timeout" error

jayanth-kumar-morem:
1. Fix Default Filtering in ProjectFilterSet:
   - Modified the `ProjectFilterSet` to include `is_marked_for_deletion` in the default filtering when `is_archived` is not provided.
   - This ensures that projects marked for deletion are excluded from default views.

2. Migration for `is_marked_for_deletion`:
   - Added a migration (`0052_project_is_marked_for_deletion.py`) to introduce the `is_marked_for_deletion` field in the `Project` model.

3. Updated `Project` Model:
   - Added `is_marked_for_deletion` field to the `Project` model with a default value of `False`.

4. Modified `ProjectListView` to Exclude Marked Projects:
   - Modified the `get_queryset` method in `ProjectListView` to only fetch projects with `is_marked_for_deletion=False`.

5. Introduced `mark_for_deletion` Method:
   - Added a `mark_for_deletion` method in the `Project` model to set the `is_marked_for_deletion` flag and save the model.

6. Changed `delete` Method in `Project` Model:
   - Renamed the `delete` method to `delete_action` in the `Project` model to avoid conflicts.
   - Introduced a new `delete` method that marks the project for deletion and enqueues a background deletion task.

7. Background Deletion Task:
   - Created a background deletion task using `django_rq` to handle project deletion asynchronously.
   - Checks if the project is still marked for deletion before proceeding.
   - If an error occurs during deletion, updates project status and logs an error.

8. Updated `ProjectActionView` Success Message:
   - Modified the success message in `ProjectActionView` for the "delete" action to provide a more informative message based on the count.

Fixes: #1002


scanpipe/filters.py
```diff
@@ -349,9 +349,13 @@ def __init__(self, data=None, *args, **kwargs):
 
         # Default filtering by "Active" projects.
         if not data or data.get("is_archived", "") == "":
-            self.queryset = self.queryset.filter(is_archived=False)
+            self.queryset = self.queryset.filter(
+                is_archived=False, is_marked_for_deletion=False
+            )
```
tdruez, line 354: It would make sense to add a new `active()` QuerySet method to avoid code duplication.

scanpipe/models.py
```diff
@@ -534,6 +535,7 @@ class Project(UUIDPKModel, ExtraDataFieldMixin, UpdateMixin, models.Model):
     labels = TaggableManager(through=UUIDTaggedItem)
 
     objects = ProjectQuerySet.as_manager()
+    is_marked_for_deletion = models.BooleanField(default=False)
```
tdruez, line 538: Keep the fields grouped, objects always go last.

scanpipe/models.py
```diff
@@ -633,6 +635,16 @@ def delete(self, *args, **kwargs):
 
         return super().delete(*args, **kwargs)
 
+    def mark_for_deletion(self):
+        self.is_marked_for_deletion = True
+        self.save()
```
tdruez, line 640: Using save() as-is may have unwanted consequences.
You should at least provide the update_fields value, or better, use the `self.update(is_marked_for_deletion=True)` method.

scanpipe/models.py
```diff
@@ -633,6 +635,16 @@ def delete(self, *args, **kwargs):
...
+
+    def delete_in_background(self):
+        # Mark the project for deletion and enqueue background deletion task
+        self.mark_for_deletion()
+        q = Queue("default", connection=redis.Redis())
+        job = q.enqueue(tasks.background_delete_task, self)
```
tdruez, line 646: Why not using the `django_rq.enqueue` approach, see `execute_task_async`

scanpipe/tasks.py
```diff
@@ -76,3 +78,19 @@ def execute_pipeline_task(run_pk):
...
+    if not project.is_marked_for_deletion:
+        return
+
+    # Perform the deletion process
+    try:
+        project.delete()
+    except Exception as e:
+        # Handle errors and update project errors or display a banner
+        project.is_marked_for_deletion = False
+        project.save()
```
tdruez, line 95: Use `update` here too

scanpipe/tasks.py
```diff
@@ -76,3 +78,19 @@ def execute_pipeline_task(run_pk):
...
             next_run.start()
+
+
+@job
+def background_delete_task(project):
+    # Check if the project is still marked for deletion
+    if not project.is_marked_for_deletion:
+        return
+
+    # Perform the deletion process
```
tdruez, line 89: Comment not useful.

scanpipe/tasks.py
```diff
@@ -76,3 +78,19 @@ def execute_pipeline_task(run_pk):
...
+    try:
+        project.delete()
+    except Exception as e:
+        # Handle errors and update project errors or display a banner
```
tdruez, line 93: Probably better to convert this into a logging entry.

scanpipe/views.py
```diff
@@ -1086,6 +1089,8 @@ def perform_action(self, action, project_uuid, action_kwargs=None):
             raise Http404
 
     def get_success_message(self, action, count):
+        if action == "delete":
+            return f"{count} project{'s' if count != 1 else ''} {'is' if count == 1 else 'are'} being deleted in the background."
```
tdruez, line 1093: I don't think the user needs to know about the internals of the deletion, let's keep the initial message.

tdruez (changes requested):
Pretty good start. See my comments for improvements.

pombredanne:
This is a good approach and well explained... it is likely missing not much to get this merged... @jayanth-kumar-morem do you mind looking into some of the comments?

scanpipe/filters.py
```diff
@@ -349,9 +349,11 @@ def __init__(self, data=None, *args, **kwargs):
 
         # Default filtering by "Active" projects.
         if not data or data.get("is_archived", "") == "":
-            self.queryset = self.queryset.filter(is_archived=False)
+            self.queryset = self.queryset.active()
 
-        active_count = Project.objects.filter(is_archived=False).count()
+        active_count = Project.objects.filter(
+            is_archived=False, is_marked_for_deletion=False
```
tdruez, line 355: We should use `active()` here as well, no?

scanpipe/models.py
```diff
@@ -633,6 +638,14 @@ def delete(self, *args, **kwargs):
 
         return super().delete(*args, **kwargs)
 
+    def mark_for_deletion(self):
+        self.update(is_marked_for_deletion=True)
+
+    def delete_in_background(self):
+        # Mark the project for deletion and enqueue background deletion task
+        self.mark_for_deletion()
```
tdruez, line 646: I think we can put `self.update(is_marked_for_deletion=True)` directly here, no need for the `mark_for_deletion` method yet since it's a one-liner that is only used in one place.

scanpipe/models.py
```diff
@@ -633,6 +638,14 @@ def delete(self, *args, **kwargs):
 
         return super().delete(*args, **kwargs)
 
+    def mark_for_deletion(self):
+        self.update(is_marked_for_deletion=True)
+
+    def delete_in_background(self):
+        # Mark the project for deletion and enqueue background deletion task
```
tdruez, line 645: A docstring would be better than a comment.

scanpipe/tasks.py
```diff
@@ -76,3 +78,17 @@ def execute_pipeline_task(run_pk):
...
+def background_delete_task(project):
+    # Check if the project is still marked for deletion
+    if not project.is_marked_for_deletion:
+        return
+
+    try:
+        project.delete()
+    except Exception as e:
+        info(f"Deletion failed: {str(e)}", project.pk)
+        project.update(is_marked_for_deletion=True)
```
tdruez, line 93: I'm wondering is this line is of any use, the `is_marked_for_deletion` must be already True to reach this line, no?

tdruez (changes requested):

PR #1096, opened 2024-03-01
Update debian support

AyanSinhaMahapatra:
Also updates SCTK to latest develop and https://github.com/nexB/scancode-toolkit/pull/3682
So it has changes from https://github.com/nexB/scancode-toolkit/pull/3620 and regenerated test files.


tdruez:
@AyanSinhaMahapatra What is the status of a proper release for SCTK?

AyanSinhaMahapatra:
@tdruez we're hoping for a release this week :crossed_fingers: 
Opened this PR as a part of https://github.com/nexB/purldb/issues/245 as I'm currently testing all the updates in SCTK, SCIO and purldb together, but I will update this PR to a proper released version of SCTK before we merge this.

scanpipe/pipes/rootfs.py
```diff
@@ -265,6 +265,8 @@ def scan_rootfs_for_system_packages(project, rootfs):
     installed_packages = rootfs.get_installed_packages(package_getter)
     for index, (purl, package) in enumerate(installed_packages):
         logger.info(f"Creating package #{index}: {purl}")
+        if package.namespace != distro_id:
```
AyanSinhaMahapatra, line 268: This is related to: 

In SCTK we are now getting the namespace populated from clues present in package data with this [commit](https://github.com/nexB/scancode-toolkit/pull/3647/commits/1cd5d02b3ecf30c3dcfaedc41e1d9d1a9802240c) and if we don't have any clues we are assigning a defualt value to the namespace which is `debian` for debian packages.

In SCIO we get distro data from os-release which overrides these values, as they are always correct and takes care of occasional false-positives and default debian cases.

scanpipe/tests/data/basic-rootfs_root_filesystems.json
```diff
@@ -31,24 +31,26 @@
...
               "os": "linux",
               "architecture": null,
-              "name": "Debian GNU/Linux",
-              "version": "9 (stretch)",
-              "identifier": "debian",
-              "id_like": null,
-              "version_codename": null,
-              "version_id": "9",
-              "pretty_name": "Debian GNU/Linux 9 (stretch)",
+              "name": "Ubuntu",
```
AyanSinhaMahapatra, line 34: Here I've updated the `os-release` file in the basic-rootfs test case to the ubuntu distro file, as the packages detected below seem to be ubuntu packages, and the old example seemed to be wrong. Otherwise the namespaces being overridden from this `os-release` file seemed to be wrong.

AyanSinhaMahapatra (commented):
Ready for review now. I've rebased this on top of the merged SCTK v32.1.0 now in main.

scanpipe/pipes/rootfs.py
pombredanne: Per https://github.com/GoogleContainerTools/distroless 

see https://github.com/GoogleContainerTools/distroless/blob/f55b2b343481f52ef3bde34c8a2f3631a40a36c1/debian_packages.yaml#L75 ... the namespace is debian in all cases.... some deb11 some deb12. Eventually the way to go will be to match these.

pombredanne (approved):
LGTM!

scanpipe/pipes/rootfs.py
AyanSinhaMahapatra: In that case we are good, thanks!

Issue #1099, opened 2024-03-04
DiscoveredPackage entries missing package detection info


AyanSinhaMahapatra:
In SCTK Package model (Package Instance, detected in a codebase, created out of Package metadata and it's files) has the following fields:
- datafile_paths  (List of paths for the package metadata for this package instance)
- datasource_ids  (List of Datasource Ids for the respective package metadata)

But in SCIO DiscoveredPackage model, we only have:
- datasource_id 

As a result of the mismatch, during DiscoveredPackage creation/update from PackageData fields, we were losing these data fields and were displaying a blank Datasource IDs field  in the Package View -> Others Tab.

For example: scanning [celery-5.2.7](https://files.pythonhosted.org/packages/ce/21/41a0028f6d610987c0839250357c1a00f351790b8a448c2eb323caa719ac/celery-5.2.7.tar.gz)  with a `scan_single_package` pipeline, we would get the following in the package view:

![celery-no-datasource-ids](https://github.com/nexB/scancode.io/assets/25746274/bd79b45c-ad84-4e6e-8864-e9bfd1fe4771)

Issue #1081, opened 2024-02-19
Debian package is not extracted recursively


pombredanne:
I create a "scan_single_package" project using http://ftp.us.debian.org/debian/pool/main/e/elfutils/elfutils_0.188-2.1_amd64.deb as an input.
The package is not fully extracted recursively, even though the settings are to extract recursively

tdruez:
@pombredanne Could you add the `recurse` option to the `extractcode.api.extract_archive` function?

pombredanne:
`extractcode.api.extract_archive` cannot recurse
For this you need to use: `extractcode.api.extract_archives` instead

tdruez:
@pombredanne The current implementation of extraction for the single package scan is to extract directly the single archive to the codebase work directory.
The `extractcode.api.extract_archives` extracts in place, so we need to copy the input archive first. This would change the scan root directory of 1 level.
Are you sure you want to apply this change?

Current:
![Screenshot 2024-02-19 at 22 05 38](https://github.com/nexB/scancode.io/assets/489057/671aed8a-e6a2-43ff-9a5e-0c4af202ebef)


Using extracts:
![Screenshot 2024-02-19 at 22 05 42](https://github.com/nexB/scancode.io/assets/489057/28b2f013-5ad2-4226-86b4-87b5f1b2ac7a)

tdruez:
We could add a new step to run the `extract_archives` after extracting the single archive to the codebase dir.

pombredanne:
Ok, then may be you could keep the extraction the way it is today and add a call to `extractcode.api.extract_archives()` afterwards to force recursion?

tdruez:
Fixed in https://github.com/nexB/scancode.io/pull/1083

PR #1100, opened 2024-03-04
Fix package detection info in DiscoveredPackages #1099

AyanSinhaMahapatra:
Reference: https://github.com/nexB/scancode.io/issues/1099


tdruez:
@AyanSinhaMahapatra 

> Should we show this in the DiscoveredPackageListView (similar to how we show related fields in DiscoveredDependencyListView)?

This does not seems necessary. The space is limited in list views and the detection info are not essentials.

> In DiscoveredPackageListView and DiscoveredPackageDetailsView should we link to the CodebaseResourceDetailsView for each datafile path (similar to what we do in dependencies)?

Sure, why not? Be conscious about the db queries though, but I guess we can generate the links just using the path without any extra queries.

> Here we have an extra codebase/ appearing in packages created by scan_codebase_packages pipeline due to Debian package is not extracted recursively #1081 (comment), is it okay to remove this specifically if it appears in front of the path?

This sounds like a hack, it would be better to store the proper path in the db than presenting altered data in the UI.

> How do we show these fields in DiscoveredPackageDetailsView, since these are lists now, and with links, in case of datafile_paths?

What about a new "Detection" tab, similar to the one on the CodebaseResourceDetailsView ?

----

Also, the migration includes a file type change, from CharField to JSONField for `datasource_ids`, yet it does not include any data migration.
Are you expecting any issues from this? Have you tried to run this on a real dataset?

AyanSinhaMahapatra:
> Sure, why not? Be conscious about the db queries though, but I guess we can generate the links just using the path without any extra queries.

Yes, we're not doing any queries here, just creating the links in the templates.

> This sounds like a hack, it would be better to store the proper path in the db than presenting altered data in the UI.

Doing this on the package creation side, and storing the right path now.

> What about a new "Detection" tab, similar to the one on the CodebaseResourceDetailsView ?

This sounds good, adding this.

> Also, the migration includes a file type change, from CharField to JSONField for datasource_ids, yet it does not include any data migration. Are you expecting any issues from this? Have you tried to run this on a real dataset?

So we were not storing any datasource IDs previously, due to the bug linked, and so my assumption was we don't have any datasource IDs for any packages stored from previous scans. This was because there was a attribute mismatch and we were ignoring this data while importing from PackageData/creating a DiscoveredPackage. But when I'm checking the DiscoveredPackage objects for a SCIO instance, there were some Packages with these datasource_id fields populated for the new inspect_packages pipeline, or anywhere we had Packages created out of PackageData objects, so I added a migration for this now.

scanpipe/pipes/__init__.py
```diff
@@ -155,6 +155,16 @@ def _clean_package_data(package_data):
...
             package_data["release_date"] = datetime.fromisoformat(release_date).date()
+
+    # Strip leading "codebase/" to make path compatible with
+    # paths stored in resource database
+    cleaned_datafile_paths = []
+    for path in package_data.get("datafile_paths", []):
+        cleaned_path = path
+        if path.startswith("codebase/"):
+            cleaned_path = path[len("codebase/") :]
+        cleaned_datafile_paths.append(cleaned_path)
```
tdruez, line 166: What about: 
`cleaned_datafile_paths = [path.removeprefix("codebase/") for path in package_data.get("datafile_paths", [])]`

tdruez (changes requested):
@AyanSinhaMahapatra Could you add a changelog entry to explain those changes?
Also, have you tried the data migrations on a real dataset?

scanpipe/pipes/__init__.py
AyanSinhaMahapatra: Using `path.removeprefix` is much cleaner, thanks! Added this.

AyanSinhaMahapatra:
@tdruez I've added your feedback and this is ready to review again.

> Could you add a changelog entry to explain those changes?

Added this now.

> Also, have you tried the data migrations on a real dataset?

Yes, I had created a test dataset with pipeline runs of all pipeline types and different kinds of discovered packages created from those pipelines (including some DiscoveredPackages with `datasource_id` values from non-empty inspect_packages pipelines) and the data migration worked nicely on this without errors. See output below:
```
-> Apply database migrations
bin/python manage.py migrate
Operations to perform:
  Apply all migrations: admin, auth, authtoken, contenttypes, django_rq, scanpipe, sessions, taggit
Running migrations:
  Applying scanpipe.0054_discoveredpackage_datafile_paths...
Compute datasource_ids for 22 packages.
Updating DB objects...
 OK
-> Run the test suite
bin/python manage.py test --noinput
Found 526 test(s).
Creating test database for alias 'default'...

Compute datasource_ids for 0 packages.
Updating DB objects...
System check identified no issues (0 silenced).
.......................................................................................................................................................................................x......................................................................................................................................................................................................................................................................................................................................................
----------------------------------------------------------------------
Ran 526 tests in 100.089s

OK (expected failures=1)
Destroying test database for alias 'default'...
```

I've also merged latest main to resolve the conflicts and renamed the migration accordingly.

PR #1074, opened 2024-02-09
Restructure pipelines for verbosity

AyanSinhaMahapatra:
Remove scan_codebase_packages pipeline, and restructure inspect_packages pipeline into load_sbom and resolve_packages pipelines.

Reference: https://github.com/nexB/scancode.io/issues/1040
Reference: https://github.com/nexB/scancode.io/discussions/1035
Reference: https://github.com/nexB/scancode.io/issues/1034
Reference: #1044

PR #1118, opened 2024-03-13
Use SCTK `package_only` in inspect_packages pipeline

AyanSinhaMahapatra:
This PR adds the following changes in `inspect_packages` pipeline:

* The key update is using the newly added `package_only` attribute in the [`get_package_data` API](https://github.com/nexB/scancode-toolkit/blob/develop/src/scancode/api.py#L295) which is equivalent to the new --package-only CLI option in SCTK. This is an optional attribute, which is set True, only scans package manifests for package data, and skips license and copyright detected on extracted license statements and copyrights, hence making this step faster.
* Also adds a couple updates related to the restructured `inspect_packages` pipeline introduced in https://github.com/nexB/scancode.io/pull/1074:
  - Creation of packages/dependencies are moved to a separate pipe in the pipeline so we have info on time taken to scan for package data, and to create them seperately, to enable comparision with the previous pipeline.
  - Only creates packages if purl is present, else skip package creation. (since we are not doing package assembly, we cannot determine whether failure to create a package is an issue, so the messages were noise)
  - Skips creating packages from package handlers of type: `models.NonAssemblableDatafileHandler` (like autotools configure scripts for example: https://github.com/nexB/scancode-toolkit/blob/develop/src/packagedcode/build.py#L50C33-L50C68)


tdruez (changes requested):
@AyanSinhaMahapatra What's the latest status on a SCTK release? Also, could you add a changelog entry for those changes?

tdruez:
@AyanSinhaMahapatra please provide context and changelog for those changes.
"Update inspect_packages pipeline" is not explicit.

AyanSinhaMahapatra:
@tdruez I've rebased the changes on the latest main, where we have the new SCTK release v32.1.0, and I've also added more context for the changes in PR description above and added a CHANGELOG entry, this is ready to review now. Thanks!

PR #1133, opened 2024-03-27
Update docker-compose.yml

JonoYang:
This PR add a new service in the `docker-compose.yml` files to run the `purldb-scan-queue-worker` command.


tdruez (changes requested):
I don't think we want to add a service, in the main compose config file, that is only required for a very specific use-case.
I would suggest to have a new compose file dedicated to this.
https://docs.docker.com/compose/multiple-compose-files/

JonoYang:
@tdruez I've reverted the docker-compose files I modified and created a new docker-compose.purldb-scan-queue-worker.yml for this.

tdruez:
@JonoYang There must be a better way than duplicating the whole compose file for a few line additions.
Also, this is likely to be out of sync each time we make changes to the original compose file.

Can't we use the "merge", "extend", "include", features of compose? https://docs.docker.com/compose/multiple-compose-files/

What about a simpler filename `docker-compose.purldb-queue.yml` file including the docker-compose.yml file, such as:

```
version: "3"

include:
  - docker-compose.yml

services:
  purldb_scan_queue_worker:
    build: .
    command: wait-for-it --strict --timeout=120 web:8000 -- sh -c "
        ./manage.py purldb-scan-queue-worker --async --sleep 3"
    env_file:
      - docker.env
    volumes:
      - .env:/opt/scancodeio/.env
      - /etc/scancodeio/:/etc/scancodeio/
      - workspace:/var/scancodeio/workspace/
    depends_on:
      - db
      - web

```

JonoYang:
@tdruez I've renamed the `purldb-scan-queue-worker` management command to `purldb-scan-worker` and updated `docker-compose.purldb-scan-worker.yml` with the changes you suggested.

Issue #1120, opened 2024-03-19
Cannot copy path to input URL in project


pombredanne:
I create a project from a URL https://repo1.maven.org/maven2/com/sun/jna/jna/3.0.9/jna-3.0.9.jar 
And I cannot afterwards easily copy this URL back. It only shows in an fleeing popup that does not keep the focus.

I would like to the URLs to be fully displayed without a hover and easy to copy, and display possibly as a link.

lata-11:
Would like to work on this issue @pombredanne!

tdruez:
A "copy URL" button has been added since.

PR #1135, opened 2024-03-27
Enable copying path to input URL in project #1120

lata-11:
This PR fixes https://github.com/nexB/scancode.io/issues/1120

Intially the input URL was not clickable and easy to copy.
image

Now, I have changed the input URL to a hyperlink so when clicked it will open the URL and can be copied easily.
image

please let me know if this solution works fine or not. Thanks.

### DIFF
--- a/scanpipe/templates/scanpipe/panels/project_inputs.html
+++ b/scanpipe/templates/scanpipe/panels/project_inputs.html
@@ -29,7 +29,7 @@
         {% if input_source.filename %}
           {{ input_source.filename }}
         {% else %}
-          {{ input_source.download_url }}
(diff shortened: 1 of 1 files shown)


pombredanne:
Thanks! this looks good to me. Merging.

pombredanne:
Actually, why do we have tow tooltips showing up now?

lata-11:
> Actually, why do we have tow tooltips showing up now?

That's because I have used `title` attribute within anchor tag. Should I remove that, if that is a problem?

tdruez:
A copy button was added in https://github.com/aboutcode-org/scancode.io/pull/1627

Issue #1140, opened 2024-03-29
Resolved packages are not associated with their source codebase resource


tdruez:
Using `load_sboms`, created packages are not assigned to their SBOM codebase resource from which they were resolved.
This is problematic to track the source of a package when using multiple input files.

PR #1141, opened 2024-03-29
Associate resolved packages with their source codebase resource #1140

tdruez:


AyanSinhaMahapatra (commented):
@tdruez looks good to me, this is assigning resources to packages correctly now.

Issue #1156, opened 2024-03-29
Collect Mariner/Azurelinux packages in container scan


pombredanne:
See https://packages.microsoft.com/cbl-mariner/2.0/prod/base/x86_64/Packages/
(And https://packages.microsoft.com/cbl-mariner/)

This is an RPM distro mostly used in containers @ Azure. We are already finding installed packages in SCIO container scans.

See also these related issues:
- https://github.com/nexB/vulnerablecode/issues/751 
- https://github.com/nexB/scancode-toolkit/issues/3060
- [ ] nexB/purldb#383

pombredanne:
`/usr/share/licenses/<package>/....` has license files, as in `/usr/share/licenses/openssl/LICENSE`

pombredanne:
See also https://github.com/microsoft/marinara to find how it is built

pombredanne:
See https://mcr.microsoft.com/en-us/product/cbl-mariner/base/core/about for test images

AyanSinhaMahapatra:
There are two kinds of mariner:

* base: https://mcr.microsoft.com/en-us/product/cbl-mariner/base/core/about
* distroless: https://mcr.microsoft.com/en-us/product/cbl-mariner/distroless/base/

It seems only the distroless one here has the properties mentioned in https://github.com/nexB/scancode.io/issues/1156#issuecomment-2042738884, otherwise the base images have a `/var/lib/rpm/rpmdb.sqlite` and we can create packages out of it once `mariner` is included in the list of accepted distros. But we do not assign resources to packages correctly here btw.

AyanSinhaMahapatra:
This was completed by https://github.com/aboutcode-org/scancode.io/pull/1161, closing!

PR #1161, opened 2024-04-10
Add mariner to supported distros

AyanSinhaMahapatra:
Reference: https://github.com/nexB/scancode-toolkit/pull/3734
Reference: https://github.com/nexB/scancode.io/issues/1156

### DIFF
--- a/scanpipe/pipes/rootfs.py
+++ b/scanpipe/pipes/rootfs.py
@@ -47,6 +47,7 @@ SUPPORTED_DISTROS = [
     "fedora",
     "sles",
     "opensuse",
+    "mariner",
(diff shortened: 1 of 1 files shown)


AyanSinhaMahapatra:
Mariner support was added and merged in https://github.com/nexB/scancode-toolkit/pull/3734, so just adding this to supported distros should detect mostly correctly.
We also have some generic improvements pending at https://github.com/nexB/scancode-toolkit/pull/3770 and https://github.com/nexB/scancode.io/tree/improve-namespace-support-distros I will follow up at, to get the correct purls by detecting namespace from distro correctly.

PR #1180, opened 2024-04-25
Populate package and dependency attributes in inspect_packages

AyanSinhaMahapatra:
Properly populate dependency attributes like for_package, datasource_id and datafile_resource for DiscoveredDependencies found in inspect_packages pipeline.
Also populate datasource_ids and codebase_resources correctly for packages.
Here we only assign to the package the codebase resource of the package manifest where it was found, as assigning all other package resources would require a full package assembly, and thus out of scope for this pipeline.

### DIFF
--- a/scanpipe/pipes/__init__.py
+++ b/scanpipe/pipes/__init__.py
@@ -180,7 +180,7 @@ def update_or_create_package(project, package_data, codebase_resources=None):
     package_data = _clean_package_data(package_data)
     # No values for package_uid requires to be empty string for proper queryset lookup
     package_uid = package_data.get("package_uid") or ""
-    datasource_id = package_data.get("datasource_id")
+    datasource_id = package_data.get("datasource_id") or ""
 
     package = DiscoveredPackage.objects.get_or_none(
         project=project,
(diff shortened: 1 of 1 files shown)


scanpipe/pipes/__init__.py
```diff
@@ -180,6 +180,7 @@ def update_or_create_package(project, package_data, codebase_resources=None):
     package_data = _clean_package_data(package_data)
     # No values for package_uid requires to be empty string for proper queryset lookup
     package_uid = package_data.get("package_uid") or ""
+    datasource_id = package_data.get("datasource_id")
```
tdruez, line 183: Looking at the comment above:
"# No values for package_uid requires to be empty string for proper queryset lookup"

Could you make sure this is not also required for the `datasource_id` field?

scanpipe/pipes/__init__.py
```diff
@@ -192,8 +193,12 @@ def update_or_create_package(project, package_data, codebase_resources=None):
     else:
         package = DiscoveredPackage.create_from_data(project, package_data)
 
-    if package and codebase_resources:
-        package.add_resources(codebase_resources)
+    if package:
+        if datasource_id and datasource_id not in package.datasource_ids:
+            package.update(datasource_ids=[datasource_id])
```
tdruez, line 198: Is there a risk of losing data here? Should the datasource_id be added to the list instead?

tdruez (changes requested):

scanpipe/pipes/__init__.py
AyanSinhaMahapatra: Adding to list instead, thanks!

PR #1181, opened 2024-04-25
Addon pipeline to collect tree-sitter symbols

keshav-space:
Refrence https://github.com/nexB/purldb/issues/400


tdruez (changes requested):
Looks good but missing a changelog entry and the conflicts must be resolved.

Issue #1194, opened 2024-04-30
ScanCode.io: need to scan for license inside a detected container package


pombredanne:
It is important to have an option to scans the files of a detected package.
We should run this scans for each set of package files separately and eventually compute summaries there. 
Ideally this would be a "group" tag on extra pipeline steps

pombredanne:
@AyanSinhaMahapatra @tdruez can you guys look into this?

AyanSinhaMahapatra:
> We should run this scans for each set of package files separately

Should this part be added first and then the following

> and eventually compute summaries there.

Added later, or should this be done together? Note that the first part is a much smaller change.

Issue #762, opened 2023-06-03
scan_codebase should also scanned all the file that belong to an identifier package


pombredanne:
scan_codebase should also scanned all the file that belong to an identifier package otherwise we are not getting all details.

pombredanne:
Note that we should eventually treat each set of package files as some sort of a subscan unit where we would also compute and store summaries and clarity scores

Issue #83, opened 2021-02-04
Optionally scan package files too


pombredanne:
Files that are detected as part of a package are not further scanned in the default pipelines. We should have a few pipeline extras to allow optionally to:
- lookup extra package data and file data externally (direct API calls to a registry, upcoming PackageDB, ClearlyDefined, etc)
- scan files with scancode TK
- fetch and scan corresponding source code with scancode TK

PR #1207, opened 2024-05-06
Scan package files and extract for packages

AyanSinhaMahapatra:
In all the following pipelines:
* rootfs
* docker
* docker-windows
* scan_codebase
when we scan files for license, copyright and others, we are [skipping the scan](https://github.com/nexB/scancode.io/blob/main/scanpipe/pipes/scancode.py#L360) for codebase resources which have a status already before this step, and so anything tagged as `application-package` or `system-package` will not be scanned.

In the `match_not_analyzed_to_system_packages` pipe of the rootfs pipeline, we are matching all codebase resources which are a part of that package to the discovered package object and also [updating it's status](https://github.com/nexB/scancode.io/blob/main/scanpipe/pipes/rootfs.py#L315) to `system-package`. (It seems like earlier we were also doing this for application packages with the [`match_not_analyzed_to_application_packages`](https://github.com/nexB/scancode.io/blob/main/scanpipe/pipelines/root_filesystem.py#L110) function, but this is not used anywhere after [this](https://github.com/nexB/scancode.io/commit/b270becc48c5d6d314d34e7a800eb0c710efaad5))

Similary in the docker pipelines, in the [`create_system_package` function](https://github.com/nexB/scancode.io/blob/main/scanpipe/pipes/docker.py#L224) of the `collect_and_create_system_packages` step we are updating the status of package files to `system-package`.

We can either:
1. stop tagging the status of files which are part of a system-package 
2. or re-scan all package files tagged as system/application package

In this PR I've tried out the 2. approach, as this is what we do in SCTK also, but here we have to create a new argument `update_status` and pass it on to the function which saves data to resources after the scan to not overwrite the `system-package` or `application-package` status for codebase-resources to `scanned`, which was a side-effect of the file scans.

Since all these pipelines already did scan application package files (which were not metadata files/lockfiles) I'm assuming we also want to scan the metadata files which were not being scanned? Otherwise #762 does not make any sense. Note here that license scans which are part of a package scan (parsing the manifest and then only running license detection on the extracted part) can be different in some complex files than a simple license scan of the file, and we might need to improve how we handle this in SCTK to avoid confusion. See https://github.com/nexB/scancode-toolkit/issues/3024 for details

Reference: https://github.com/nexB/scancode.io/issues/762
Reference: https://github.com/nexB/scancode.io/issues/1194
Reference: https://github.com/nexB/scancode.io/issues/83

### DIFF
--- a/scanpipe/models.py
+++ b/scanpipe/models.py
@@ -1909,6 +1909,17 @@ class CodebaseResourceQuerySet(ProjectRelatedQuerySet):
             return self.filter(~Q(status=status))
         return self.filter(status="")
 
+    def package_files(self):
+        """
+        Filter for CodebaseResources which are part of either an application
+        package or a system package.
+        """
+        from scanpipe.pipes import flag
+
+        return self.filter(
+            Q(status=flag.APPLICATION_PACKAGE) | Q(status=flag.SYSTEM_PACKAGE)
+        )
+
     def empty(self):
         return self.filter(Q(size__isnull=True) | Q(size=0))
 
--- a/scanpipe/pipelines/docker.py
+++ b/scanpipe/pipelines/docker.py
@@ -33,6 +33,7 @@ class Docker(RootFS):
         return (
             cls.extract_images,
             cls.extract_layers,
+            cls.extract_archives,
             cls.find_images_os_and_distro,
             cls.collect_images_information,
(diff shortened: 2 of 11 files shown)


pombredanne (changes requested):
Thanks!
This feature's step should be optional, e.g., a "group"

Issue #1148, opened 2024-04-04
ScanCode.io: Design and implement discrepancies reporting


pombredanne:
We need to create data models in ScanCode.io to track d2d discrepancies and incorporate these into the binary to source pipelines. This could be flexible and be part of a general discrepancy reporting and does not have to be dedicated only to d2d.

TG1999:
@pombredanne please give me design on this when you get chance.

TG1999:
This is done. We now have:

an endpoint on matchcodeio /api/d2d to run d2d and report discrepancies
This PR has been merged

https://github.com/nexB/purldb/pull/433
To test this feature:

Install PurlDB and matchcode.io from the main branch
Run
`make dev`
`make run_matchcodeio`

Make a request to /api/d2d

and enter Input URLs

For example these:

https://github.com/nexB/scancode.io/raw/main/scanpipe/tests/data/d2d-elfs/from-data.zip#from

https://github.com/nexB/scancode.io/raw/main/scanpipe/tests/data/d2d-elfs/to-data.zip#to

![Screenshot from 2024-05-09 14-14-17](https://github.com/nexB/purldb/assets/34160672/da526ca9-6a8c-4883-951e-26e92597ce0d)
Then click POST button and do refresh.

![Screenshot from 2024-05-24 11-27-14](https://github.com/nexB/scancode.io/assets/34160672/9718d089-e939-4433-b80a-f0db5d67ac7e)

The `codebase_resources_discrepancies` field indicates the number of resources that have discrepancies.

pombredanne:
This looks good! closing now.

PR #1218, opened 2024-05-08
Add requires-review tag for resources not mapped

### LINKED ISSUE #1223: Add requires-review tag for resources not mapped
Currently if any path is not mapped in d2d it's not tagged as requires-review

TG1999:
closes #1223

Reference #1148


scanpipe/pipes/d2d.py
```diff
@@ -1694,6 +1694,7 @@ def map_paths_resource(
                 if rel_key not in relations_to_create:
                     relations_to_create[rel_key] = relation
         if paths_not_mapped:
+            to_resource.update(status=flag.REQUIRES_REVIEW)
             to_resource.save()
```
tdruez, line 1698: Since a save() is triggered right after the update, wouldn't it be better to simply set the value?
`to_resource.status = flag.REQUIRES_REVIEW`

Issue #1237, opened 2024-05-16
SCIO: Resolve npm dependencies based on lockfiles


pombredanne:
I would like to have a pipeline (likely an addon pipeline) that can resolve the dependencies given this context:
- We have a project with a package.json and a package-lock.json or yarn.lock (and the future pnpm)
- We should be able to create the new "resolved" packages and the relationship using the new models and this data
  - nexB/scancode.io#1066

We also need the same elsewhere:
- [x] nexB/scancode-toolkit#3780

AyanSinhaMahapatra:
This is completed in https://github.com/nexB/scancode.io/pull/1244, closing.

Issue #1201, opened 2024-05-06
SCIO: Lock and resolve locked npm dependencies from yarn and npm


pombredanne:
We should be able to:
- [x] Resolve locked npm dependencies from yarn and npm lockfiles https://github.com/nexB/scancode.io/pull/1244
- [x] Generate lockfiles for npm and yarn if they are missing https://github.com/nexB/dependency-inspector/pull/7
- [x] Also support pnpm lockfiles resolution https://github.com/nexB/scancode-toolkit/issues/3766

AyanSinhaMahapatra:
This is completed, see [this comment](https://github.com/nexB/scancode.io/issues/1201#issue-2280203996) for the details.

Issue #1066, opened 2024-01-31
RFC: Design improved relationships for Package, Dependencies, Embeds and Requirements 


pombredanne:
We need to design improved models for Package, Dependencies, Embeds and Requirements. 

Some background:

- What we routinely calls "dependencies" are not really dependencies: more often these are requirements that are "resolved" to actual concrete dependencies.  To really "depend" on a software package, you need to know the exact resolved version. On the other hand, a requirement expresses the need for some range of versions (even if it's a range of a single version).

- We should distinguish a "require" relationship between packages vs. an actual "resolved" dependency to a concrete version, and track if the resolved dependency is present in the codebase or not.
 
- Dependencies (or requirements) may be only documented in a package manifest (like pom.xml) or a lockfile (like package-lock.json), or also be present in the codebase.

- We need to track also when a package contains another package (or for that matter when an app contains or embeds packages). 

Therefore we may have three levels of tree-like data we need to track for packages:
- "potential" requirements (we call these dependencies today). This is a logically a flat list of PURLs with VERS attached to a package instance. 
- "resolved" dependencies (we call these dependencies today too, but sometimes also packages). This is logically a tree (or graph) of PURLs with a resolved version. Each such dependency is derived and satisfies one or more "requirement".   
- "embeds" are packages embedded in other packages, e.g., they are physically copied, vendor and or built into a product codebase. These could have been the results of provisioning/fetching resolved dependencies, or could be a straight copy. These are not much tracked as such today.

We have agreed on some key model updates, based on discussions here and in our weekly calls:
- [x] Enhance ScanCode.io models to support relating dependencies to packages
- [x] Enhance ScanCode Toolkit models to support relating dependencies to packages
- [ ] nexB/dejacode#138
- [ ] nexB/purldb#485
-  ~~Enhance python-inspector models to support relating dependencies to packages using the new models~~
- ~~Enhance nuget-inspector models to support relating dependencies to packages using the new models~~
- ~~Create dependency-inspector to relate dependencies to packages using the new models~~

The https://github.com/nexB/dependency-inspector tickets to create lockfiles are:
- [x] https://github.com/nexB/dependency-inspector/issues/3
- [x] https://github.com/nexB/dependency-inspector/issues/5
- [x] https://github.com/nexB/dependency-inspector/issues/4
- [x] https://github.com/nexB/dependency-inspector/issues/12
- [x] https://github.com/nexB/dependency-inspector/issues/10
[...]

PR #1244, opened 2024-05-22
Resolve dependencies from lockfiles

AyanSinhaMahapatra:
This PR ensures we provide general support for dependency resolution in ScanCode.io:
- https://github.com/nexB/scancode.io/issues/1237
- https://github.com/nexB/scancode.io/issues/1279
- https://github.com/nexB/scancode.io/issues/1263
- https://github.com/nexB/scancode.io/issues/1278
- https://github.com/nexB/scancode.io/issues/1262
- https://github.com/nexB/scancode.io/issues/1201

This is the actual implementation of:
- https://github.com/nexB/scancode.io/issues/1066


scanpipe/pipes/__init__.py
```diff
@@ -194,6 +199,9 @@ def update_or_create_package(project, package_data, codebase_resources=None):
         package = DiscoveredPackage.create_from_data(project, package_data)
 
     if package:
+        if is_virtual:
+            package.update(is_virtual=is_virtual)
```
tdruez, line 203: Why not including this in package_data directly?

tdruez (changes requested):

scanpipe/models.py
```diff
@@ -3584,10 +3609,36 @@ def create_from_data(
         return cls.objects.create(
             project=project,
             for_package=for_package,
+            resolved_to_package=resolved_to_package,
             datafile_resource=datafile_resource,
             **cleaned_data,
         )
 
+    @classmethod
+    def extract_purl_data(cls, dependency_data, ignore_nulls=False):
```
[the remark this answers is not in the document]
AyanSinhaMahapatra: yes, extracting this to a different function.

scanpipe/models.py
```diff
@@ -1012,6 +1012,19 @@ def walk_codebase_path(self):
         """Return files and directories path of the codebase/ directory recursively."""
         return self.codebase_path.rglob("*")
 
+    def get_resource(self, path):
```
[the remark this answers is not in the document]
AyanSinhaMahapatra: This is a function of the Project class, and this has been done to make sure this mirrors the Codebase class on the SCTK side, as this is passed to individual package assembly functions. See
https://github.com/nexB/scancode.io/blob/main/scanpipe/pipes/scancode.py#L458
So this is not `resource_instance.get_resource()`, rather this is `project.get_resource()`

scanpipe/filters.py
```diff
@@ -761,6 +763,7 @@ class DependencyFilterSet(FilterSetUtilsMixin, django_filters.FilterSet):
     is_runtime = StrictBooleanFilter()
     is_optional = StrictBooleanFilter()
     is_resolved = StrictBooleanFilter()
+    is_direct = StrictBooleanFilter()
```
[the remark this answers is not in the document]
AyanSinhaMahapatra: Added this, thanks!

scanpipe/migrations/0061_dependency_resolver_update.py
```diff
@@ -0,0 +1,34 @@
...
+        migrations.AddField(
+            model_name="discoveredpackage",
+            name="is_virtual",
+            field=models.BooleanField(default=False),
+        ),
+        migrations.AddIndex(
+            model_name="discovereddependency",
+            index=models.Index(
+                fields=["is_direct"], name="scanpipe_di_is_dire_6dc594_idx"
+            ),
```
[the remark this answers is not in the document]
AyanSinhaMahapatra: I thought this was just used to display fields on the list view, but it seems that is not the case, adding these too.

scanpipe/models.py
```diff
@@ -3454,6 +3463,7 @@ class DiscoveredDependency(
     is_runtime = models.BooleanField(default=False)
     is_optional = models.BooleanField(default=False)
     is_resolved = models.BooleanField(default=False)
+    is_direct = models.BooleanField(default=False)
```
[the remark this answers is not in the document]
AyanSinhaMahapatra: Added, thanks!

AyanSinhaMahapatra:
@tdruez ready for review again!

scanpipe/models.py
```diff
@@ -3559,6 +3577,13 @@ def create_from_data(
                     package_uid=for_package_uid
                 )
 
+        if not resolved_to_package:
+            resolved_to_uid = dependency_data.get("resolved_to_uid")
+            if resolved_to_uid:
+                resolved_to_package = project.discoveredpackages.get(
+                    package_uid=resolved_to_uid
+                )
```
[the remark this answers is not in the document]
tdruez: Any update on this?

scanpipe/models.py
```diff
@@ -3190,15 +3207,7 @@ def purl(self):
...
-        purl_data = {}
-
-        for field_name in PURL_FIELDS:
-            value = package_data.get(field_name)
-            if field_name == "qualifiers":
-                value = normalize_qualifiers(value, encode=True)
-            purl_data[field_name] = value or ""
-
-        return purl_data
+        return normalize_package_url_data(package_data)
```
tdruez, line 3210: Do we need to keep the extract_purl_data since you refactored into an external function?

scanpipe/models.py
```diff
@@ -3694,6 +3746,25 @@ def as_spdx(self):
         )
 
 
+def normalize_package_url_data(purl_mapping, ignore_nulls=False):
```
tdruez, line 3749: Missing a uni test.

scanpipe/pipes/__init__.py
```diff
@@ -254,27 +260,65 @@ def update_or_create_dependency(
...
+        if dependency:
+            dependencies.append(dependency)
+
+    for dependency in dependencies:
+        # This dependency relationship is for a new package
+        if (
+            for_package
+            and dependency.for_package
+            and dependency.for_package != for_package
+        ):
```
tdruez, line 295: Instead of a comment, put the condition in a variable with an explicit name, for example, `is_for_new_package`.

scanpipe/pipes/scancode.py
```diff
@@ -516,6 +515,145 @@ def process_package_data(project):
...
+                        break
+
+            dependencies = package_mapping.get("dependencies") or []
+            update_packages_and_dependencies(
+                project=project,
+                dependencies=dependencies,
+                package=package,
+                resource=resource,
+                datasource_id=pd.datasource_id,
+            )
```
tdruez, line 547: That could be refactored as a function

scanpipe/tests/test_pipelines.py
```diff
@@ -569,15 +569,23 @@ def _normalize_package_uids(self, data):
                 if isinstance(value, (list, dict)):
                     value = self._normalize_package_uids(value)
                 if (
-                    key in ("package_uid", "dependency_uid", "for_package_uid")
+                    key
+                    in (
+                        "package_uid",
+                        "dependency_uid",
+                        "for_package_uid",
+                        "resolved_to_package_uid",
```
tdruez, line 577: Probably time to put this in a variable.

tdruez (changes requested):
@AyanSinhaMahapatra That looks good, see my minor suggestions.
Also, do you think we have a good enough test coverage for the new resolution features?

scanpipe/models.py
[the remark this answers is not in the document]
AyanSinhaMahapatra: So to summarize: I think it is guaranteed here, but there were other places where there was similar issues, which I resolved.

The `get()` function you highlighted is only called if the for/resolved packages are not present and if there is an uid for these in the dependency data.
The `DiscoveredDependency.create_from_data` function is only called in `update_or_create_dependency` function, and this is called in a couple places:
1. the new resolve dependencies code (here we never have the uids in the data, but we call the `populate_dependency_uuid` before calling this)
2. loading from a SCTK/SCIO scan (this is guaranteed here, and also when we do a assembly on SCTK we are generating unique uids)
3. other places in the code (this is also guaranteed)

scanpipe/models.py
AyanSinhaMahapatra: We are using `DiscoveredDependency.extract_purl_data()` in 2 places, and also `DiscoveredPackage.extract_purl_data()` in 2 places. The `DiscoveredPackage` method is exactly `models.normalize_package_url_data()` but the `DiscoveredDependency` method has some extra modifications, so I was looking to keep the same method for purl related fields.

scanpipe/models.py
AyanSinhaMahapatra: Added, thanks!

scanpipe/pipes/__init__.py
AyanSinhaMahapatra: Added, thanks!

scanpipe/pipes/scancode.py
AyanSinhaMahapatra: Done, thanks!

scanpipe/pipes/__init__.py
```diff
@@ -254,27 +260,47 @@ def update_or_create_dependency(
...
+            dependency = DiscoveredDependency.create_from_data(
+                project,
+                dependency_data,
+                for_package=for_package,
+                resolved_to_package=resolved_to_package,
+                datafile_resource=datafile_resource,
+                datasource_id=datasource_id,
+                strip_datafile_path_root=strip_datafile_path_root,
+            )
+            break
```
tdruez, line 288: Is the `break` intentional here? `break` exits the entire loop while `continue` proceed to the next iteration.

scanpipe/pipes/__init__.py
```diff
@@ -283,6 +309,35 @@ def update_or_create_dependency(
...
+def get_dependencies(project, dependency_data):
+    """
+    Given a `dependency_data` mapping, get a list of DiscoveredDependency objects
+    for that `project` with similar dependency data.
+    """
+    dependency = None
+    dependency_uid = dependency_data.get("dependency_uid")
+    extracted_requirement = dependency_data.get("extracted_requirement")
+    if not extracted_requirement:
+        extracted_requirement = ""
```
tdruez, line 321:
```
extracted_requirement = dependency_data.get("extracted_requirement") or ""
```

scanpipe/pipes/scancode.py
```diff
@@ -482,40 +485,210 @@ def process_package_data(project):
...
+            # of a package is present for an environment
+            if not other_dependencies:
+                other_dependencies = [
+                    dep
+                    for dep in matched_dependencies
+                    if (
+                        dep.base_purl == dependency.base_purl
+                        and dep.resolved_to_package
+                    )
+                ]
```
tdruez, line 679: The var naming is not great and makes the code hard to read: eg: `dep.purl != dependency.purl`

tdruez (changes requested):
@AyanSinhaMahapatra Looks good, almost ready for merge.
See a few minor change requests.

We are missing a changelog entry that summarize all the changes of this PR.
It's quite an important one.

scanpipe/pipes/__init__.py
AyanSinhaMahapatra: Added, thanks!

scanpipe/pipes/scancode.py
AyanSinhaMahapatra: Renaming `dep` to `matched_dependency`, thanks!

scanpipe/pipes/__init__.py
AyanSinhaMahapatra: Yes, this is intentional to put `break` here. In the case where this dependency data is for a new package, we want to create a new `DiscoveredDependency` object with all the attributes properly populated and exit from this loop, as there is no point continuing anymore.

Issue #1287, opened 2024-06-24
purldb scan worker: have worker thread send scan results back to purldb


JonoYang:
When using the purldb scan worker command, the main worker loop gets the next job from purldb and creates a scan project for it. It then polls the project until it is complete and then sends the results back to purldb. It would be better for the main worker loop to just collect new download urls from purldb and whenever we create a new scan project, we append a pipeline at the end of the list of pipelines to be run that sends the results back to purldb. This way, the main worker loop does not need to poll and see if the scan project is done. It would also behoove us to implement some scan counter to ensure we do not have more than a few package scans running at the same time.

JonoYang:
We now use the webhook subscriptions to send package scan results back to purldb. A new webhook endpoint has been created on purldb for this purpose.

PR #1320, opened 2024-07-09
1287 purldb scan worker update

JonoYang:
This PR updates the `purldb-scan-worker` management command to send project results back using a new addon pipeline, `send_project_results`. This pipeline is appended to the list of pipelines to be run on a package, allowing the worker that handles the project execution to send back the data when all of the pipelines have been run. This allows us to not have the main task loop to monitor a single project run for completion in order to return data, and allows us to request multiple scan projects and work on them simultaneously. A new option, `--max-concurrent-projects`, has been added to set the number of purldb packages that can be requested and processed at once.


tdruez:
@JonoYang The code looks fine but I don't feel great about adding a specialized pipeline only useful in the context of the PurlDB worker.

Couldn't we use the existing Webhook system for this instead?

JonoYang:
@tomd I forgot about the webhook functions on the Project model. I will revise my code to use that instead.

JonoYang:
@tomd I am looking at the webhook code and its relation to Projects and Runs. I see that in https://github.com/nexB/scancode.io/blob/main/scanpipe/tasks.py#L72 that we send back the project results after every Run on a Project. I would like to send back the results after all of the requested pipelines are run, as is the behavior of my code that adds the `send_project_results` pipeline. I have the idea to add a flag or some other attribute on Project to tell scancode.io when to process webhooks. I would like your opinion on the best way of triggering a webhook reply once all Project Runs are completed.

tdruez:
>  I see that in https://github.com/nexB/scancode.io/blob/main/scanpipe/tasks.py#L72 that we send back the project results after every Run on a Project. I would like to send back the results after all of the requested pipelines are run,

Yes, I think that's a good opportunity to improve the Webhook system.
We could add a flag as you suggested to control if the webhook should be sent after each pipeline completion, or only at the end. 
We need to take into account that a pipeline (last or not) may fail and we need to decide what should happen to the Wehbook in that case.
Also, we need to define the default behavior. Posting only when all pipelines have been completed as default probably makes the most sense. With maybe a `send_on_failure` flag enabled by default.

tdruez:
@JonoYang Added ``results_url`` and ``summary_url`` in the Webhook payload in https://github.com/nexB/scancode.io/pull/1326/files 

> We could add a flag as you suggested to control if the webhook should be sent after each pipeline completion, or only at the end.

Actually in the short term, since the `next_run` value is available in the payload, could your receiving view check if all the runs have completed to trigger the import process?

JonoYang:
> We need to take into account that a pipeline (last or not) may fail and we need to decide what should happen to the Wehbook in that case. Also, we need to define the default behavior. Posting only when all pipelines have been completed as default probably makes the most sense. With maybe a `send_on_failure` flag enabled by default.

For what it's worth, in the context of the `purldb-scan-worker` command, if any of the requested pipelines failed, then we consider the entire Project to have failed and then send back a response to the purldb with that status.

> Actually in the short term, since the next_run value is available in the payload, could your receiving view check if all the runs have completed to trigger the import process?

Do you mean having the receiver (purldb) poll to see if the project is done? The endpoint that issues and handles the package scan requests on purldb currently does not poll outstanding package scan requests, but just waits for updates from the worker.

tdruez:
@JonoYang The code looks good! Is this ready for merge or do you plan to add more?

JonoYang:
I have nothing else to add. This is good to go for me

tdruez:
Meged! @JonoYang feel free to close https://github.com/nexB/scancode.io/issues/1287 if completed 🚀

PR #1332, opened 2024-07-22
Refactor pipeline to extract reusable basepipeline in module and library 

tdruez:
This PR refactors the pipelines architecture to extract a reusable BasePipeline in its own module and also available as standalone library

This is going to support adopting pipelines in VulnerableCode and PurlDB

Reference:
- https://github.com/nexB/scancode.io/issues/1351
- https://github.com/nexB/vulnerablecode/issues/1509
- https://github.com/nexB/purldb/issues/521

### DIFF
--- a/pipeline/__init__.py
+++ b/pipeline/__init__.py
@@ -1,0 +1,226 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# http://nexb.com and https://github.com/nexB/scancode.io
+# The ScanCode.io software is licensed under the Apache License version 2.0.
+# Data generated with ScanCode.io is provided as-is without warranties.
+# ScanCode is a trademark of nexB Inc.
+#
+# You may not use this software except in compliance with the License.
+# You may obtain a copy of the License at: http://apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software distributed
+# under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations under the License.
+#
+# Data Generated with ScanCode.io is provided on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, either express or implied. No content created from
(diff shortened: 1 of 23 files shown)


tdruez:
The code in this branch is currently experimental and not ready for merge/production.

pipeline/__init__.py
```diff
@@ -0,0 +1,330 @@
+# SPDX-License-Identifier: Apache-2.0
```
pombredanne, line 1: the "pipeline" module name needs to be changed to something better that will not conflict when installed with other libraries.

I suggest either:

- using a namespace (we never did before, but we could start). This would mean creating a top level `aboutcode` directory with no `__init__.py` and a few more details TBD. This is PEP420 https://peps.python.org/pep-0420/ and https://docs.python.org/3/glossary.html#term-namespace-package  and this namespace would become reusable across other projects for shared modules. This is OK because we own aboutcode on PyPI and none of our projects ever used the `aboutcode` module name. Some exmaples include jaraco's repos at https://github.com/jaraco/jaraco.vcs https://github.com/jaraco/jaraco.abode and https://github.com/jaraco/jaraco.context all living in the same "jaraco" namespace.
In this case I would suggest that we call the package pipeline under aboutcode namespace. And end up with this path `aboutcode/pipeline/__init__.py`

- using another name for the package directory that contains `__init__.py` and use that as a regular, non-namespaced module. For instance `aboutcode_pipeline/__init__.py`
 
There are no specific benefits nor issues with the second approach. Using the first approach with a namespace would be the pythonic way.

We should  name the wheel `aboutcode.pipeline` to make it unique.

pombredanne (changes requested):

tdruez:
This PR is already too complex combining the module refactoring, the Pipeline classes refactoring, packaging logic, and new CI for release publishing.

Let's split this into manageable and simpler-to-review parts:
- [ ] Move the current BasePipeline class to a new `aboutcode.pipeline` module without any code logic changes. https://github.com/nexB/scancode.io/pull/1357
- [ ] Refactor the BasePipeline to remove the scanpipe.Project logic and make it reusable without Project context.
- [ ] Add the packaging and release system

PR #1338, opened 2024-07-24
Webhook behavior customization and delivery records

tdruez:
This PR expands on the existing WebhookSubscription model by adding a few fields to configure the behavior of the Webhook, and moving some of the fields to a new WebhookDelivery model, which captures the results of a WebhookSubscription "delivery".

@JonoYang Could you give it a try and see if the current addition would work for your needs in the context of the PurlDB workers?

For example, adding a Webhook from the code:

```
project.add_webhook_subscription(
    target_url="URL",
    trigger_on_each_run=False,
    include_summary=True,
    include_results=True,
)
```

Also available using the API:
```
data = {
    "name": "Project1",
    "webhooks": [
        {
            "target_url": "URL",
            "trigger_on_each_run": False,
            "include_summary": True,
            "include_results": True,
            "is_active": True,
        }
    ],
}
```


scanpipe/models.py
```diff
@@ -3870,36 +3934,135 @@ def get_payload(self, pipeline_run):
...
+            project=self.project,
+            webhook_subscription=self,
+            run=pipeline_run,
+            target_url=self.target_url,
+            payload=payload,
+        )
 
-        logger.info(f"Sending Webhook uuid={self.uuid}.")
         try:
             response = requests.post(
```
JonoYang, line 3966: I am having trouble sending the project results back to purldb because the authentication token is not used here.

JonoYang (changes requested):
The updates look good and I'm able to use the new webhook subscriptions, but I'm running into a problem with authentication when sending the results back to purldb.

PR #1389, opened 2024-09-23
Regarding globbing-support

VarshaUN:
regarding the issue https://github.com/aboutcode-org/scancode.io/issues/1347    i have tested some testcases.

Signed-off by: VarshaUN <varshaun58@gmail.com>


pombredanne (changes requested):
Thanks! this is a good start. See how we have test functions in other test code to improve this. The important thing to test is that multiple licenses are matched.

You can then move on towards an implementation.
Then you may want to check also how policies would be impacted in ScanCode Toolkit and DejaCode

keshav-space:
@VarshaUN see some suggestion here https://github.com/aboutcode-org/scancode.io/issues/1347#issuecomment-2383575395

VarshaUN:
Closing this PR as I am freshly working with new fork. Thank you

PR #1400, opened 2024-10-07
Add pipeline to publish scan to federatedcode

keshav-space:
- Closes https://github.com/aboutcode-org/federatedcode/issues/23


setup.cfg
```diff
@@ -105,6 +105,8 @@ install_requires =
     bleach==6.1.0
     # Antivirus
     clamd==1.0.2
+    # FederatedCode
+    aboutcode.hashid>=0.1.0
```
tdruez, line 109:
```suggestion
    aboutcode.hashid==0.1.0
```

tdruez:
@keshav-space Could you provide some context about the need for adding a `project_purl` field? Why not use the `uuid` for example?
It seems to me that this not directly related, and the addition of new concept such as this one should be discussed and handle separatly.

scancodeio/settings.py
```diff
@@ -418,3 +418,10 @@
 MATCHCODEIO_USER = env.str("MATCHCODEIO_USER", default="")
 MATCHCODEIO_PASSWORD = env.str("MATCHCODEIO_PASSWORD", default="")
 MATCHCODEIO_API_KEY = env.str("MATCHCODEIO_API_KEY", default="")
+
+# FederatedCode integration
+
+FEDERATEDCODE_GIT_ACCOUNT = env.str("FEDERATEDCODE_GIT_ACCOUNT", default="")
+FEDERATEDCODE_GIT_SERVICE_TOKEN = env.str("FEDERATEDCODE_GIT_SERVICE_TOKEN", default="")
+FEDERATEDCODE_GIT_SERVICE_NAME = env.str("FEDERATEDCODE_GIT_SERVICE_NAME", default="")
+FEDERATEDCODE_GIT_SERVICE_EMAIL = env.str("FEDERATEDCODE_GIT_SERVICE_EMAIL", default="")
```
tdruez, line 427: This needs documentation.

scanpipe/pipelines/publish_to_federatedcode.py
```diff
@@ -0,0 +1,116 @@
...
+            cls.get_project_purl,
+            cls.get_package_repository,
+            cls.clone_repository,
+            cls.add_scan_result,
+            cls.commit_and_push_changes,
+            cls.delete_local_clone,
+        )
+
+    def get_project_purl(self):
+        """Get the PURL for the project."""
```
tdruez, line 48: This does way more than getting a purl.

scanpipe/pipelines/publish_to_federatedcode.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+from packageurl import PackageURL
+
+from scanpipe.pipelines import Pipeline
+from scanpipe.pipes import federatedcode
+
+
+class PublishToFederatedCode(Pipeline):
+    """Publish package scan to FederatedCode."""
```
tdruez, line 31: Documentation needs more substance.

scanpipe/pipes/federatedcode.py
```diff
@@ -0,0 +1,132 @@
...
+        missing_vars.append("FEDERATEDCODE_GIT_SERVICE_TOKEN")
+    if not settings.FEDERATEDCODE_GIT_SERVICE_NAME:
+        missing_vars.append("FEDERATEDCODE_GIT_SERVICE_NAME")
+    if not settings.FEDERATEDCODE_GIT_SERVICE_EMAIL:
+        missing_vars.append("FEDERATEDCODE_GIT_SERVICE_EMAIL")
+
+    if missing_vars:
+        return False, f'Missing environment variables: {", ".join(missing_vars)}'
+
+    return True, ""
```
tdruez, line 53: This is not consistent with existing integrations.

tdruez (changes requested):
Missing documentation about the concept of publishing to federatedcode.

keshav-space:
@tdruez 

> Could you provide some context about the need for adding a `project_purl` field?

We want to store the scancode.io scan results in git repositories, and use PURL to determine the git repository and the exact directory path where the scan should be stored. This optional `project_purl` field would be needed to push the final scan results to FederatedCode.


> Why not use the `uuid` for example? 

Project uuid would be specific to a particular scancode.io instance. We want to store package scan/vulnerability data in a way that it can be retrieved using just the PURL, which won't be possible with uuid.


> It seems to me that this is not directly related, and the addition of a new concept such as this one should be discussed and handled separately. 

Sure, let's discuss this and we can split this into two different PRs.


> Also, we recently introduced `name` and `version` fields for the project. Allowing for a manually provided PURL does not take into consideration those fields.

My understanding was that the product name and product version were closely related to DejaCode.

tdruez:
> We want to store the scancode.io scan results in git repositories, and we use PURL to determine the git repository and the exact directory path where the scan should be stored. This optional project_purl field would be needed to push the final scan results to FederatedCode.

This should be documented in the code.

> My understanding was that the product name and product version were closely related to DejaCode.

You're right, this seems quite untreated.

scancodeio/settings.py
keshav-space: Done.

scanpipe/pipelines/publish_to_federatedcode.py
keshav-space: Done

scanpipe/pipelines/publish_to_federatedcode.py
keshav-space: Yeah, this is misleading. I've renamed this step to `check_federatedcode_eligibility`.

pombredanne:
is there something missing to get this merged?

scanpipe/models.py
```diff
@@ -561,6 +561,11 @@ class Project(UUIDPKModel, ExtraDataFieldMixin, UpdateMixin, models.Model):
     notes = models.TextField(blank=True)
     settings = models.JSONField(default=dict, blank=True)
     labels = TaggableManager(through=UUIDTaggedItem)
+    project_purl = models.CharField(
```
tdruez, line 564: Why the `project` prefix for this field? Why not simply `purl`?

scanpipe/models.py
```diff
@@ -561,6 +561,11 @@ class Project(UUIDPKModel, ExtraDataFieldMixin, UpdateMixin, models.Model):
...
+        max_length=2048,
+        blank=True,
+        help_text=_("Project Package URL."),
```
tdruez, line 567: This is not helpful, we should instead document the purpose of this field and how it differs from  the usual "Package URL".

scanpipe/pipelines/publish_to_federatedcode.py
```diff
@@ -0,0 +1,131 @@
...
+        if not source_is_download_url:
+            raise Exception("Project input should be download_url.")
+
+        if not self.project.project_purl:
+            raise Exception("Missing Project PURL.")
+
+        project_package_url = PackageURL.from_string(self.project.project_purl)
+
+        if not project_package_url.version:
+            raise Exception("Missing version in Project PURL.")
```
tdruez, line 90: This logic should be refactored into a pipe function to keep the pipeline class simple and readable.

tdruez (changes requested):

scanpipe/models.py
```diff
@@ -562,6 +562,14 @@ class Project(UUIDPKModel, ExtraDataFieldMixin, UpdateMixin, models.Model):
     notes = models.TextField(blank=True)
     settings = models.JSONField(default=dict, blank=True)
     labels = TaggableManager(through=UUIDTaggedItem)
+    purl = models.CharField(
+        max_length=2048,
+        blank=True,
+        help_text=_(
+            "Package URL for the project, used for pushing project scan result to "
+            "FederatedCode. This should be the PURL of the input."
```
tdruez, line 570: > This should be the PURL of the input.

Could you clarify what this means?

scanpipe/templates/scanpipe/project_settings.html
```diff
@@ -26,6 +26,14 @@
...
                   </div>
                 </div>
+                <div class="field">
+                  <label class="label" for="{{ form.name.id_for_label }}">
+                    PURL
+                  </label>
+                  <div class="control">
+                    {{ form.purl }}
+                  </div>
+                </div>
```
tdruez, line 36: Missing help about the purpose of this field and placeholder example.

scanpipe/models.py
keshav-space: > Could you clarify what this means?

Suppose we have https://registry.npmjs.org/lodash/-/lodash-4.17.21.tgz as the input, then the PURL should be pkg:npm/lodash@4.17.21.

scanpipe/tests/pipes/test_federatedcode.py
```diff
@@ -0,0 +1,77 @@
...
+        repo = git.Repo.init(local_dir)
+
+        federatedcode.add_scan_result(
+            self.project1, repo, Path("repo/npm/foobar/v1.2.3/scancodeio.json")
+        )
+
+        self.assertIn("npm/foobar/v1.2.3/scancodeio.json", repo.untracked_files)
+        shutil.rmtree(repo.working_dir)
+
+    def test_scancpipe_pipes_federatedcode_delete_local_clone(self):
```
tdruez, line 72: Typo

scanpipe/forms.py
```diff
@@ -480,12 +481,31 @@ class Meta:
...
+        """Validate the Project PURL."""
+        purl = self.cleaned_data.get("purl")
+
+        if purl:
+            try:
+                PackageURL.from_string(purl)
+            except ValueError:
+                raise forms.ValidationError("PURL must be a valid PackageURL")
+
+        return purl
```
tdruez, line 507: Could you add a simple unit test for this?

scanpipe/models.py
tdruez: What about:
```
help_text=_(
    "Package URL (PURL) for the project, required for pushing the project's scan result to FederatedCode. "
    "For example, if the input is an input URL like https://registry.npmjs.org/lodash/-/lodash-4.17.21.tgz, "
    "the corresponding PURL would be pkg:npm/lodash@4.17.21."
)
```

tdruez (changes requested):
@keshav-space The code looks pretty good, we are almost ready for merge, see the last few comments.

Also, is this https://github.com/aboutcode-org/scancode.io/pull/1400/files#diff-71c80d25cae67eed0aa112b1d847002632d97e7f223d9df6109d39d9e26bc577 a wanted change? That file is needed for proper packaging.

scanpipe/forms.py
keshav-space: Done

keshav-space:
> is this https://github.com/aboutcode-org/scancode.io/pull/1400/files#diff-71c80d25cae67eed0aa112b1d847002632d97e7f223d9df6109d39d9e26bc577 a wanted change?

@tdruez  yes, namespace package directory should not contain `__init__.py` https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#native-namespace-packages.

tdruez:
> @tdruez yes, namespace package directory should not contain __init__.py

Fair enough, but it seems quite unrelated to the context of this PR. It would be better to open an issue for discussion.

keshav-space:
> Fair enough, but it seems quite unrelated to the context of this PR. It would be better to open an issue for discussion.

It is related to this PR because the pipeline uses another namespace package, `aboutcode.hashid`. If we place an empty `__init__.py` in our local `aboutcode` directory, the resolution for all the `aboutcode` namespace packages will fail.

```python
  File "/scancode.io/scanpipe/pipelines/publish_to_federatedcode.py", line 25, in <module>
    from scanpipe.pipes import federatedcode
  File "/scancode.io/scanpipe/pipes/federatedcode.py", line 35, in <module>
    from aboutcode import hashid
ImportError: cannot import name 'hashid' from 'aboutcode' (/scancode.io/aboutcode/__init__.py)
make: *** [Makefile:126: test] Error 1
```

tdruez:
Thanks for the clarification. Let's merge then!

Issue #1426, opened 2024-10-31
Adding projects through the API runs all groups


AyanSinhaMahapatra:
Sending the following scan to a scancode.io instance triggers a scan with all the groups of the selected pipeline:
```
import requests

api_url = "http://localhost/api/projects/"
data = {
    "name": "worksbench_via_api",
    "input_urls": "https://github.com/aboutcode-org/scancode-workbench/archive/refs/tags/v4.0.3.tar.gz",
    "pipeline": "inspect_packages",
    "execute_now": True,
}
response = requests.post(api_url, data=data)
response.json()
```

From the log:
```
2024-10-31 14:48:35.585 Pipeline [inspect_packages] starting
2024-10-31 14:48:35.587 Step [download_missing_inputs] starting
2024-10-31 14:48:35.588 Fetching input from https://github.com/aboutcode-org/scancode-workbench/archive/refs/tags/v4.0.3.tar.gz
2024-10-31 14:48:40.984 Step [download_missing_inputs] completed in 5 seconds
2024-10-31 14:48:40.986 Step [copy_inputs_to_codebase_directory] starting
2024-10-31 14:48:41.013 Step [copy_inputs_to_codebase_directory] completed in 0 seconds
2024-10-31 14:48:41.016 Step [extract_archives] starting
2024-10-31 14:48:42.888 Step [extract_archives] completed in 2 seconds
2024-10-31 14:48:42.891 Step [collect_and_create_codebase_resources] starting
2024-10-31 14:48:44.034 Step [collect_and_create_codebase_resources] completed in 1 seconds
2024-10-31 14:48:44.036 Step [flag_empty_files] starting
2024-10-31 14:48:44.040 Step [flag_empty_files] completed in 0 seconds
2024-10-31 14:48:44.043 Step [flag_ignored_resources] starting
2024-10-31 14:48:44.044 Step [flag_ignored_resources] completed in 0 seconds
2024-10-31 14:48:44.046 Step [scan_for_application_packages] starting
2024-10-31 14:48:44.156 Progress: 10% (44/440) ETA: 1 seconds
2024-10-31 14:48:44.172 Progress: 20% (88/440)
2024-10-31 14:48:44.192 Progress: 30% (132/440)
2024-10-31 14:48:44.219 Progress: 40% (176/440)
2024-10-31 14:48:44.246 Progress: 50% (220/440)
2024-10-31 14:48:44.273 Progress: 60% (264/440)
[...]

PR #1442, opened 2024-11-07
Refactor the `group` decorator for pipeline steps as `optional_step` …

tdruez:
…#1426


AyanSinhaMahapatra:
@tdruez I've fixed the tests by adding explicit options for pipeline integration tests and updated some docs too.
This is probably much cleaner and faster than (changing instances in code is probably not required as most of these are not user facing anyway) the approach at https://github.com/aboutcode-org/scancode.io/commit/55072281b4702f9a8417557296023fdca0c2452f#diff-4b8ad9b60fea03b9d86372b709ec186cf84bfb260e215d2f31c688fa3991d3bfL274 so this makes much more sense.

Thanks and this is ready for your review.

Issue #1454, opened 2024-11-26
Projects creation API does not support multiple pipeline options correctly


AyanSinhaMahapatra:
There is a bug preventing running a `map_deploy_to_develop` pipeline with multiple options enabled.

When trying to create a project from the API like this:
```
import requests

api_url = "http://127.0.0.1:8001/api/projects/"
data = {
    "name": "d2d_test_v3",
    "input_urls": [
        "https://repo1.maven.org/maven2/javax/portlet/portlet-api/3.0.1/portlet-api-3.0.1.jar#to",
        "https://repo1.maven.org/maven2/javax/portlet/portlet-api/3.0.1/portlet-api-3.0.1-sources.jar#from"
    ],
    "pipeline": ["map_deploy_to_develop:Java,JavaScript"],
    "execute_now": True,
}

response = requests.post(api_url, data=data)
response.json()
```

We get the following error:
```
{'pipeline': ['"JavaScript" is not a valid choice.']}
```

But if we use one option like the following:

- "pipeline": ["map_deploy_to_develop:Java"]
- "pipeline": ["map_deploy_to_develop:JavaScript"]

It runs the pipeline with the specified optional steps selected without any issues.
So the pipeline option specified do exist and specifying options work nicely through the API for a single pipeline option, but fails for multiple pipeline options.

This is across any pipelines, i.e. `"pipeline": ["resolve_dependencies:StaticResolver,DynamicResolver"],` fails too with `{'pipeline': ['"DynamicResolver" is not a valid choice.']}`.

The same issue is not present in the CLI. When I run `docker compose exec -it web scanpipe  create-project test-cli --pipeline resolve_dependencies:StaticResolver,DynamicResolver --input-url <INPUT_URL> --execute` it creates and runs the project with the specified selected options allright.

 Relevant docs: https://scancodeio.readthedocs.io/en/latest/rest-api.html#create-a-project

tdruez:
There is a syntax conflict as the comma `,` is used for both separating the pipeline names and optional steps.

`["map_deploy_to_develop:Java,JavaScript"]` ends up splitted as those 2 pipelines:
`['map_deploy_to_develop:Java', 'JavaScript']` where it should be a single pipeline with 2 optional steps.

We may have to drop the support for providing multiple pipelines as a comma-separated string, such as `"analyze_docker_image,scan_single_package"`, to only support providing a list. 
[...]

PR #1455, opened 2024-11-27
Fix the ability to provide multiple optional step in API #1454

tdruez:
Reference: https://github.com/aboutcode-org/scancode.io/issues/1454


AyanSinhaMahapatra (approved):
@tdruez Thanks a lot for the quick fix!
This fixes #1454

Issue #1435, opened 2024-11-04
back2source: Extend ScanCode.io pipeline for Rust source and binaries


pombredanne:
Create a new set of pipelines steps for the binary to source analysis for Rust source and binaries using strings and symbols

pombredanne:
@AyanSinhaMahapatra can you add a completion report there?

pombredanne:
The Rust option is now available and functional in the map-deploy-to_devel pipeline
 ![Image](https://github.com/user-attachments/assets/5b810737-d23b-41cc-a5df-e03f5bed5986)

AyanSinhaMahapatra:
This is now completed.

To complete this feature we had the following PRs/items:

- Rust source symbols parsers were already present in https://github.com/aboutcode-org/source-inspector
- https://github.com/aboutcode-org/source-inspector/pull/36 Added tests for Rust symbols parsing and released source-inspector
-  https://github.com/aboutcode-org/rust-inspector/pull/1 Get symbols from Rust elf binaries using blint (uses lief and symbolic internally)
- Released rust-inspector https://github.com/aboutcode-org/rust-inspector/releases/tag/v0.1.0
- https://github.com/aboutcode-org/scancode.io/pull/1488 Add a new pipeline step to match symbols from Rust source files to Rust binaries
- Support was also added to parse dependency information from rust binaries with a scancode plugin https://github.com/aboutcode-org/rust-inspector/releases/tag/v0.1.0

See the related pipe in SCIO documentation: https://scancodeio.readthedocs.io/en/latest/built-in-pipelines.html#scanpipe.pipelines.deploy_to_develop.DeployToDevelop.map_rust
This is available as a new optional step in the `deploy_to_develop` pipeline: `Rust` (you can select this through the UI or pass this as an option when using the API/CLI: `deploy_to_develop:Rust` )

Select the option `Rust` in the new project form like this:

![Image](https://github.com/user-attachments/assets/ea442796-f95a-4981-bcde-a855cd8487c5)

PR #1488, opened 2024-12-30
Add rust binary support #1435

AyanSinhaMahapatra:
Add support for parsing rust binaries (#1435):

* Add a new optional step (`Rust`) in the `deploy_to_develop` pipeline to map rust binaries to rust source files
* Add support for parsing rust binaries using `rust-inspector` at https://pypi.org/project/rust-inspector/0.1.0/
   * Add support for getting symbols from rust binaries (elf)
   * Add support for getting packages and dependencies from rust binaries (which are built using `cargo-auditable`)


scanpipe/pipes/symbolmap.py
```diff
@@ -0,0 +1,156 @@
...
+        else:
+            rel_key, relation = item
+            if rel_key not in relations_to_create:
+                relations_to_create[rel_key] = relation
+
+    # If there are any non-test files in the rust source files which
+    # are not mapped, we mark the binary as REQUIRES_REVIEW
+    if paths_not_mapped and any(
+        [True for path in paths_not_mapped if "/tests/" not in path]
+    ):
```
tdruez, line 70: Put the condition in a well-named variable for readability.

scanpipe/pipes/symbolmap.py
```diff
@@ -0,0 +1,156 @@
...
+        to_resource.status = flag.REQUIRES_REVIEW
+        to_resource.save()
```
tdruez, line 72: `to_resource.update(status=flag.REQUIRES_REVIEW)` would be a bit better, single line, and the `save(update_fields=)` is applied, which limit the save to a single field in that case.

scanpipe/pipes/symbolmap.py
```diff
@@ -0,0 +1,156 @@
...
+    stats = {
+        "common_symbols_unique_ratio": common_symbols_unique_ratio,
+        "common_symbols_ratio": common_symbols_ratio,
+    }
+
+    if common_symbols_ratio > 0.5 or common_symbols_unique_ratio > 0.5:
+        return True, stats
+    elif source_symbols_count > 20 and (
+        common_symbols_ratio > 0.4 or common_symbols_unique_ratio > 0.4
+    ):
```
tdruez, line 118: Using named-vairable for the condition could hep to understand this magic ;)

tdruez (approved):
The good looks good, I've added a few suggestions.
Also, make sure to always include new pipe and pipeline modules in the documentation.

scanpipe/pipes/symbolmap.py
AyanSinhaMahapatra: Thanks, this is added.

scanpipe/pipes/symbolmap.py
AyanSinhaMahapatra: Thanks, this is much better.

PR #1504, opened 2025-01-06
Do not fail project on memory spikes

AyanSinhaMahapatra:
We have an unresolved issued of multi-gigabyte memory spikes when scanning files for licenses in some rare cases, like it is reported at https://github.com/aboutcode-org/scancode-toolkit/issues/3711 Currently we fail the project on encountering such memory spikes, but we don't get any scan info from the rest of the project because we raise the Exception. This adds a project error instead of raising an Exception, so we can continue scanning the rest of the resources.

Reference: https://github.com/aboutcode-org/scancode-toolkit/issues/3711

### DIFF
--- a/scanpipe/pipes/scancode.py
+++ b/scanpipe/pipes/scancode.py
@@ -286,7 +286,11 @@ def save_scan_package_results(codebase_resource, scan_results, scan_errors):
 
 
 def scan_resources(
-    resource_qs, scan_func, save_func, scan_func_kwargs=None, progress_logger=None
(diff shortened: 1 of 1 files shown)


tdruez (commented):
@AyanSinhaMahapatra Why not use `resource.project` instead of adding a new argument?

AyanSinhaMahapatra:
@tdruez good point, I did not remember that we could reach the project through the resource. Let me update this to use `resource.project` instead, thanks!

PR #1508, opened 2025-01-06
Ignore large data files and bump scancode-toolkit

AyanSinhaMahapatra:
This PR:
* Updates `scancode-toolkit` to the latest release [`v32.3.2`](https://github.com/aboutcode-org/scancode-toolkit/releases/tag/v32.3.2)
* Rolls back the changes made in https://github.com/aboutcode-org/scancode.io/pull/1504 as this can be problematic at scale in multiprocessing context
* Adds a scancode.io settings field `SCANCODEIO_SCAN_MAX_FILE_SIZE` and a project settings `scan_max_file_size` to skip scanning files above a certain file size (in bytes) as a temporary fix for https://github.com/aboutcode-org/scancode-toolkit/issues/3711


scanpipe/pipes/scancode.py
```diff
@@ -58,6 +59,13 @@
 Utilities to deal with ScanCode toolkit features and objects.
 """
 
+# skip large data files which are causing memory spikes
+# See https://github.com/aboutcode-org/scancode-toolkit/issues/3711
+# TODO: this has to be removed once the issue is fixed
+SKIP_DATA_FILE_EXTENSIONS = [".json", ".xml", ".yaml", ".yml"]
+SKIP_DATA_FILE_SIZE = 1048576
```
pombredanne, line 66: 1MB is too small. make it 5 MB

pombredanne (approved):
Looking good!

pombredanne:
we will need some tests later.

scanpipe/pipes/scancode.py
AyanSinhaMahapatra: Done!

tdruez (changes requested):
Instead of hardcoded values, `SKIP_DATA_FILE_SIZE` should be a SCIO setting, similar to `SCANCODEIO_SCAN_FILE_TIMEOUT`, off by default. Also, I don't think the `SKIP_DATA_FILE_EXTENSIONS` is relevant.

I would suggest adding a `SCANCODEIO_SCAN_MAX_FILE_SIZE`, default to None. This is the least intrusive way and will stay useful in the future.

scanpipe/pipes/scancode.py
```diff
@@ -350,14 +366,7 @@ def scan_resources(
...
                 )
-
-                resource.project.add_error(
-                    exception=broken_pool_error,
-                    model="scan_resources",
-                    description=message,
-                    object_instance=resource,
-                )
-                continue
+                raise broken_pool_error from InsufficientResourcesError(message)
```
tdruez, line 369: Why are we reverting to the old behavior?

scanpipe/pipes/scancode.py
```diff
@@ -310,9 +318,17 @@ def scan_resources(
     if not scan_func_kwargs:
         scan_func_kwargs = {}
 
-    resource_count = resource_qs.count()
+    # Skip scannning data files larger than 5 MB
+    if not scan_func == scan_for_package_data:
+        resource_qs.filter(
+            Q(extension__in=SKIP_DATA_FILE_EXTENSIONS)
+            & Q(size__gte=SKIP_DATA_FILE_SIZE)
+        ).update(status=flag.IGNORED_LARGE_DATA_FILE)
```
tdruez, line 326: This should be a pipe in flag.py

scanpipe/pipes/scancode.py
AyanSinhaMahapatra: This was problematic at scale with multiprocessing and creating a lot more errors other than the resource which caused the memory spike, which was not apparent at first from testing smaller codebases.

scanpipe/pipes/scancode.py
AyanSinhaMahapatra: Thanks, I've moved this to flag.py.

AyanSinhaMahapatra:
@tdruez I've added a `SCANCODEIO_SCAN_MAX_FILE_SIZE` as suggested by you since this seemed less problematic, this is ready for review.

scanpipe/pipes/flag.py
```diff
@@ -102,6 +105,20 @@ def flag_ignored_patterns(project, patterns):
...
+    Flag codebase resources which are over the max file size for scanning
+    and return all other files within the file size limit.
+    """
+    if not file_size_limit:
+        return resource_qs
+
+    resource_qs.filter(size__gte=file_size_limit).update(
+        status=IGNORED_BY_MAX_FILE_SIZE
+    )
+    return resource_qs.filter(~Q(status=IGNORED_BY_MAX_FILE_SIZE))
```
tdruez, line 119: This is not consistent with the existing flag method. It should return the update count.
The filtering should be done outside of the flag function

scanpipe/tests/pipes/test_scancode.py
```diff
@@ -445,6 +446,24 @@ def test_scanpipe_pipes_scancode_run_scan_args(self, mock_run_scan):
...
+            pipeline = run.make_pipeline_instance()
+
+            exitcode, out = pipeline.execute()
+            self.assertEqual(0, exitcode, msg=out)
+            self.assertEqual(
+                project1.codebaseresources.get(
+                    path="to-trustier-binary-linux.tar.gz-extract/trustier"
+                ).status,
+                flag.IGNORED_BY_MAX_FILE_SIZE,
+            )
```
tdruez, line 465: Make this more readable.

setup.cfg
```diff
@@ -72,16 +72,16 @@ install_requires =
     # Docker
     container-inspector==33.0.0
     # ScanCode-toolkit
-    scancode-toolkit[packages]==32.3.0
+    scancode-toolkit[packages] @ git+https://github.com/nexB/scancode-toolkit.git@c9c023e7712044475578d03814f34bb03435d949
```
tdruez, line 75: Not great. Could you push a release instead?

tdruez (changes requested):
This looks good, ready to merge once the couple small issues are addressed.

scanpipe/tests/pipes/test_scancode.py
AyanSinhaMahapatra: done, thanks

pombredanne (commented):
IMHO we should have a scancod-config entry for this parameter, otherwise this requires changing the installation settings. This could be a follow up issue.

setup.cfg
AyanSinhaMahapatra: Done! Updated to https://github.com/aboutcode-org/scancode-toolkit/releases/tag/v32.3.2

AyanSinhaMahapatra:
Thanks for the reviews @tdruez and @pombredanne, all addressed and ready now!

tdruez:
@AyanSinhaMahapatra Looks great, thanks!

PR #1529, opened 2025-01-12
Tag files that are whitedouts

### LINKED ISSUE #1528: Tag files that are whitedouts as "white-out" in container scans
We should tag files that are "whited-outs" as "white-out" in container scans so that we can possibly treat as less risky packages that are "whited-outs" as less likely to be used at run time.

VarshaUN:
Fixes #1528 
Used Taggit method to tag white-outs
Signed-off-by : Varsha U N <varshaun58@gmail.com>


VarshaUN:
@pombredanne I have made sure ruff checks are passed but these above checks are not getting passed!! .

PR #1532, opened 2025-01-14
Skip package creation when no package_data from purldb

keshav-space:
- Fixes https://github.com/aboutcode-org/scancode.io/issues/1530


tdruez (changes requested):
@keshav-space Could you add a unit test to ensure that we will not reintroduce the issue in the future?

keshav-space:
> Could you add a unit test to ensure that we will not reintroduce the issue in the future?

@tdruez done!

tdruez (approved):

Issue #1404, opened 2024-10-17
Refine ScanCode.io pipeline for Java


pombredanne:
Update pipelines steps for the binary to source analysis for Java using strings and symbols

pombredanne:
This PR is a step in that direction:
- https://github.com/aboutcode-org/purldb/pull/538

pombredanne:
The first step would be to collect a few example with these:
- kotlin: .kt files: See https://github.com/aboutcode-org/scancode.io/issues/1806
- scala: .scala files: See https://github.com/aboutcode-org/scancode.io/issues/1805
- For JAXB and other less common cases, see: https://github.com/aboutcode-org/scancode.io/issues/1900

TG1999:
java_package_re = re.compile(r"^\s*package\s+([\w\.]+)\s*;?")

Try this for Kotlin, Scala and Java

pombredanne:
There are related issues
- https://github.com/aboutcode-org/scancode.io/issues/1805
- https://github.com/aboutcode-org/scancode.io/issues/1806
- https://github.com/aboutcode-org/scancode.io/issues/1807
- https://github.com/aboutcode-org/scancode.io/issues/1808

TG1999:
This issue is done now

Reference:
- https://github.com/aboutcode-org/scancode.io/pull/1898

Both sub-issues have detailed completion reports

- https://github.com/aboutcode-org/scancode.io/issues/1806 -> https://github.com/aboutcode-org/scancode.io/issues/1806#issuecomment-3390069114
- https://github.com/aboutcode-org/scancode.io/issues/1805 -> https://github.com/aboutcode-org/scancode.io/issues/1805#issuecomment-3390107622

<img width="1811" height="849" alt="Image" src="https://github.com/user-attachments/assets/939734c2-c9fd-4d2e-8388-2a8c69ff447e" />
<img width="1805" height="636" alt="Image" src="https://github.com/user-attachments/assets/40c0995a-5de1-426f-bcfd-4bc857dfed74" />

PR #1550, opened 2025-01-20
Add ecosystem specific inclusions or exclusions

AyanSinhaMahapatra:
Also ignore specific files paths containing metadata in ruby gems.

Reference: https://github.com/aboutcode-org/scancode.io/issues/1438
Reference: https://github.com/aboutcode-org/scancode.io/issues/1476


scanpipe/config.py
```diff
@@ -0,0 +1,115 @@
+# SPDX-License-Identifier: Apache-2.0
```
pombredanne, line 1: IMHO we need a better name as this is not a general purpose configuration. It could be either:
- `ecosystem_config.py`
- or better `d2d_config.py` as this is really focused on d2d for now, and may even need to be brought down one dir under `/pipelines`

pombredanne (changes requested):
Looking fine, the name "config.py" is too generic though for now.

scanpipe/config.py
```diff
@@ -0,0 +1,115 @@
...
+    # Symbols which are found in ecosystem specific standard libraries
+    # which are not so useful in mapping
+    standard_symbols_to_exclude = []
+
+
+class DefaultEcosystemConfig(EcosystemConfig):
+    """Configurations which are common across multiple ecosystems."""
+
+    ecosystem_option = "Default"
+    purldb_package_extensions = [".zip", ".tar.gz", ".tar.xz"]
```
pombredanne, line 63: This looks fairly limited but OK as a default.

scanpipe/config.py
AyanSinhaMahapatra: Done, thanks!

AyanSinhaMahapatra:
@tdruez fixed the tests, this is ready for your review. :smile:

scanpipe/pipelines/deploy_to_develop.py
```diff
@@ -152,6 +128,15 @@ def flag_whitespace_files(self):
...
         d2d.flag_whitespace_files(project=self.project)
 
+    def load_ecosystem_config(self):
+        """Load ecosystem specific configurations for d2d steps for selected options."""
+        d2d.load_ecosystem_config(pipeline=self, options=self.selected_groups)
+
+    @optional_step("Ruby")
+    def load_ecosystem_config_ruby(self):
+        """Load Ruby specific configurations for d2d steps."""
+        pass
```
tdruez, line 138: Is this a leftover or are we planning to duplicate those methods?

scanpipe/pipelines/deploy_to_develop.py
```diff
@@ -29,9 +29,10 @@
 from scanpipe.pipes import matchcode
 from scanpipe.pipes import purldb
 from scanpipe.pipes import scancode
+from scanpipe.pipes.d2d_config import DefaultEcosystemConfig
 
 
-class DeployToDevelop(Pipeline):
+class DeployToDevelop(Pipeline, DefaultEcosystemConfig):
```
tdruez, line 35: I think we should make this work without the need for the extra mixing. See further suggestions.

scanpipe/pipes/d2d.py
```diff
@@ -114,6 +125,55 @@ def get_best_path_matches(to_resource, matches):
...
 
 
+def load_ecosystem_config(pipeline, options):
+    """
+    Add ecosystem specific configurations for each ecosystem selected
+    as `options` to the `pipeline`.
+    """
+    configs_by_ecosystem = {
+        ecosystem.ecosystem_option: ecosystem for ecosystem in ECOSYSTEM_CONFIGS
+    }
```
tdruez, line 135: This could be replaced by a `def get_ecosystem_config(ecosystem)` imported from the d2d_config module. See the implementation suggestion below.

scanpipe/pipes/d2d.py
```diff
@@ -114,6 +125,55 @@ def get_best_path_matches(to_resource, matches):
     return matches
 
 
+def load_ecosystem_config(pipeline, options):
+    """
+    Add ecosystem specific configurations for each ecosystem selected
+    as `options` to the `pipeline`.
```
tdruez, line 131: We need to provide more details about what's actually happening when "adding" a config.

scanpipe/pipes/d2d.py
```diff
@@ -114,6 +125,55 @@ def get_best_path_matches(to_resource, matches):
...
+            continue
+
+        add_ecosystem_config(
+            pipeline=pipeline,
+            configs_by_ecosystem=configs_by_ecosystem,
+            selected_option=selected_option,
+        )
+
+
+def add_ecosystem_config(pipeline, configs_by_ecosystem, selected_option):
```
tdruez, line 156: Missing a detailed docstring.

scanpipe/pipes/d2d.py
```diff
@@ -114,6 +125,55 @@ def get_best_path_matches(to_resource, matches):
...
+    for pipeline_config in d2d_pipeline_configs:
+        config_value = getattr(ecosystem_config, pipeline_config)
+        pipeline_config_value = getattr(pipeline, pipeline_config)
+        if config_value:
+            if not pipeline_config_value:
+                new_config_value = config_value
+            else:
+                new_config_value = pipeline_config_value.extend(config_value)
+
+            setattr(pipeline, pipeline_config, new_config_value)
```
tdruez, line 174: Not ideal to set values from all the way down here, shouldn't we return those to a higher location that will explicitly set the values?

tdruez (changes requested):
See some suggestions about refining how we declare the configs. 
Now, I'm struggling to understand the way we load and then use the config values on the pipeline. The implementation is probably a bit too complex and we should look into ways to simplify this so it's not too hard to maintain in the future.

scanpipe/pipes/d2d.py
```diff
@@ -66,6 +67,16 @@
...
 
 
+ECOSYSTEM_CONFIGS = [
+    d2d_config.DefaultEcosystemConfig,
+    d2d_config.JavaEcosystemConfig,
+    d2d_config.JavaScriptEcosystemConfig,
+    d2d_config.RubyEcosystemConfig,
+    d2d_config.RustEcosystemConfig,
+    d2d_config.GoEcosystemConfig,
+]
```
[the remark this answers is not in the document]
AyanSinhaMahapatra: Yes, this is now in d2d_config and we just use ECOSYSTEM_CONFIGS as a mapping suggested below

scanpipe/pipes/d2d_config.py
```diff
@@ -0,0 +1,115 @@
...
+# Data Generated with ScanCode.io is provided on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, either express or implied. No content created from
+# ScanCode.io should be considered or used as legal advice. Consult an Attorney
+# for any legal advice.
+#
+# ScanCode.io is a free software code scanning tool from nexB Inc. and others.
+# Visit https://github.com/aboutcode-org/scancode.io for support and download.
+
+
+class EcosystemConfig:
```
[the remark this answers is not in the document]
AyanSinhaMahapatra: Thanks, added this as suggested.

scanpipe/pipes/d2d.py
AyanSinhaMahapatra: Added, thanks!

scanpipe/pipelines/deploy_to_develop.py
AyanSinhaMahapatra: I have made this a pipeline attribute instead now. Thanks

scanpipe/pipes/d2d.py
AyanSinhaMahapatra: Added, thanks!

scanpipe/pipes/d2d.py
AyanSinhaMahapatra: Added, thanks!

scanpipe/pipes/d2d.py
AyanSinhaMahapatra: I'm not sure if I can change this, as this is the cleanest way where we can update a bunch of configuration attributes based on all the ecosystems/options selected. If we return these values, there's no use of this function at all, which refactors and gets rid of duplicate code.

scanpipe/pipelines/deploy_to_develop.py
AyanSinhaMahapatra: Ruby is the only option which does not require any code to match source/binary as we already have sha1 matching working there correctly. But the loading of ecosystem configurations and showing/selecting these options in the UI happens on the basis of having optional steps, so I'm having a dummy step here to enable loading the config and showing this as an option everywhere that we can select.

scanpipe/pipes/d2d_config.py
```diff
@@ -0,0 +1,167 @@
...
+            ".pptx",
+            ".tex",
+            ".odt",
+            ".odp",
+        ],
+    ),
+    "Java": EcosystemConfig(
+        ecosystem_option="Java",
+        purldb_package_extensions=[".jar", ".war"],
+        purldb_resource_extensions=[".class"],
```
pombredanne: what about .aj, java, .kt, and so on? I guess they will be handled in:
- https://github.com/aboutcode-org/scancode.io/issues/1404

scanpipe/pipes/d2d_config.py
```diff
@@ -0,0 +1,167 @@
...
+    """
+    Base class for ecosystem-specific configurations to be defined
+    for each ecosystem.
+    """
+
+    # This should be defined for each ecosystem which
+    # are options in the pipelines
+    ecosystem_option: str = "Default"
+
+    # These are extensions for packages of this ecosystem which
```
pombredanne:
```
    # These are extensions for package archive files for this ecosystem
    # that are matchable against the purldb using matchcode
    matchable_package_extensions: list = field(default_factory=list)

    # These are extensions for file of this ecosystem that are
    # matchable against the purldb using matchcode
    matchable_resource_extensions: list = field(default_factory=list)
```
... this may be a better name?

scanpipe/pipes/d2d_config.py
AyanSinhaMahapatra: Yeah, these will be handled in #1404, added when the support is added and tested.

scanpipe/pipes/d2d_config.py
AyanSinhaMahapatra: Added, thanks!

AyanSinhaMahapatra:
@tdruez this is ready for another round of review, I've updated this from your and Philippe's suggestions.
Also updated this to include source symbols related config and tests for a ruby d2d example.

scanpipe/pipelines/deploy_to_develop.py
pombredanne: Ruby should not be special. I get that the mapping is already done with sha1matching, but I think you should not name this "load_ecosystem_config_ruby" as a "tag" step for ruby: this is confusing. I suggest instead to rename this to "map_ruby": it may not do anything just now, but it will likely do things later as explained in:
https://github.com/aboutcode-org/scancode.io/issues/1436#issuecomment-3018881766

therefore the "pass" there will be temporary.

scanpipe/pipes/d2d_config.py
```diff
@@ -0,0 +1,185 @@
...
+    ),
+    "Ruby": EcosystemConfig(
+        ecosystem_option="Ruby",
+        matchable_package_extensions=[".gem"],
+        matchable_resource_extensions=[".rb"],
+        deployed_resource_path_exclusions=["*checksums.yaml.gz*", "*metadata.gz*"],
+    ),
+    "Elf": EcosystemConfig(
+        ecosystem_option="Elf",
+        source_symbol_extensions=[".c", ".cpp", ".h"],
```
pombredanne, line 124: Please file an issue to improve this list which is way too small. Check the common or scancode function to see a list of these.

pombredanne (changes requested):
All good except for the ruby config step that I think should be renamed.

scanpipe/pipes/d2d_config.py
AyanSinhaMahapatra: Issue opened at https://github.com/aboutcode-org/scancode.io/issues/1707

scanpipe/pipes/d2d.py
```diff
@@ -1463,6 +1464,20 @@ def match_resources_with_no_java_source(project, logger=None):
...
+    """Ignore unmapped resources for a project using `patterns_to_ignore`."""
+    ignored_resources_count = flag.flag_ignored_patterns(
+        codebaseresources=project.codebaseresources.to_codebase().no_status(),
+        patterns=patterns_to_ignore,
+        status=flag.IGNORED_FROM_CONFIG,
+    )
+    if logger:
+        logger(
+            f"Ignoring {ignored_resources_count:,d} to/ resources with "
+            "from ecosystem specific configurations."
```
pombredanne:
```suggestion
            "ecosystem specific configurations."
```

PR #1572, opened 2025-02-01
Add ability to export filtered QuerySet of `FilterView` to JSON format

### LINKED ISSUE #1319: Add JSON format in list view export
We only offer XLSX at the moment. I would be useful to offer JSON as well so Package selection could be exported to be imported in another context.

aayushkdev:
fix #1319 

Added the ability to export the current filtered QuerySet of a `FilterView` into the JSON format.

### DIFF
--- a/scanpipe/views.py
+++ b/scanpipe/views.py
@@ -36,6 +36,7 @@ from django.contrib.auth.mixins import LoginRequiredMixin
 from django.core.exceptions import SuspiciousFileOperation
 from django.core.exceptions import ValidationError
 from django.core.files.storage.filesystem import FileSystemStorage
+from django.core.serializers.json import DjangoJSONEncoder
 from django.db.models import Prefetch
 from django.db.models.manager import Manager
 from django.http import FileResponse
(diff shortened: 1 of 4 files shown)


aayushkdev:
Hey, thanks for the review. I have exams going on right now, so I'll add the unit tests as soon as they're finished.

aayushkdev:
Hey @tdruez, I have pushed the required unit tests. Could you please check if they are fit for merging

tdruez (changes requested):
1. The ExportJSONMixin is used for all the object models but only the CodebaseResourceListView returns a valid response.
The code breaks if you try to export from any other list view, such as packages.
The tests are incomplete as those only test for the resource type.

2. The returned JSON is not valid when multiple records are included. It seems that the comma is missing.

aayushkdev:
Hey @tdruez I have pushed the required changes, please have a look.

scanpipe/views.py
```diff
@@ -499,6 +499,73 @@ def get(self, request, *args, **kwargs):
...
+
+        output_file = io.BytesIO()
+
+        first = True
+        for chunk in self.get_filtered_files(queryset):
+            if not first:
+                output_file.write(b",\n")
+            output_file.write(chunk.encode("utf-8"))
+            first = False
+        output_file.seek(0)
```
tdruez, line 534: The exported content is still not proper JSON because the records are not contained in a list.

Also, you re-use some logic from the JSONResultsGenerator, which is use to stream the JSON content as it is generated. In the context of the simple JSON export view, we do not need that complexity.
The Mixin class can be simplified to something along:

```
class ExportJSONMixin:
    """
    Add the ability to export the current filtered QuerySet of a `FilterView`
    into JSON format.
    """

    export_json_query_param = "export_json"

    def get_export_json_queryset(self):
        return self.filterset.qs

    def get_export_json_filename(self):
        return f"{self.project.name}_{self.model._meta.model_name}.json"

    def export_json_file_response(self):
        from scanpipe.api.serializers import get_model_serializer

        queryset = self.get_export_json_queryset()
        serializer_class = get_model_serializer(queryset.model)
        serializer = serializer_class(queryset, many=True)
        serialized_data = json.dumps(serializer.data, indent=2, cls=DjangoJSONEncoder)
        # Convert JSON string into a BytesIO file-like object
        buffer = io.BytesIO(serialized_data.encode('utf-8'))

        return FileResponse(
            buffer,
            as_attachment=True,
            filename=self.get_export_json_filename(),
            content_type="application/json",
        )

    def get_context_data(self, **kwargs):
        context = super().get_context_data(**kwargs)

        query_dict = self.request.GET.copy()
        query_dict[self.export_json_query_param] = True
        context["export_json_url_query"] = query_dict.urlencode()

        return context

    def get(self, request, *args, **kwargs):
        response = super().get(request, *args, **kwargs)

        if request.GET.get(self.export_json_query_param):
            return self.export_json_file_response()

        return response
```

tdruez (changes requested):
@aayushkdev thanks for the changes, we are getting closer but the JSON output is still invalid. See the comment for suggestions.

Also on a more general note: avoid tempering with the git history by merging all previous commits and force-pushing in the branch, this makes looking at the code diffs since the last review much harder.
All the commits will be merged into a single commit once the PR is done, no need to rewrite history during the developement phase.

aayushkdev:
Hey @tdruez I have made the required changes please have a look

tdruez:
@aayushkdev Everything looks good! Could you add a simple entry in the CHANGELOG.rst file so I can merge the PR?

aayushkdev:
> @aayushkdev Everything looks good! Could you add a simple entry in the CHANGELOG.rst file so I can merge the PR?

done!!

tdruez:
> I have also rebased the commits into one

As mentioned earlier, you never need to do this, and should never force-push. It's valuable to keep the commit history.
The merge into `main` is done as a "Squash and merge", which appears as a single commit in the main branch, but we want to keep each single development commit untouched.

tdruez:
@aayushkdev Thanks for your contribution! But keep in mind, if you force-push, you are doing it wrong ;)

aayushkdev:
> > I have also rebased the commits into one
> 
> @aayushkdev As mentioned earlier, you never need to do this, and should never force-push. It's valuable to keep the commit history. The merge into `main` is done as a "Squash and merge", which appears as a single commit in the main branch, but we want to keep each single development commit untouched.

Ah, I see - thanks for clarifying. I misunderstood and thought the final rebase was expected. I’ll make sure not to force-push or squash commits like this again. Appreciate the guidance!

tdruez:
@aayushkdev No worries. Thanks for this super useful feature!

PR #1588, opened 2025-02-20
Add extracted_to and extracted_from methods

AyanSinhaMahapatra:
Adds methods in CodebaseResource objects to navigate from an archive to their corresponding extracted directory and vice versa. This is added to match the functions in commoncode.resource.Resource so these functions work in the scancode.io context similarly to how they work in the scancode-toolkit context.

Reference: https://github.com/aboutcode-org/scancode.io/issues/1585

### DIFF
--- a/scanpipe/models.py
+++ b/scanpipe/models.py
@@ -2889,6 +2889,18 @@ class CodebaseResource(
             if not topdown:
                 yield child
 
+    def extracted_to(self, codebase=None):
+        """Return the path this Resource archive was extracted to or None."""
+        extract_path = f"{self.path}-extract"
+        return self.project.get_resource(extract_path)
+
+    def extracted_from(self, codebase=None):
+        """Return the path to an archive this Resource was extracted from or None."""
+        path = self.path
+        if "-extract" in path:
+            archive_path, _, _ = self.path.rpartition("-extract")
+            return self.project.get_resource(archive_path)
+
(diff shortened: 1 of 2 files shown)


scanpipe/models.py
```diff
@@ -2889,6 +2889,18 @@ def walk(self, topdown=True):
             if not topdown:
                 yield child
 
+    def extracted_to(self, codebase=None):
+        """Return the path this Resource archive was extracted to or None."""
+        extract_path = f"{self.path}-extract"
+        return self.project.get_resource(extract_path)
+
+    def extracted_from(self, codebase=None):
```
tdruez, line 2897: Are the `codebase` args required?

tdruez (changes requested):

scanpipe/models.py
AyanSinhaMahapatra: Yes, this is to have the same function arguments as the functions in commoncode, so any code written in scancode-toolkit (like in package assembly, which is run in both sctk/scio context) works here too. See the functions at commoncode: https://github.com/aboutcode-org/commoncode/blob/main/src/commoncode/resource.py#L1265

See https://github.com/aboutcode-org/scancode.io/blob/add-resource-methods/scanpipe/models.py#L2826 for other functions in the CodebaseResource model which have the same codebase arg for the same reason.

PR #1590, opened 2025-02-21
Reorder XLSX output fields in RESOURCES sheet #1546

### LINKED ISSUE #1546: Restructure order of fields in XLSX output for RESOURCES
When you are reviewing project data in the RESOURCES sheet of the XLSX output there are 2 sets of fileinfo fields that we rarely use during such a review. It would save time reformatting the data to shift these right.

The fields are:
- md5
- sha1
- sha256
- sha512
- 
and

- is_binary
- is_text
- is_archive
- is_media
- is_legal
- is_manifest
- is_readme
- is_top_level
- is_key_file

The request is to move these right to a position between urls and xlsx-errors

aayushkdev:
Fix #1546


tdruez:
@aayushkdev We need a simple unit test to ensure the order does not get changed back later.

aayushkdev:
Hey @tdruez I've added the necessary tests, please check and see if its fit for merging.

tdruez:
@aayushkdev Thanks!

Issue #1601, opened 2025-02-25
`scan_for_application_packages` pipe progress logging is not relevant


tdruez:
The scan_resources is properly using a progress_logger
https://github.com/aboutcode-org/scancode.io/blob/12f33c92fb5574c803bf55caae2b852ea7b44add/scanpipe/pipes/scancode.py#L425C5-L431

```
2025-02-25 04:50:56.948 Step [scan_for_application_packages] starting
2025-02-25 04:51:22.889 Progress: 10% (7795/77950) ETA: 232 seconds (3.9 minutes)
2025-02-25 04:51:26.211 Progress: 20% (15590/77950) ETA: 116 seconds (1.9 minutes)
2025-02-25 04:51:32.919 Progress: 30% (23385/77950) ETA: 84 seconds (1.4 minutes)
2025-02-25 04:51:41.521 Progress: 40% (31180/77950) ETA: 67 seconds (1.1 minutes)
2025-02-25 04:51:51.047 Progress: 50% (38975/77950) ETA: 54 seconds
2025-02-25 04:52:00.779 Progress: 60% (46770/77950) ETA: 42 seconds
2025-02-25 04:52:10.712 Progress: 70% (54565/77950) ETA: 32 seconds
2025-02-25 04:52:20.462 Progress: 80% (62360/77950) ETA: 21 seconds
2025-02-25 04:52:27.001 Progress: 90% (70155/77950) ETA: 10 seconds
2025-02-25 04:52:36.937 Progress: 100% (77950/77950)
```

The issue is that the `assemble_packages` function call is not part of this logging.
It let the user wondering why the progress is stopped at 100% without any feedback on the `assemble` progress, which is the time consuming part.

tdruez:
Logging added in https://github.com/aboutcode-org/scancode.io/pull/1602

PR #1602, opened 2025-02-25
Add proper progress logging for the ``assemble`` function #1601

tdruez:


tdruez:
@AyanSinhaMahapatra could you double-check this? Let me know if you have a better approach in mind ;)

AyanSinhaMahapatra (approved):
@tdruez I don't have anything else to add here, the logging here is working perfectly for now.

Issue #1672, opened 2025-05-16
Bad VCID link to public.vulnerablecode.io


mjherzog:
**Describe the bug**
The VCID links from an SCIO scan are incorrectly constructed.
Example for pkg:maven/org.elasticsearch/elasticsearch@7.17.9
SCIO link to VCIO from the Affected by field in the Vulnerabilities tab is: https://public.vulnerablecode.iovulnerabilities/VCID-1etj-yya3-aaak
It should, of course, be https://public.vulnerablecode.io/vulnerabilities/VCID-1etj-yya3-aaak

**System configuration**
ScanCode.io v3.11.0

**Expected behavior**
SCIO creates a valid link address to VCIO for each VCID reference

JonoYang:
I am able to reproduce this error by downloading and scanning `jackson-databind 2.14.2` and running the `find_vulnerabilities` pipeline on it. When I configure the `VULNERABLECODE_URL` environment variable in `/etc/scancodeio/.env`, I added a trailing `/`  to the URL (`VULNERABLECODE_URL=https://public.vulnerablecode.io/` vs `VULNERABLECODE_URL=https://public.vulnerablecode.io`) and the vulnerablecode URL in the `affected by` column of the vulnerabilities tab in the CodebaseResource view was properly created. When I removed the trailing slash and restarted the server, the URL is incorrect as above. 

The interim fix would be to add a trailing `/` to the `VULNERABLECODE_URL` environment variable. The proper fix is to update the code that creates these URLs to properly handle whether or not the `VULNERABLECODE_URL` envvar has a trailing slash or not.

JonoYang:
The issue is at https://github.com/aboutcode-org/scancode.io/blob/main/scanpipe/templates/scanpipe/tabset/tab_vulnerabilities.html#L14 where there is no `/` after `VULNERABLECODE_URL`, leading to a malformed url if `VULNERABLECODE_URL` does not have a trailing `/`.

settings.py has been updated to strip the rightmost `/` from `PURLDB_URL` , `MATCHCODEIO_URL`,  `VULNERABLECODE_URL`, and `FEDERATEDCODE_GIT_ACCOUNT_URL` when importing the variables from .env

PR #1673, opened 2025-05-20
Strip rightmost / in settings.py, from URL envvars #1672

JonoYang:
Update tab_vulnerabilities.html to add a slash after VULNERABLECODE_URL to properly create the link

Fix for:
- https://github.com/aboutcode-org/scancode.io/issues/1672


pombredanne (approved):
Looking good. Thanks! Do we need a small test?

PR #1674, opened 2025-05-21
Add d2d symbols matching for winpe macho binaries

AyanSinhaMahapatra:
Reference: https://github.com/aboutcode-org/scancode.io/issues/1431
Reference: https://github.com/aboutcode-org/scancode.io/issues/1432
Reference: https://github.com/aboutcode-org/scancode.io/issues/1433


AyanSinhaMahapatra:
This is missing a release of `source-inspector` as this [failed to upload to pypi](https://github.com/aboutcode-org/source-inspector/actions/runs/14412710977/job/40423983686#step:3:151) because we are using an unreleased dependency, see https://github.com/aboutcode-org/source-inspector/blob/v0.6.0/setup.cfg#L60

Attempting to have a release/build a local wheel so we can release properly here.
After merging this we can run the following:
- https://github.com/aboutcode-org/scancode.io/issues/1470
- https://github.com/aboutcode-org/scancode.io/issues/1471
- https://github.com/aboutcode-org/scancode.io/issues/1472

PR #1691, opened 2025-06-13
Add support for tracking parent of CodebaseResource entries and ensure top level paths are stored

### LINKED ISSUE #1687: Add support for tracking parent of CodebaseResource entries and ensure top level paths are stored

aayushkdev:
fix #1687 

**Summary/Goals of this change**
This PR introduces an `parent_directory_path` field to the `CodebaseResource` model to optimize fetching children of a directory. The existing `.children()` method which was intended to fetch children uses regex filtering on path, which is slow for large codebases.
Additionally, this PR also ensures that top-level paths are stored during the resource creation step which is essential for rendering the root level nodes in the file tree

**Breaking Changes**
The file tree will rely on the newly introduced ancestor field and the presence of top-level paths. As a result, existing projects scanned before this PR will need to be re-scanned to work with the file tree view.

**Open Questions**
 Should we deprecate or remove the following methods now that ancestor handles relationship resolution more efficiently?
- parent_path()
- has_parent()
- parent()

Would love feedback on this!

### DIFF
--- a/scanpipe/pipes/scancode.py
+++ b/scanpipe/pipes/scancode.py
@@ -900,6 +900,84 @@ def get_virtual_codebase(project, input_location):
     return VirtualCodebase(input_location, temp_dir=str(temp_path), max_in_memory=0)
 
 
+def create_codebase_resource(project, scanned_resource):
+    """Create a CodebaseResource entry from ScanCode scanned data."""
+    resource_data = {}
+
+    for field in CodebaseResource._meta.fields:
+        # Do not include the path as provided by the scanned_resource since it
+        # includes the "root". The `get_path` method is used instead.
+        if field.name in ["path", "parent_path"]:
+            continue
+        value = getattr(scanned_resource, field.name, None)
+        if value is not None:
+            resource_data[field.name] = value
+
+    resource_type = "FILE" if scanned_resource.is_file else "DIRECTORY"
+    resource_data["type"] = CodebaseResource.Type[resource_type]
+    resource_path = scanned_resource.get_path(strip_root=True)
+
+    parent_path = str(Path(resource_path).parent)
+    if parent_path == ".":
+        parent_path = ""
+    resource_data["parent_path"] = parent_path
+
+    codebase_resource, _ = CodebaseResource.objects.get_or_create(
+        project=project,
+        path=resource_path,
+        defaults=resource_data,
+    )
+
+    # Handle package assignments
+    for_packages = getattr(scanned_resource, "for_packages", [])
+    for package_uid in for_packages:
+        logger.debug(f"Assign {package_uid} to {codebase_resource}")
+        package = project.discoveredpackages.get(package_uid=package_uid)
+        set_codebase_resource_for_package(
+            codebase_resource=codebase_resource,
+            discovered_package=package,
+        )
+
+    # Handle license detections
+    license_detections = getattr(scanned_resource, "license_detections", [])
+    for detection_data in license_detections:
+        detection_identifier = detection_data.get("identifier")
+        pipes.update_or_create_license_detection(
+            project=project,
+            detection_data=detection_data,
+            resource_path=resource_path,
+            count_detection=False,
+        )
+        logger.debug(f"Add {codebase_resource} to {detection_identifier}")
+
+    # Handle license clues
+    license_clues = getattr(scanned_resource, "license_clues", [])
+    for clue_data in license_clues:
+        pipes.update_or_create_license_detection(
+            project=project,
+            detection_data=clue_data,
+            resource_path=resource_path,
+            is_license_clue=True,
+        )
+        logger.debug(f"Add license clue at {codebase_resource}")
+
+    # Handle package data
+    packages = getattr(scanned_resource, "package_data", [])
+    for package_data in packages:
+        license_detections = package_data.get("license_detections", [])
+        license_detections.extend(package_data.get("other_license_detections", []))
+        for detection_data in license_detections:
+            detection_identifier = detection_data.get("identifier")
+            pipes.update_or_create_license_detection(
+                project=project,
+                detection_data=detection_data,
+                resource_path=resource_path,
+                count_detection=False,
+                from_package=True,
+            )
+            logger.debug(f"Add {codebase_resource} to {detection_identifier}")
+
+
 def create_codebase_resources(project, scanned_codebase):
     """
     Save the resources of a ScanCode `scanned_codebase` scancode.resource.Codebase
@@ -909,78 +987,7 @@ def create_codebase_resources(project, scanned_codebase):
     skipped.
     """
     for scanned_resource in scanned_codebase.walk(skip_root=True):
-        resource_data = {}
-
-        for field in CodebaseResource._meta.fields:
-            # Do not include the path as provided by the scanned_resource since it
-            # includes the "root". The `get_path` method is used instead.
-            if field.name == "path":
-                continue
-            if field.name == "parent_path":
-                continue
-            value = getattr(scanned_resource, field.name, None)
-            if value is not None:
-                resource_data[field.name] = value
-
-        resource_type = "FILE" if scanned_resource.is_file else "DIRECTORY"
-        resource_data["type"] = CodebaseResource.Type[resource_type]
-        resource_path = scanned_resource.get_path(strip_root=True)
-
-        parent_path = str(Path(resource_path).parent)
-        if parent_path == ".":
-            parent_path = ""
-        resource_data["parent_path"] = parent_path
-
-        codebase_resource, _ = CodebaseResource.objects.get_or_create(
-            project=project,
-            path=resource_path,
-            defaults=resource_data,
-        )
-
-        for_packages = getattr(scanned_resource, "for_packages", [])
-        for package_uid in for_packages:
-            logger.debug(f"Assign {package_uid} to {codebase_resource}")
-            package = project.discoveredpackages.get(package_uid=package_uid)
-            set_codebase_resource_for_package(
-                codebase_resource=codebase_resource,
-                discovered_package=package,
-            )
-
-        license_detections = getattr(scanned_resource, "license_detections", [])
-        for detection_data in license_detections:
-            detection_identifier = detection_data.get("identifier")
-            pipes.update_or_create_license_detection(
-                project=project,
-                detection_data=detection_data,
-                resource_path=resource_path,
-                count_detection=False,
-            )
-            logger.debug(f"Add {codebase_resource} to {detection_identifier}")
-
-        license_clues = getattr(scanned_resource, "license_clues", [])
-        for clue_data in license_clues:
-            pipes.update_or_create_license_detection(
-                project=project,
-                detection_data=clue_data,
-                resource_path=resource_path,
-                is_license_clue=True,
-            )
-            logger.debug(f"Add license clue at {codebase_resource}")
-
-        packages = getattr(scanned_resource, "package_data", [])
-        for package_data in packages:
-            license_detections = package_data.get("license_detections", [])
-            license_detections.extend(package_data.get("other_license_detections", []))
-            for detection_data in license_detections:
-                detection_identifier = detection_data.get("identifier")
-                pipes.update_or_create_license_detection(
-                    project=project,
-                    detection_data=detection_data,
-                    resource_path=resource_path,
-                    count_detection=False,
-                    from_package=True,
-                )
-                logger.debug(f"Add {codebase_resource} to {detection_identifier}")
+        create_codebase_resource(project, scanned_resource)
 
 
 def create_discovered_packages(project, scanned_codebase):


scanpipe/models.py
```diff
@@ -2685,6 +2685,17 @@ class CodebaseResource(
             'Eg.: "/usr/bin/bash" for a path of "tarball-extract/rootfs/usr/bin/bash"'
         ),
     )
+
+    ancestor = models.CharField(
```
tdruez: `ancestor` seems to generic as a name, it gives the feeling that this may be a FK to another resource.
I think `parent_path` would be much more explicit.

scanpipe/pipes/__init__.py
```diff
@@ -91,6 +92,7 @@ def make_codebase_resource(project, location, save=True, **extra_fields):
     codebase_resource = CodebaseResource(
         project=project,
         path=relative_path,
+        ancestor=parent_path,
```
tdruez: We may want to handle this directly in the save method of the CodebaseResource to ensure the ancestor/parent_path is always set. Some part of the code may not rely on make_codebase_resource to create resource instances.

scanpipe/models.py
aayushkdev: Yeah, I was considering naming it something like parent_path, but there's already a property with that name, which causes a conflict.
How about parent_resource instead?

tdruez: > How about parent_resource instead?

Not great, as the content is a path, not a FK to a resource.
What about `parent_directory_path` for now, we can remove the property and rename later on.

aayushkdev: yeah sure that sounds good I will rename it to `parent_directory_path`

aayushkdev:
### Use of `parent_directory_path` in the Rendering Context

- Fetching children of a directory is now a simple filter on the `parent_directory_path` field, which is indexed and fast. where previously the `.children()` was used method that used regex based filtering which was slow and inefficient

- Adding the `parent_directory_path` field also makes determining whether a directory has children easier as this can now be handled via a single annotated query that counts children per resource. This eliminates the need for per-directory .exists() checks, which gets very expensive with hundreds or thousands of folders.

 - Top-level resources (those with parent_directory_path=None) can be queried directly and efficiently, making the initial rendering of the tree straightforward and performant.
 
 ### Step by step flow
```
README.md
src/
  └── main.py
  └── utils/
       └── helpers.py
data/
  └── input.txt
bin/
```

**1. Initial Load:**

- The codebase fetches the top-level directories/files using:
```
CodebaseResource.objects.filter(project=project, parent_directory_path=None)
```
- This fetches: `README.md`, `src/`, `data/`, and `bin/`, as they have their `parent_directory_path` field set to None.

**2. Determining Expandability**

- Another query will be fired using `parent_directory_path=None` for fetching and annotating an extra field `has_children`.
- This will indicate whether a directory can be expanded in the UI.
- In this case:
- `bin/` and `README.md` have no children => `has_children` = False
- `src/` and `data/` have children => `has_children` = True

**3. Expanding a Folder (e.g., src/):**
- Clicking on the src/ folder sends a request with:
```
CodebaseResource.objects.filter(project=project, parent_directory_path="src")
```
- This fetches its children: `main.py `and `utils/`, as their parent is `src/` .

**4. Determining Children of Fetched Items:**
- Like before another query is fired to check if these items have children.
- In this case:
- `main.py` => `has_children` = False
- `utils/` => `has_children` = True

scanpipe/pipes/__init__.py
aayushkdev: The `collect_and_create_codebase_resources` pipeline internally uses `bulk_create`, which bypasses the model’s save() method. so I would have to keep the current logic in the `make_codebase_resource` when creating the save() method.

tdruez: Ok, can you make sure we have a unit tests that make sure the parent_path is set in those 2 cases?

tdruez:
@aayushkdev Thanks for all the implementation details!

I've renamed the old `parent_path` function to `parent_directory` in https://github.com/aboutcode-org/scancode.io/pull/1694 so it's not in the way of the new field.
Could you merge those into your branch and define `parent_directory_path` as `parent_path` before the merge?

aayushkdev:
Hey @tdruez I have merged your changes and renamed the required fields.

scanpipe/models.py
```diff
@@ -2685,6 +2685,17 @@ class CodebaseResource(
...
         ),
     )
+
+    parent_path = models.CharField(
+        max_length=2000,
+        null=True,
+        blank=True,
+        help_text=_(
+            "Path of the immediate parent directory of a resource. "
+            "For top level resources the value is set to None"
```
tdruez:
```suggestion
            "The path of the resource's parent directory. "
            "Set to None for top-level (root) resources. "
            "Used to efficiently retrieve a directory's contents."
```

scanpipe/models.py
```diff
@@ -2784,6 +2796,11 @@ class Meta:
     def __str__(self):
         return self.path
 
+    def save(self, *args, **kwargs):
+        if self.path and not self.parent_path:
+            self.parent_path = parent_directory(str(self.path), with_trail=False)
```
tdruez: Let's use a single common way to generate the parent_path value reusing
`self.parent_directory()` and preventing code duplication.

```suggestion
            self.parent_path = self.parent_directory()
```

scanpipe/pipes/rootfs.py
```diff
@@ -139,6 +139,13 @@ def get_res(parent, fname):
             rootfs_path=rootfs_path,
         )
 
+    if with_dir:
```
tdruez, line 143: A simple comment could be useful here since it's not obvious when looking at the code.

tdruez: Also, how is this change relates to the new field implementation?

scanpipe/pipes/rootfs.py
```diff
@@ -139,6 +139,13 @@ def get_res(parent, fname):
...
+        rootfs_path = pipes.normalize_path("")
```
tdruez: Why not using directly `rootfs_path = "/"` ?

scanpipe/tests/test_models.py
```diff
@@ -1593,6 +1593,16 @@ def test_scanpipe_codebase_resource_model_compliance_alert_update_fields(self):
         # Reset the index value
         scanpipe_app.license_policies_index = None
 
+    def test_scanpipe_codebase_root_parent_path(self):
+        resource1 = self.project1.codebaseresources.create(path="file")
+
+        self.assertEqual("", resource1.parent_path)
```
tdruez, line 1652: Shouldn't this be `None` instead of empty string?

tdruez: The point is that we want to avoid allowing a mixture of `""` and `None` as it will make the QuerySet more complex to always deal with both possibility.

scanpipe/tests/test_pipelines.py
```diff
@@ -860,6 +860,60 @@ def test_scanpipe_scan_codebase_pipeline_integration(self):
...
+        run = project1.add_pipeline(pipeline_name)
+        pipeline = run.make_pipeline_instance()
+
+        exitcode, out = pipeline.execute()
+        self.assertEqual(0, exitcode, msg=out)
+
+        expected_top_level_paths = ["is-npm-1.0.0.tgz", "is-npm-1.0.0.tgz-extract"]
+
+        top_level_resources = project1.codebaseresources.filter(parent_path=None)
+        top_level_paths = [res.path for res in top_level_resources]
```
tdruez: When possible, use full variable name for better readability and consistency with the codebase.

tdruez (changes requested):
@aayushkdev Looking good, see my suggestion for improvements.

scanpipe/pipes/rootfs.py
aayushkdev: This change makes it so we add root directory as a resource so we can later query it with parent_path for top level resources.
The `with_dir` argument controls whether only files or both files and directories are returned as resources. It's used later in the `os.walk()` loop to include directories, which is why I added it here for consistency.

scanpipe/models.py
aayushkdev: yeah sure I will add it but this will require modifying the parent_directory() method so it converts path to string before calling the parent_directory function as it expects path in str.

scanpipe/tests/test_models.py
aayushkdev: parent_directory() previously returned `""` if a path doesnt have a parent_directory I have modified  it to return `None` if a path doesnt have a parent_directory

scanpipe/pipes/rootfs.py
tdruez: Ok, let's make sure we have a test that execute the new code.

scanpipe/models.py
```diff
@@ -2868,7 +2869,8 @@ def get_path_segments_with_subpath(self):
 
     def parent_directory(self):
         """Return the parent path for this CodebaseResource or None."""
-        return parent_directory(self.path, with_trail=False)
+        parent_path = parent_directory(str(self.path), with_trail=False)
+        return None if parent_path == "" else parent_path
```
tdruez: `return parent_path or None` is more readable.

scanpipe/pipes/rootfs.py
aayushkdev: yes the test `test_scanpipe_scan_codebase_creates_parent_path_field` in test_pipelines executes the new code

tdruez:
> Breaking Changes
> The file tree will rely on the newly introduced ancestor field and the presence of top-level paths. As a result, existing projects scanned before this PR will need to be re-scanned to work with the file tree view.

A data migration could solve this. The question is how fast could we migrate the data on existing databases that include millions of CodebaseResource records.

aayushkdev:
Yes, iterating through all the resources in CodebaseResource to add the parent_path value to each resource would be quite slow, especially with multiple projects. If you'd like, I can prepare a migration script to test this out and see how it affects performance.

tdruez:
I don't think a "script" is the right approach here. We need to look into a raw SQL approach for acceptable performances.

aayushkdev:
Yeah, I think this will be doable. I’ll explore this further and test it out.

aayushkdev:
Hey @tdruez, I’ve got the migration for setting `parent_path` working. However, I could use your input on the migration for inserting the missing root directories. I initially assumed I could just insert the `path`, `project_id`, and `parent_path` fields only which would be doable through sql, but it turns out there are other non-nullable fields like sha1 and md5 that typically require a scan to populate which isnt possible with pure sql. One workaround could be to use dummy values for these fields. What do you think?

tdruez:
>  I initially assumed I could just insert the path, project_id, and parent_path fields only which would be doable through sql
 
Why would you have to insert those field? We are talking about generating a value for `parent_path` on existing CodebaseResource. Those existing records already have a `path` and a `project_id`.

> I’ve got the migration for setting parent_path working.

Can you paste your progress on this migration in a comment, it'll be easier to provide you feedback.

aayushkdev:
> Why would you have to insert those field? We are talking about generating a value for `parent_path` on existing CodebaseResource. Those existing records already have a `path` and a `project_id`.

This migration has two parts first, adding the parent_path field to each resource, which I’ve already completed and second, inserting new resources with top-level root paths into CodebaseResource for each project. This is necessary because pipelines like `map_deploy_to_develop` don’t store root-level directories like `from` and `to` in the CodebaseResource, so we need to add that with the migration.



> Can you paste your progress on this migration in a comment, it'll be easier to provide you feedback.

```python
from django.db import migrations

class Migration(migrations.Migration):

    dependencies = [
        ('scanpipe', '0073_codebaseresource_parent_path'),
    ]

    operations = [
        migrations.RunSQL(
            """
            UPDATE scanpipe_codebaseresource
            SET parent_path = regexp_replace(path, '/[^/]+$', '')
            WHERE path LIKE '%/%';
            """
        ),
    ]

```

aayushkdev:
Hey @tdruez  @pombredanne , I explored different ways to store trees (like MPTT, Treebeard and ltree), but they all involve more complex integration to scancode.io, introduce even more additional fields, or require us to change the way Resources are created which slows down creation of resources,  all while offering a minimal performance benefit over the `parent_path` field approach. Introducing the `parent_path` field provides a more lightweight and efficient solution that integrates cleanly with our existing path logic.

 I also tested the `parent_path` field approach on deeply nested trees, and performance remains strong even with thousands of resources.
 
So I don't think its necessary to change our approach on how we handle the storage of trees.

scanpipe/models.py
```diff
@@ -2695,6 +2695,17 @@ class CodebaseResource(
             'Eg.: "/usr/bin/bash" for a path of "tarball-extract/rootfs/usr/bin/bash"'
         ),
     )
+
+    parent_path = models.CharField(
+        max_length=2000,
+        blank=True,
+        help_text=_(
+            "The path of the resource's parent directory. "
+            "Set to None for top-level (root) resources. "
```
tdruez: This is not true, as the field is non-nullable.

tdruez:
@aayushkdev There's an issue when running the `scan_single_package`, as the make_codebase_resource logic is not called, as the data is loaded from the scancode-toolkit results.
We end up with the `codebase` value for root resource `parent_path`, and the `codebase/` prefix for all other resource. Which then breaks the tree UI.

The `scanpipe.pipes.scancode.create_codebase_resources` needs to be adapted to provide the proper value for parent_path. Also make sure to add a unit test to cover this case.

aayushkdev:
@tdruez I have done the requested changes

tdruez:
@aayushkdev Make sure to fix the failing tests first ;)

aayushkdev:
hey @tdruez there is a code formatting issue in scancode.py where the `create_codebase_resources` function is too complex and I am not really sure how to handle this should I split the function into helper functions?

tdruez:
>  the create_codebase_resources function is too complex and I am not really sure how to handle this should I split the function into helper functions?

Refactoring the complexity is a possibility, but in this case the code can be simplified. For example:
```
if field.name == "path":
    continue
if field.name == "parent_path":
    continue
```

Can be simplified as
```
if field.name in ["path", "parent_path"]:
    continue
```

Also, the main loop could be refactored as:

```
def create_codebase_resource(project, scanned_resource):
    """Create a CodebaseResource entry from ScanCode scanned data."""
    resource_data = {}
    [...]


def create_codebase_resources(project, scanned_codebase):
    """
    Save the resources of a ScanCode `scanned_codebase` scancode.resource.Codebase
    object to the database as a CodebaseResource of the `project`.
    This function can be used to expend an existing `project` Codebase with new
    CodebaseResource objects as the existing objects (based on the `path`) will be
    skipped.
    """
    for scanned_resource in scanned_codebase.walk(skip_root=True):
        create_codebase_resource(project, scanned_resource)
```

PR #1708, opened 2025-06-30
Update scancode-toolkit to v32.4.0

AyanSinhaMahapatra:


scanpipe/pipes/scancode.py
```diff
@@ -767,7 +767,7 @@ def run_scan(location, output_file, run_scan_args):
     """Scan the `location` content and write the results into an `output_file`."""
     _success, results = scancode_run_scan(
         input=shlex.quote(location),
-        processes=get_max_workers(keep_available=1),
+        processes=get_max_workers(keep_available=2),
```
tdruez: What's the context behind this change?

scanpipe/models.py
```diff
@@ -2748,6 +2748,12 @@ class Type(models.TextChoices):
     is_text = models.BooleanField(default=False)
     is_archive = models.BooleanField(default=False)
     is_media = models.BooleanField(default=False)
+    sha1_git = models.CharField(
+        _("SHA1_git"),
+        max_length=40,
+        blank=True,
+        help_text=_("SHA1 git checksum hex-encoded, as in git genreated checksum."),
```
tdruez: Typo in "genreated".
Also, help could be improved: "git checksum ... as in git generated checksum"

scanpipe/models.py
```diff
@@ -2748,6 +2748,12 @@ class Type(models.TextChoices):
     is_text = models.BooleanField(default=False)
     is_archive = models.BooleanField(default=False)
     is_media = models.BooleanField(default=False)
+    sha1_git = models.CharField(
```
tdruez, line 2751: How do we plan to leverage this value on the ScanCode.io side?
I'm wondering if it requires an index.

tdruez:
Missing a changelog entry.

scanpipe/pipes/scancode.py
AyanSinhaMahapatra: Reverting this, committed by mistake. My session was crashing while applying a test fixtures regen so I was trying to get around that.

scanpipe/models.py
AyanSinhaMahapatra: Done, thanks!

scanpipe/models.py
AyanSinhaMahapatra: I don't think we are planning to leverage this atm, all the matching/other checksum related features are already handled by other specific checksums and so we probably don't need an index right away. We can always add one if required, afterwards. Note also that I've not added this to `HashFieldsMixin` because this `HashFieldsMixin` isinherited by both `CodebaseResource` and `DiscoveredPackage`, and we don't have the `git_sha1` checksum for packages in the SCTK side as this was not required. See https://github.com/aboutcode-org/scancode-toolkit/issues/624 also for the original issue.

AyanSinhaMahapatra:
@tdruez thanks for the review comments, I've addressed them and this is ready for review again

scanpipe/models.py
tdruez: > "I don't think we are planning to leverage this atm"

Is it useful to store it then? What is the usecase for this value?

AyanSinhaMahapatra: From the issue in scancode-toolkit you can see this was a requested feature, and this would also be useful in the mass-scan project (software heritage uses this type of checksum too) which is why @pombredanne asked me to include it in scancode-toolkit AFAIK. When we use a updated version of scancode-tookit, the schema is heavily coupled so we need to migrate and use the new attributes there (Or most of the tests will start failing, which is why I added it), and this could be quite useful for anyone who is using the `git_sha1`  checksum in any way downstream, since this is included in the results and available in general in the UI/API. I don't think there is a question of whether this is useful or not in the SCIO/SCTK context. But I'm not sure whether we have some use where we need a specific index for this field as we are not doing any queries with `git_sha1`, and AFAIK indexes of other checksum fields are there because those are used to query resources/packages? Could you explain why we index other checksum fields? Maybe then I can understand better.

tdruez: > Could you explain why we index other checksum fields? Maybe then I can understand better.

Usually used for matching a file or a package archive. I was wondering about the contexts where one would use a Git SHA1 over a regular SHA1.

PR #1721, opened 2025-07-08
add API end-points for indexing

VarshaUN:
I built an API to query and download stored package data by implementing a viewset that handles listing, retrieving, and downloading packages by ID, URL, or SHA256 checksum, with filtering capabilities using a dependency for query refinement and read-only access for unauthenticated users. I set up API routing with a dedicated router to manage endpoints under a distinct /api/ path, cleanly separating them from web routes. This enables efficient package reuse, reduces redundant downloads, and enhances ScanCode.io’s functionality with programmatic access to package archives, pending final testing to confirm seamless integration.


VarshaUN:
Hey @AyanSinhaMahapatra , CI tests have failed since I am trying to import `DownloadedPackage`  model that isn't in this branch , it is present in my last PR . How should I approach this??

PR #1722, opened 2025-07-08
Add a QuickStart introduction chapter in the documentation

tdruez:
Also add a `run-scan.sh` bash script to wrap the run command


DennisClark:
@tdruez   the new quickstart is very impressive!

PR #1810, opened 2025-08-12
Improve Go/Rust/Elf/Mach0 binary symbol mapping

AyanSinhaMahapatra:
Reference: https://github.com/aboutcode-org/scancode.io/issues/1707
Reference: https://github.com/aboutcode-org/scancode.io/issues/1475


tdruez (changes requested):
Lack of unit test for the new functions.

AyanSinhaMahapatra:
@tdruez thanks! we have better test coverage now

PR #1828, opened 2025-08-28
Add filter and search support to the codebase tree

aayushkdev:
This pull request introduces searching and filtering functionality. It uses htmx to perform in-place updates, so filtered results show instantly without any page reloads.


tdruez:
@aayushkdev I think improving the navigation first, see https://github.com/aboutcode-org/scancode.io/issues/1840, would be more useful than duplicating the "searching and filtering functionalities" of the existing Resource list table.

aayushkdev:
> @aayushkdev I think improving the navigation first, see #1840, would be more useful than duplicating the "searching and filtering functionalities" of the existing Resource list table.

Sure I will get started working on those issues first

Issue #1840, opened 2025-09-03
Resources tree view additional features


tdruez:
## Navigation

- [x] 1. Display the **current path** above the right pane table. PR: https://github.com/aboutcode-org/scancode.io/pull/1851
  - Follow the style of the resource details view header  
  - Add a "copy path" icon action  (also available in the resource details view header  )

- [x] 2. Track the **current browsing path** in the browser URL. PR: https://github.com/aboutcode-org/scancode.io/pull/1853
  - Update the URL when navigating nodes in the left pane  
  - Ensure page refresh restores the same context  
  - Implement with **htmx**  (see the built-in functions)

- [x] 3. Add a **".." (go to parent directory)** entry as the first row in the right-side table  https://github.com/aboutcode-org/scancode.io/pull/1869

- [x] 4. Load **ResourceDetailsView** in the right pane when clicking a file https://github.com/aboutcode-org/scancode.io/pull/1919
  - Allow rendering the view without the global header context 

- [x] 5. Add a "Go to file" type search, similar to this GitHub view: https://github.com/aboutcode-org/scancode.io/tree/main/scancodeio https://github.com/aboutcode-org/scancode.io/pull/1915
  - Add the search input above the tree nodes.
  - Let's start simple, return the first 30 results, sorted by path, by doing a basic database lookup in the Resources path of the project. (We can add regex support in a future enhancement.)
  - This requires a new view that returns the list of results and fetch it using HTMX.
  - Click on a result should behave the same as a click on one of the tree nodes.

----

Notes: Let's do 1 PR per item.

PR #1851, opened 2025-09-05
add path breadcrumbs and copy-to-clipboard button in resource_table_view

aayushkdev:
Reference #1840 

- Added a path display at the top of resource_table_view, which also serves as breadcrumbs for navigating to parent directories.
- Added a Copy to Clipboard button that copies the current path.

image

### DIFF
--- a/scanpipe/templates/scanpipe/panels/resource_table_panel.html
+++ b/scanpipe/templates/scanpipe/panels/resource_table_panel.html
@@ -1,4 +1,23 @@
 {% load humanize %}
+<div class="mb-4">
+  <div class="has-text-weight-semibold is-flex is-align-items-center is-family-monospace">
+    {% if path_segments %}
+      <span id="resource-path" class="has-text-weight-medium">
+        {% for subpath, segment in path_segments %}
+          {% if not forloop.first %}<span class="mx-1">/</span>{% endif %}
+          {% if not forloop.last %}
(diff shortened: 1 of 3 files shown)


tdruez:
@aayushkdev The copy works fine but the "Path copied" label does not display properly, it's hidden behind the top element:
Screenshot 2025-09-05 at 18 09 53

Also, there's too much white space around the "/" separator.

aayushkdev:
> @aayushkdev The copy works fine but the "Path copied" label does not display properly, it's hidden behind the top element: Screenshot 2025-09-05 at 18 09 53
> 
> Also, there's too much white space around the "/" separator.


image

I have fixed it

tdruez:
@aayushkdev Works great! The next improvement should be to store the current state (current path) in the browser URL.
Also, we need to be able to load the state from thw URL, see https://github.com/aboutcode-org/scancode.io/blob/0067725c1ccaa085e2da7b08edfcfa6b2679423e/scanpipe/urls.py#L39-L43 for providing the URL path to the view.

aayushkdev:
> @aayushkdev Works great! The next improvement should be to store the current state (current path) in the browser URL. Also, we need to be able to load the state from thw URL, see
> 
> https://github.com/aboutcode-org/scancode.io/blob/0067725c1ccaa085e2da7b08edfcfa6b2679423e/scanpipe/urls.py#L39-L43
> for providing the URL path to the view.

Hey @tdruez I worked on this improvement over the weekend but my implementation was a bit different I used the url params which were already being passed to the template to load the state. Would be great if you could have a look at that pr #1853

PR #1853, opened 2025-09-06
Track the current browsing path in the browser URL

aayushkdev:
- Refreshing now keeps you on the same folder/files
- Moved redundant code into functions to reduce code duplication in resource_tree

### DIFF
--- a/scanpipe/templates/scanpipe/panels/codebase_tree_panel.html
+++ b/scanpipe/templates/scanpipe/panels/codebase_tree_panel.html
@@ -6,7 +6,7 @@
           <span class="icon is-small chevron mr-1{% if not node.has_children %} is-invisible{% endif %} is-clickable is-flex is-align-items-center" data-chevron>
             <i class="fas fa-chevron-right"></i>
           </span>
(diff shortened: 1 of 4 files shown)


tdruez:
@aayushkdev I generally work fine but there's an issue:
I cannot open a link of the current path, neither with the links of the table, and open into a new tab/new window. Same issue if I want to copy one of those links.

I think the `href="#"` should actually contain the link.

On a minor note, any reason to use "is-family-monospace"? I think the UI looks better without this class.

aayushkdev:
> @aayushkdev I generally work fine but there's an issue: I cannot open a link of the current path, neither with the links of the table, and open into a new tab/new window. Same issue if I want to copy one of those links.
> 
> I think the `href="#"` should actually contain the link.
> 
> On a minor note, any reason to use "is-family-monospace"? I think the UI looks better without this class.

Using is-family-monospace was more of a personal preference also I also wanted the breadcrumbs to look distinct from the table, but you’re right, it’s better if they use the usual font, so I’ve removed it.

tdruez:
@aayushkdev It looks and works great! 🚀 

Since the next item on the list is quite small, it would be a good time to document a bit more each JS function to make the code easier to read and wok on.

PR #1888, opened 2025-09-26
Avoid creating empty commits when pushing changes to FederatedCode

keshav-space:
Related: https://github.com/aboutcode-org/purldb/issues/667 and https://github.com/aboutcode-org/purldb/pull/721

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -121,7 +121,7 @@ android_analysis = [
   "android_inspector==0.0.1"
 ]
 mining = [
-  "minecode_pipelines==0.0.1b1"
+  "minecode_pipelines==0.0.1b8"
 ]
 
 [project.urls]


scanpipe/pipes/federatedcode.py
```diff
@@ -239,7 +242,11 @@ def commit_changes(
         """
 
     repo.index.add(files_to_commit)
-    repo.index.commit(textwrap.dedent(commit_message))
+    repo.git.commit(
```
JonoYang, line 245: @keshav-space just wondering what the difference between `repo.index.commit` and `repo.git.commit`

keshav-space: @JonoYang `repo.index` is python abstraction of git while `repo.git` runs the native command line git. We switched from `repo.index` to `repo.git` for committing our changes because `repo.index` by default allows empty commits and there is no way to change this behavior. So we ended up using the native `repo.git` to disallow empty commits.

pombredanne: This means that git is now a system requirement that needs to be installed too. Please double check the docs.

keshav-space: @pombredanne we already have git as system requirement https://github.com/aboutcode-org/scancode.io/blob/5e4aff91fd00ede494d5a54067f973654fa116c0/Dockerfile#L66

PR #1890, opened 2025-09-30
Update federatedcode.clone_repository

ziadhany:
Update clone_repository to have an option to clone the entire Git repository history.

### DIFF
--- a/scanpipe/pipes/federatedcode.py
+++ b/scanpipe/pipes/federatedcode.py
@@ -145,7 +145,7 @@ def check_federatedcode_configured_and_available(logger=None):
         logger("Federatedcode repositories are configured and available.")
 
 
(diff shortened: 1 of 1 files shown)


scanpipe/pipes/federatedcode.py
```diff
@@ -145,15 +145,19 @@ def check_federatedcode_configured_and_available(logger=None):
         logger("Federatedcode repositories are configured and available.")
 
 
-def clone_repository(repo_url, logger=None):
+def clone_repository(repo_url, full_history=False, logger=None):
```
keshav-space: How about `shallow_clone` for more clarity?

```suggestion
def clone_repository(repo_url, shallow_clone=True, logger=None):
```

PR #1893, opened 2025-10-01
Fix placeholder text for Download URLs field to use correct purl-spec…

### LINKED ISSUE #1882: Download URLs inputs field shows invalid package url
In ScanCode.io v35.3.0 the Download URLs field under Input shows it accepts package URLs with format `pkg://type/name@version` but per https://github.com/package-url/purl-spec it should be like `pkg:type/name@version`

Propose we change `pkg://type/name@version` to `pkg:type/name@version` and maybe add an example like `pkg:pypi/django@1.11.1` to end up with something like `pkg:type/name@version (e.g.: pkg:pypi/django@1.11.1)`

Aryan-SINGH-GIT:
Replaced incorrect `pkg://type/name@version` with correct `pkg:type/name@version`
according to the official purl-spec. Added example `pkg:pypi/django@1.11.1` for clarity.

Fixes #1882 

Signed-off-by: Aryan Singh <aryansingh12oct2005@gmail.com>

### DIFF
--- a/scanpipe/forms.py
+++ b/scanpipe/forms.py
@@ -74,7 +74,7 @@ class InputsBaseForm(forms.Form):
                 "placeholder": (
                     "https://domain.com/archive.zip\n"
                     "docker://docker-reference (e.g.: docker://postgres:13)\n"
-                    "pkg://type/name@version"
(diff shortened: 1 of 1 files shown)


tdruez:
@Aryan-SINGH-GIT Thanks for the fix! 👍

PR #1894, opened 2025-10-06
Update commit_and_push_changes function to add a default commit message

ziadhany:
- Add a default print logger to prevent exceptions
- Bump minecode_pipelines version to 0.0.1b24

### DIFF
--- a/scanpipe/pipes/federatedcode.py
+++ b/scanpipe/pipes/federatedcode.py
@@ -198,12 +198,31 @@ def commit_and_push_changes(
     commit_message=None,
     purls=None,
     remote_name="origin",
-    logger=None,
+    mine_type="packageURL",
+    tool_name="pkg:github/aboutcode-org/scancode.io",
+    tool_version=VERSION,
+    logger=print
 ):
     """
     Commit and push changes to remote repository.
     Returns True if changes are successfully pushed, False otherwise.
     """
+
+    if not commit_message:
+        author_name = settings.FEDERATEDCODE_GIT_SERVICE_NAME
+        author_email = settings.FEDERATEDCODE_GIT_SERVICE_EMAIL
+
+        purls = "\n".join(purls)
+        commit_message = textwrap.dedent(f"""\
+        Add {mine_type} results for:
+        {purls}
+
+        Tool: {tool_name}@v{tool_version}
+        Reference: https://{settings.ALLOWED_HOSTS[0]}
+
+        Signed-off-by: {author_name} <{author_email}>
+        """)
+
     try:
         commit_changes(repo, files_to_commit, commit_message, purls, logger)
         push_changes(repo, remote_name)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -121,7 +121,7 @@ android_analysis = [
   "android_inspector==0.0.1"
 ]
 mining = [
-  "minecode_pipelines==0.0.1b23"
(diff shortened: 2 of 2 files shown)


scanpipe/pipes/federatedcode.py
```diff
@@ -198,12 +198,31 @@ def commit_and_push_changes(
...
+        purls = "\n".join(purls)
+        commit_message = textwrap.dedent(f"""\
+        Add {mine_type} results for:
+        {purls}
+
+        Tool: {tool_name}@v{tool_version}
+        Reference: https://{settings.ALLOWED_HOSTS[0]}
+
+        Signed-off-by: {author_name} <{author_email}>
+        """)
```
keshav-space, line 224: Can you explain why we need to compute commit_message here? when we already take care of missing commit message in `commit_changes` function.

keshav-space (changes requested):
@ziadhany See comment below.
Also why use `print`? why not simply pass `self.log` while calling `commit_and_push_changes`? Spraying custom logging mechanisms in pipeline functions defeats the whole purpose of having a centrally controlled logger managed by the pipeline, which helps in proper dissemination of pipeline logs to the run instance and UI.

ziadhany:
@keshav-space The main error comes from If a pipeline fails before updating a checkpoint, when a new pipeline runs, it starts and tries to commit some data that was already pushed. This raises the error:
`Nothing to commit, working tree clean` This disrupts the pipeline and causes it to fail

One way to handle this is to use `commit_and_push_changes`, since it pushes and commits at the same time and can catch `GitCommandError`.

Why add a print logger by default?

- This prevents pipeline disruption and avoids unreadable errors such as `'NoneType' object is not callable`.' Since the logger is optional, using a print statement with a `False` return provides a simple and safe fallback.

Why add commit_and_push_changes?

- To avoid duplicating commit messages for every ecosystem (e.g., CRAN, Swift, Composer).

keshav-space:
@ziadhany 
> The main error comes from If a pipeline fails before updating a checkpoint, when a new pipeline runs, it starts and tries to commit some data that was already pushed. This raises the error: `Nothing to commit, working tree clean` This disrupts the pipeline and causes it to fail
> 
> One way to handle this is to use `commit_and_push_changes`, since it pushes and commits at the same time and can catch `GitCommandError`.

Please use `commit_and_push_changes` we already handle empty commit scenario see https://github.com/aboutcode-org/scancode.io/pull/1888, no special error handling is needed on call side.

> Why add a print logger by default?
> 
> This prevents pipeline disruption and avoids unreadable errors such as 'NoneType' object is not callable.' Since the logger is optional, using a print statement with a False return provides a simple and safe fallback.

Let's not use `print` as default logger and if needed we should instead make logger required field that would be much simpler.


> Why add commit_and_push_changes?
> 
> To avoid duplicating commit messages for every ecosystem (e.g., CRAN, Swift, Composer).

I don't get you here.
Can you explain why the below fallback commit is not sufficient for CRAN, Swift and Composer? https://github.com/aboutcode-org/scancode.io/blob/23b94ade588a412c4bcf17d895cdb7c9aeaa2277/scanpipe/pipes/federatedcode.py#L229-L242 

And why would we need to compute default commit message in `commit_and_push_changes` when we already have default commit message in downstream `commit_changes` function?

ziadhany:
@keshav-space Ok, I get your point. I didn’t notice we already have a default commit message, and I will close this PR.

PR #1921, opened 2025-10-28
Update "find_jvm_packages" to include all resources

chinyeungli:
- Ensure extra_data is populated in "find_jvm_packages()" for all resources for mapping purposes
- Adjust test logic to update extra_data after map_checksum, matching d2d pipeline sequence

Signed-off-by: Chin Yeung Li <tli@nexb.com>


scanpipe/pipes/d2d.py
```diff
@@ -262,9 +262,7 @@ def find_jvm_packages(project, jvm_lang: jvm.JvmLanguage, logger=None):
 
     Note: we use the same API as the ScanCode scans by design
     """
-    resources = (
-        project.codebaseresources.files().no_status().from_codebase().has_no_relation()
-    )
+    resources = project.codebaseresources.files().no_status().from_codebase()
```
TG1999, line 265: What issue or problem this change solves ?

chinyeungli: This is for https://github.com/aboutcode-org/scancode.io/issues/1854
This update addresses a scenario where both the development and deployment codebases contain the same `.java` file, but the deployed version also includes the corresponding `.class` file.

I think the tool relies on the "extra_data" field such as `extra_data={"java_package": "org.apache.flume.node"},` (https://github.com/aboutcode-org/scancode.io/blob/main/scanpipe/tests/pipes/test_d2d.py#L402) to perform a mapping from `.class` to `.java`. (I may be wrong, but it seems I need this field to trigger the "path" mapping in the test)

However, if a `.java` file in the development codebase has already been checksum-matched to its counterpart in the deployed codebase, it won't be indexed again, meaning it won’t receive extra_data and won’t be available for mapping.

Since "extra_data" is generated in find_jvm_packages(), this change ensures that all source files in the from_codebase are indexed, even if they’ve already been matched by checksum.

TG1999: Makes sense to me, @tdruez please have a look as well. Thanks!

tdruez:
@chinyeungli we need unit tests related to those changes.

Issue #1808, opened 2025-08-11
Back2Source: Java cases where .class map with .g/.g4


chinyeungli:
Input: antlr4-4.5.1-beta-1
URL:
Devel: https://github.com/antlr/antlr4/archive/refs/tags/4.5.1-beta-1.zip
Deploy: https://github.com/antlr/antlr4/releases/download/4.5.1-beta-1/antlr-4.5.1-beta-1-complete.jar

I ran the `map_deploy_to_develop:Java` pipeline in SCIO and found out there are some strange cases that the following .class files don't have match
```
to/org/antlr/v4/runtime/tree/xpath/XPathLexer.class
to/org/antlr/v4/codegen/SourceGenTriggers.class
to/org/antlr/v4/parse/ActionSplitter.class
to/org/antlr/v4/parse/ANTLRLexer.class
to/org/antlr/v4/parse/ANTLRParser.class
to/org/antlr/v4/parse/ATNBuilder.class
to/org/antlr/v4/parse/BlockSetTransformer.class
to/org/antlr/v4/parse/GrammarTreeVisitor.class
to/org/antlr/v4/parse/LeftRecursiveRuleWalker.class
```
but these actually have matches with
```
from/antlr4-4.5.1-beta-1/runtime/Java/src/org/antlr/v4/runtime/tree/xpath/XPathLexer.g4
from/antlr4-4.5.1-beta-1/tool/src/org/antlr/v4/codegen/SourceGenTriggers.g
from/antlr4-4.5.1-beta-1/tool/src/org/antlr/v4/parse/ActionSplitter.g
from/antlr4-4.5.1-beta-1/tool/src/org/antlr/v4/parse/ANTLRLexer.g
from/antlr4-4.5.1-beta-1/tool/src/org/antlr/v4/parse/ANTLRParser.g
from/antlr4-4.5.1-beta-1/tool/src/org/antlr/v4/parse/ATNBuilder.g
from/antlr4-4.5.1-beta-1/tool/src/org/antlr/v4/parse/BlockSetTransformer.g
from/antlr4-4.5.1-beta-1/tool/src/org/antlr/v4/parse/GrammarTreeVisitor.g
from/antlr4-4.5.1-beta-1/tool/src/org/antlr/v4/parse/LeftRecursiveRuleWalker.g
```

Reference: https://github.com/aboutcode-org/scancode.io/issues/1510

chinyeungli:
[...]

PR #1942, opened 2025-10-29
Refine D2D with Grammar (.g/.g4) Class Mapping #1808

chinyeungli:
Screenshot 2025-10-29 113847

Support mapping `.class` to their corresponding grammar definitions (.g / .g4) files.

Note that grammar files are not exclusive to Java; they can also be used for Python, C++, PHP, and other languages.
This PR focuses solely on `.class` mapping.
Further enhancements can be made over time.

### DIFF
--- a/scanpipe/pipelines/deploy_to_develop.py
+++ b/scanpipe/pipelines/deploy_to_develop.py
@@ -80,6 +80,9 @@ class DeployToDevelop(Pipeline):
             cls.find_kotlin_packages,
             cls.map_kotlin_to_class,
             cls.map_jar_to_kotlin_source,
+            cls.find_grammar_packages,
+            cls.map_grammar_to_class,
+            cls.map_jar_to_grammar_source,
             cls.map_javascript,
             cls.map_javascript_symbols,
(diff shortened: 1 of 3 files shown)


scanpipe/pipelines/deploy_to_develop.py
```diff
@@ -235,6 +238,27 @@ def map_jar_to_kotlin_source(self):
             project=self.project, jvm_lang=jvm.KotlinLanguage, logger=self.log
         )
 
+    @optional_step("Grammar")
+    def find_grammar_packages(self):
+        """Find the java package of the .g/.g4 source files."""
+        d2d.find_jvm_packages(
+            project=self.project, jvm_lang=jvm.GrammarLanguage, logger=self.log
```
TG1999, line 245: Is GrammarLanguage a JVM based language?

chinyeungli: Not exactly, but it follows the same basename mapping logic in the Java case.
To make its purpose clearer, perhaps we could rename it to find_grammar_java_packages()?

PR #1943, opened 2025-10-29
Added "META-INF" and "module-info.class" to ignore list

chinyeungli:
- Everything under META-INF/ is generally metadata or configuration files that can be ignored.
- module-info.class is typically a generated file - it's the compiled version of a Java module descriptor that can be ignored as well.

Signed-off-by: Chin Yeung Li <tli@nexb.com>

### DIFF
--- a/scanpipe/tests/pipes/test_d2d.py
+++ b/scanpipe/tests/pipes/test_d2d.py
@@ -1305,7 +1305,7 @@ class ScanPipeD2DPipesTest(TestCase):
 
         self.assertEqual(1, expected)
 
-    def test_scan_ignored_to_files(self):
+    def test_scanpipe_pipes_d2d_scan_ignored_to_files(self):
         to_dir = (
             self.project1.codebase_path / "to/project.tar.zst-extract/META-INF/foo-bar"
         )


scanpipe/tests/pipes/test_d2d.py
```diff
@@ -1289,6 +1305,30 @@ def test_flag_undeployed_resources(self):
 
         self.assertEqual(1, expected)
 
+    def test_scan_ignored_to_files(self):
```
tdruez: The test name is not consistent with the existing.

chinyeungli: @tdruez what do you mean by "The test name is not consistent with the existing.". It is used to test the `d2d.scan_ignored_to_files()`

tdruez: You need to follow the existing naming conventions. See all the tests function start with `test_scanpipe_pipes_d2d_`
Note that the other test you added in the PR respects the convention: `test_scanpipe_pipes_d2d_java_ignore_pattern`

chinyeungli: okay, thank. Updated.

Issue #1906, opened 2025-10-15
JVM d2d: .class file not mapped with .xtend


chinyeungli:
<img width="1076" height="517" alt="Image" src="https://github.com/user-attachments/assets/af196eed-8421-433f-b3c4-695402e88a68" />

Xtend compiles `.xtend` files directly into `.class` files, the `.java` was likely generated temporarily during compilation but not retained

Reference:
deploy: https://github.com/openhab/openhab1-addons/releases/download/v1.6.2/distribution-1.6.2-addons.zip#to
devel: https://github.com/openhab/openhab1-addons/archive/refs/tags/v1.6.2.zip#from

chinyeungli:
@TG1999 I checked xtend is not exactly Java, but it is closely integrated with Java and it runs on JVM. Should we treat it as part of the "Java" step in the D2D pipeline?

TG1999:
@chinyeungli IMO we should have a different step for .xtend mapping. Also can .xtend appear in other ecosystems apart from Java? If yes then we definitely should have a step for .xtend that should run for every D2D

tdruez:
@chinyeungli Can we close this one following the merge of https://github.com/aboutcode-org/scancode.io/pull/1947 ?

chinyeungli:
Yes. Closing as it's been merged in https://github.com/aboutcode-org/scancode.io/pull/1947

PR #1947, opened 2025-11-07
Refine D2D with Xtend and Class Mapping #1906

chinyeungli:
Fix for #1906

Support mapping `.class` to their corresponding `.xtend` files.

The reported .class is now mapped with the corresponding `.xtend`
Screenshot 2025-11-07 103543

Map type is `xtend_to_class`
Screenshot 2025-11-07 103601

### DIFF
--- a/scanpipe/pipelines/deploy_to_develop.py
+++ b/scanpipe/pipelines/deploy_to_develop.py
@@ -83,6 +83,8 @@ class DeployToDevelop(Pipeline):
             cls.find_grammar_packages,
             cls.map_grammar_to_class,
             cls.map_jar_to_grammar_source,
+            cls.find_xtend_packages,
+            cls.map_xtend_to_class,
             cls.map_javascript,
             cls.map_javascript_symbols,
             cls.map_javascript_strings,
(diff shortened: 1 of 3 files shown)


scanpipe/pipelines/deploy_to_develop.py
```diff
@@ -259,6 +261,20 @@ def map_jar_to_grammar_source(self):
             project=self.project, jvm_lang=jvm.GrammarLanguage, logger=self.log
         )
 
+    @optional_step("Xtend")
```
TG1999, line 264: Should extend be an optional step or be part of all JVM based languages like Scala, Java and Kotlin ?

chinyeungli: aren't the code for Java, Scala and Koltin also treat as an optional step
https://github.com/aboutcode-org/scancode.io/blob/f6687869f8febbee59b0c4542b069b1328d961f5/scanpipe/pipelines/deploy_to_develop.py#L180
https://github.com/aboutcode-org/scancode.io/blob/f6687869f8febbee59b0c4542b069b1328d961f5/scanpipe/pipelines/deploy_to_develop.py#L201
https://github.com/aboutcode-org/scancode.io/blob/f6687869f8febbee59b0c4542b069b1328d961f5/scanpipe/pipelines/deploy_to_develop.py#L222

TG1999: Yes, they are. Coz they all are independent of each other. We won't run the same thing for Scala that we are running for Java. But if we are running same things then it wouldn't have been a separate optional step.

chinyeungli: Here is the background.

Xtend is a flexible and expressive dialect of Java (https://eclipse.dev/Xtext/xtend/)
It compiles `.xtend` files into readable `.java` source code, which is then compiled into JVM bytecode (`.class` files).
Xtend uses a source-to-source compiler, meaning it translates Xtend code into Java before passing it to a standard Java compiler.

There are couple ways to compile `.xtend` such as

Maven
-------
```
mvn compile
```

OR

Gradle
------
```
gradle build
```

Xtend plugin will generate `.java` files for the above

OR

Eclipse IDE
------------
Just save the .xtend file (Eclipse automatically generates the corresponding .java file)

TG1999: @chinyeungli makes sense!

PR #1953, opened 2025-11-13
feat: add pipeline specialized in Maven package scan

chinyeungli:
- Issue: https://github.com/aboutcode-org/scancode.io/issues/1763
- Create a pipeline for Maven package


scanpipe/pipelines/scan_maven_package.py
```diff
@@ -0,0 +1,74 @@
...
+        pom_url_list = get_pom_url_list(self.project.input_sources[0], packages)
+        pom_file_list = download_pom_files(pom_url_list)
+        scanned_pom_packages, scanned_dependencies = scan_pom_files(pom_file_list)
+
+        updated_packages = packages + scanned_pom_packages
+        # Replace/Update the package and dependencies section
+        data["packages"] = updated_packages
+        data["dependencies"] = scanned_dependencies
+        with open(self.scan_output_location, "w") as file:
+            json.dump(data, file, indent=2)
```
tdruez: Code logic should not be on the pipeline itself but in dedictated and easilly testable pipe functions

scanpipe/pipelines/scan_single_package.py
```diff
@@ -51,6 +52,7 @@ def steps(cls):
             cls.extract_input_to_codebase_directory,
             cls.extract_archives,
             cls.run_scan,
+            cls.update_package_license_from_resource_if_missing,
```
tdruez: This may have quite an impact on the default `ScanSinglePackage` results. We should probably handle this one separatly of the Maven context.

scanpipe/pipelines/scan_single_package.py
```diff
@@ -126,6 +128,23 @@ def run_scan(self):
...
+    def update_package_license_from_resource_if_missing(self):
+        """Update PACKAGE license from the license detected in RESOURCES if missing."""
+        with open(self.scan_output_location) as file:
+            data = json.load(file)
+            packages = data.get("packages", [])
+            resources = data.get("files", [])
+            if not packages or not resources:
+                return
+
+        updated_packages = update_package_license_from_resource_if_missing(
```
tdruez: We should use database queries instead of manipulating complex dictionaries.

scanpipe/pipes/resolve.py
```diff
@@ -521,3 +525,235 @@ def extract_headers(input_location, extract_fields):
...
+        downloaded_pom = fetch.fetch_http(pom_url)
+        print("download_pom.path", str(downloaded_pom.path))
+        pom_file_dict["pom_file_path"] = str(downloaded_pom.path)
+        pom_file_dict["output_path"] = str(downloaded_pom.path) + "-output.json"
+        pom_file_dict["pom_url"] = pom_url
+        pom_file_list.append(pom_file_dict)
+    return pom_file_list
+
+
+def scan_pom_files(pom_file_list):
```
tdruez: This is too complex, it needs to be refactored as smaller functions

scanpipe/pipes/resolve.py
```diff
@@ -521,3 +525,235 @@ def extract_headers(input_location, extract_fields):
...
+            if scanned_dependencies:
+                for scanned_dep in scanned_dependencies:
+                    # Replace the 'datafile_path' with empty string
+                    # See https://github.com/aboutcode-org/scancode.io/issues/1763#issuecomment-3525165830
+                    scanned_dep["datafile_path"] = ""
+                    scanned_pom_deps.append(scanned_dep)
+    return scanned_pom_packages, scanned_pom_deps
+
+
+def update_package_license_from_resource_if_missing(packages, resources):
```
tdruez: This should be query-based.

tdruez (changes requested):
- Create a new maven pipe module in place of use resolve
- Opening and loading a large file to make edits multiple times in various steps is not great.
- To be discussed: Do we need a dedicated pipeline for just an extra step? Shouldn't the original `scan_single_package` detect that it's a Maven package and apply the necessary? Any reason to keep this new logic separated?

chinyeungli:
@tdruez I’ve updated the code to include support for the "D2D" option.

Screenshot 2025-12-29 113240

The "deploy_to_devel" option is equivalent to the "map_deploy_to_develop" pipeline, which runs on Java, JavaScript, Kotlin, and Scala as these are the languages commonly found in Maven projects.

JonoYang:
@chinyeungli Can you merge in the new changes and fix the failing test for scan_maven_package_single_file (https://github.com/aboutcode-org/scancode.io/actions/runs/26379105691/job/77644819092?pr=1953#step:10:1020)

chinyeungli:
> @chinyeungli Can you merge in the new changes and fix the failing test for scan_maven_package_single_file (https://github.com/aboutcode-org/scancode.io/actions/runs/26379105691/job/77644819092?pr=1953#step:10:1020)

The failing test is related to https://github.com/aboutcode-org/scancode.io/issues/2172 (https://github.com/aboutcode-org/scancode.io/issues/1763#issuecomment-4531559279)

Reference data:
https://github.com/aboutcode-org/scancode.io/blob/1763_pipeline_for_maven/scanpipe/tests/data/jvm/scancodeio_wisp-logging.json#L739
```
      "is_binary": true,
      "is_text": false,
```

However, in SCIO the file reports:
```
      "is_binary": false,
      "is_text": true,
```

So this discrepancy is not something to be fixed in this issue?

chinyeungli:
I decided to change to another test file so that the test is passing.

JonoYang:
There seems to be duplicate commits coming in when you updated your branch, I'm not sure how to resolve this.

chinyeungli:
> There seems to be duplicate commits coming in when you updated your branch, I'm not sure how to resolve this.

Are you referring to the DCO? Should I force it and click “Set DCO to pass”?

tdruez:
I have doubts about this design, and I want to revisit a concern I raised earlier that wasn't addressed: https://github.com/aboutcode-org/scancode.io/pull/1953#pullrequestreview-3480768894

> "Do we need a dedicated pipeline for just an extra step? Shouldn't the original scan_single_package detect that it's a Maven package and apply the necessary? Any reason to keep this new logic separated?"

This new pipeline ScanMavenPackage implementation merges two fundamentally different processing paths: a basic scan and a full D2D analysis, into one class controlled by a d2d_option_enabled boolean. 

The current design will be confusing to maintain and to users.

chinyeungli:
> I have doubts about this design, and I want to revisit a concern I raised earlier that wasn't addressed: [#1953 (review)](https://github.com/aboutcode-org/scancode.io/pull/1953#pullrequestreview-3480768894)
> 
> > "Do we need a dedicated pipeline for just an extra step? Shouldn't the original scan_single_package detect that it's a Maven package and apply the necessary? Any reason to keep this new logic separated?"
> 
> This new pipeline ScanMavenPackage implementation merges two fundamentally different processing paths: a basic scan and a full D2D analysis, into one class controlled by a d2d_option_enabled boolean.
> 
> The current design will be confusing to maintain and to users.

@tdruez  The current pipeline design combines the basic scan and the D2D steps that already exist in other pipelines, with two additional steps:
`fetch_and_scan_remote_pom()` and `update_package_license_from_resource_if_missing()`.

 * `fetch_and_scan_remote_pom()`
If a package doesn't have a pom.xml, try to find and fetch and scan the remote (upstream) pom.xml if exist.

 * `update_package_license_from_resource_if_missing()`
If a package doesn't have a declared/detected package license, but a license is found at the file level, update the package license using the license detected at the file level.

So, I guess the question is is it necessary to have a dedicated pipeline for these 2 extra steps.

@pombredanne any comment on it?

pombredanne:
@chinyeungli I like the dedicated pipeline as it helps focus the analysis, we just need to make sure we have design that @tdruez agrees to. One question is the way to invoke this. IMHO the input should be ONE PURL, and based on that the POM, the src JAR, the binary JAR(s) should be fetched, and possibly also even the corresponding source git repo should be fetched and processed. That way I could run it at the scale of a whole Maven repo, spawning one SCIO project for each PURL. Another question is if we want to also include C/C++ and native binaries steps, but that's for later. Let's discuss the details in the next call, if that works for you and @tdruez

scanpipe/pipelines/scan_maven_package.py
```diff
@@ -0,0 +1,165 @@
...
+        self.collect_and_create_codebase_resources()
+        self.fingerprint_codebase_directories()
+        self.flag_empty_files()
+        self.flag_whitespace_files()
+        self.flag_ignored_resources()
+
+        options = ["Java", "Kotlin", "Scala", "JavaScript"]
+        d2d_config.load_ecosystem_config(pipeline=self, options=options)
+
+        self.map_about_files()
```
pombredanne: I am not sure you want to call all mapping steps inside a step? Should not these be their own steps?

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,268 @@
...
+    updated_packages = packages + scanned_pom_packages
+    # Replace/Update the package and dependencies section
+    data["packages"] = updated_packages
+    data["dependencies"] = scanned_dependencies
+    with open(scan_output_location, "w") as file:
+        json.dump(data, file, indent=2)
+    return scanning_errors
+
+
+def parse_maven_filename(filename):
```
pombredanne: Should this be in ScanCode? Do we have that already somewhere in PurlDB too?

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,268 @@
...
+def get_pom_url_list(input_source, packages):
+    """Generate Maven POM URLs from package metadata or input source."""
+    pom_url_list = []
+    if packages:
+        for package in packages:
+            if package.get("type") == "maven":
+                package_ns = package.get("namespace", "")
+                package_name = package.get("name", "")
+                package_version = package.get("version", "")
+                pom_url = (
```
pombredanne: Using this hardcoded URL means that we cannot run that on any other repo. We need to find a way to avoid that. If we use a PURL as a project input, then the Maven repo will be part of the PURL (most of the times as default), but then we do not have to further dive into that

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,268 @@
...
+
+    if match:
+        artifact_id = match.group(1)
+        version = match.group(2)
+        return artifact_id, version
+    else:
+        return None, None
+
+
+def get_pom_url_list(input_source, packages):
```
pombredanne:
```suggestion
def get_pom_urls(input_source, packages):
```

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,268 @@
...
+    """Generate Maven POM URLs from package metadata or input source."""
+    pom_url_list = []
```
pombredanne: May be use:
```suggestion
    pom_urls = []
```

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,268 @@
...
+    if packages:
```
pombredanne: Not sure the if is needed at all here?

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,268 @@
...
+        for package in packages:
```
pombredanne: What is package here? Could we get an object rather than a dict?

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,268 @@
...
+                )
+                pom_url_list.append(pom_url)
+    if not pom_url_list:
+        from urllib.parse import urlparse
+
+        # Check what's the input source
+        input_source_url = input_source.get("download_url", "")
+
+        parsed_url = urlparse(input_source_url)
+        maven_hosts = {
```
pombredanne: See my comment above: no hardcoding of hosts should be needed

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,268 @@
...
+            "repo1.maven.org",
+            "repo.maven.apache.org",
+            "maven.google.com",
+        }
+        if input_source_url and parsed_url.netloc in maven_hosts:
```
pombredanne: Do not we have code to do that in FetchCode, packageurl-python and/or scancode-toolkit? Also I have a hard time to understand all the parsing with indices. Do you have an example or tests of what you are trying to accomplish?

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,268 @@
...
+                # Only work with input that's a .jar or .aar file
+                return []
+
+    return pom_url_list
+
+
+def construct_pom_url_from_filename(artifact_id, version):
+    """Construct a pom.xml URL from the given Maven filename."""
+    # Search Maven Central for the artifact to get its groupId
+    url = f"https://search.maven.org/solrsearch/select?q=a:{artifact_id}&wt=json"
```
pombredanne: I would like to avoid using search.maven.org as this is an overlay on the main Maven central repo AND this is  not guaranteed to be there AND it does not exist for other Maven repos.

pombredanne: Also how could we get only an artifact and no groupid? Where do we get only a filename and no URL?

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,268 @@
...
+        if is_xml and is_pom:
+            return True
+        else:
+            # It's probably the Maven Central error page
+            return False
+    except requests.RequestException:
+        return False
+
+
+def download_pom_files(pom_url_list):
```
pombredanne: It is best to avoid a _list suffix IMHO
```suggestion
def download_pom_files(pom_url):
```

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,268 @@
...
+    """Fetch the pom file from the input pom_url_list"""
+    pom_file_list = []
+    for pom_url in pom_url_list:
+        pom_file_dict = {}
+        try:
+            downloaded_pom = fetch.fetch_http(pom_url)
```
pombredanne: Could we use fetchcode to fetch instead?

pombredanne (changes requested):
At last some feedback!

chinyeungli:
I've redesigned the code. It now only takes 1 PURL as input.
I've also removed the manual option for d2d. If both the source and binary are available, d2d will be performed automatically, otherwise, only a basic scan is performed.

Example 1:
PURL: pkg:maven/aero/champ/cargojson@1.0
Screenshot 2026-07-14 102531

Without fetching the remote POM and updating the package data, no packages are detected because the fetched binary (https://repo1.maven.org/maven2/aero/champ/cargojson/1.0/cargojson-1.0.jar) doesn't contain a pom.xml.

However, with remote POM fetching and package data updates enabled, we successfully detect 1 package and update 17 dependencies in the summary:
Screenshot 2026-07-14 102430
Screenshot 2026-07-14 102457
Screenshot 2026-07-14 102509
(We can see the POM file was fetched remotely under "Datafile paths".)

Example 2:
PURL: pkg:maven/activemq/activemq-axis@1.2
https://repo1.maven.org/maven2/activemq/activemq-axis/1.2/

Since there is no source available for this package, only a basic scan is performed.
Screenshot 2026-07-14 102622

scanpipe/pipelines/scan_maven_package.py
chinyeungli: I've separated the logic into different steps

scanpipe/pipes/maven.py
chinyeungli: removed as this is not needed.

scanpipe/pipes/maven.py
chinyeungli: Updated code to use PURL as an input.

scanpipe/pipes/maven.py
chinyeungli: Removed as this is not needed.

scanpipe/pipes/maven.py
chinyeungli: Yes, we could, but the issue is do we want to.
Following is the code if we switch to fetchcode which seems making things complicated, OR we can move the "fetch_http" to fetchcode and use it.

```
        import os
        import fetchcode
        downloaded_pom = fetchcode.fetch(pom_url)
        location = str(downloaded_pom.location)
        path = location + ".pom"
        # The fetch function from fetchcode save the file as /tmp/name
        # without an extension. We need to add the ".pom" extension so that
        # the package scan can work properly for this file.
        os.rename(location, path)
```

scanpipe/pipelines/scan_maven_package.py
```diff
@@ -0,0 +1,179 @@
...
+    def extract_input(self):
+        """Extract the input to the codebase directory."""
+        if not self.d2d_enabled:
+            self.extract_input_to_codebase_directory()
+        else:
+            self.extract_inputs_to_codebase_directory()
+
+    def maven_d2d_steps(self):
+        if self.d2d_enabled:
+            """Run D2D steps for Maven projects."""
```
tdruez: Docstrings should be after the function def

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,237 @@
...
+from packageurl.contrib import purl2url
+
+from scanpipe.pipes import fetch
+from scanpipe.pipes import scancode
+
+logger = logging.getLogger(__name__)
+
+
+def check_input_and_return_purl(project):
+    """Validate the input and return a PURL."""
```
tdruez:
```suggestion
    """Validate the input and return a maven PURL."""
```

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,237 @@
...
+    if len(input_sources) != 1:
+        error_msg = "Only 1 maven purl is accepted."
+        raise ValueError(error_msg)
+    project_input = str(input_sources[0])
+    input_purl = PackageURL.from_string(project_input)
+
+    if input_purl.type != "maven":
+        error_msg = "Only maven purl is supported."
+        raise ValueError(error_msg)
+    # Should we hande all available versions if no version is provided?
```
tdruez: Let' not leave todos.

scanpipe/templates/scanpipe/includes/vulnerability_id.html
```diff
@@ -27,4 +27,4 @@
       </li>
     {% endfor %}
   </ul>
-{% endif %}
\ No newline at end of file
+{% endif %}
```
tdruez, line 30: Unwanted change

scanpipe/templates/scanpipe/tabset/tab_vulnerabilities.html
```diff
@@ -35,4 +35,4 @@
       </tr>
     {% endfor %}
   </tbody>
-</table>
\ No newline at end of file
+</table>
```
tdruez, line 38: Unwanted change

scanpipe/views.py
```diff
@@ -2351,6 +2352,7 @@ class DiscoveredPackageDetailsView(
             ],
             "icon_class": "fa-solid fa-bug",
             "template": "scanpipe/tabset/tab_vulnerabilities.html",
+            "tab_context": {"VULNERABLECODE_URL": settings.VULNERABLECODE_URL},
```
tdruez: Unwanted change?

scanpipe/tests/test_pipelines.py
```diff
@@ -922,6 +924,28 @@ def mock_make_git_directory(**kwargs):
         self.assertEqual(2, project1.codebaseresources.count())
         self.assertEqual(0, project1.discoveredpackages.count())
 
+    @skipIf(sys.platform == "darwin", "Not supported on macOS")
+    def test_scanpipe_scan_maven_package_single_file(self):
```
tdruez, line 928: We should have a test for the d2d option as well

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,237 @@
...
+from packageurl import PackageURL
+from packageurl.contrib import purl2url
+
+from scanpipe.pipes import fetch
+from scanpipe.pipes import scancode
+
+logger = logging.getLogger(__name__)
+
+
+def check_input_and_return_purl(project):
```
tdruez, line 39: Lack of unit tests

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,237 @@
...
+        type=input_purl.type,
+        namespace=input_purl.namespace,
+        name=input_purl.name,
+        version=input_purl.version,
+    )
+
+    return purl
+
+
+def fetch_inputs(purl):
```
tdruez, line 59: Lack of unit tests

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,237 @@
...
+    # Fetch the bianry and source for the given input purl and return the
```
tdruez: Typo

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,237 @@
...
+    except Exception as e:
+        logger.info("Failed to fetch binary package: %s", e)
+        purl_bin_path = []
+
+    try:
+        package_src = fetch.fetch_url(url=purl_src)
+        purl_src_path = [package_src.path]
+    except Exception as e:
+        logger.info("Failed to fetch source package: %s", e)
+        purl_src_path = []
```
tdruez: Code duplication
Also, why are path stored as lists?

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,237 @@
...
+
+    if not purl_bin_path and not purl_src_path:
+        err_msg = f"No source or binary could be resolved for {purl}."
+        raise ValueError(err_msg)
+
+    return purl_src_path, purl_bin_path
+
+
+def fetch_and_scan_remote_pom(input_purl, scan_output_location):
```
tdruez, line 82: Lack of unit tests

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,237 @@
...
+        """
+        downloaded_pom = fetch.fetch_http(pom_url)
+        path = str(downloaded_pom.path)
+
+        pom_file_dict["pom_file_path"] = path
+        pom_file_dict["output_path"] = path + "-output.json"
+        pom_file_dict["pom_url"] = pom_url
+    except requests.RequestException:
+        # Return an empty dictionary
+        pass
```
tdruez: Try block should wrap a single instruction

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,237 @@
...
+    return pom_file_dict
+
+
+def scan_pom_file(pom_file_dict):
```
tdruez, line 150: No tests

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,237 @@
...
+            for scanned_package in scanned_packages:
+                # Replace the 'datafile_path' with the pom_url
+                scanned_package["datafile_paths"] = [pom_url]
+                scanned_pom_packages.append(scanned_package)
+        if scanned_dependencies:
+            for scanned_dep in scanned_dependencies:
+                # Replace the 'datafile_path' with empty string
+                # See https://github.com/aboutcode-org/scancode.io/issues/1763#issuecomment-3525165830
+                scanned_dep["datafile_path"] = ""
+                scanned_pom_deps.append(scanned_dep)
```
tdruez: Does this need to be in the context processor?

scanpipe/pipelines/scan_maven_package.py
```diff
@@ -0,0 +1,179 @@
...
+    """
+    Scan a single Maven package and run a D2D scan.
+
+    This pipeline takes a Maven PURL as input, fetches the binary and
+    source archives (if they exist), and then performs scans for package
+    metadata, declared dependencies, licenses, license clarity scores, and
+    copyrights. It also performs a D2D scan if both the source and binary
+    archives are available.
+
+    The output is a summary of the scan results in JSON format.
```
tdruez: Let's not use the D2D wording here since it's the main documentation for the pipeline.

scanpipe/pipelines/scan_maven_package.py
```diff
@@ -0,0 +1,179 @@
...
+        self.collect_and_create_license_detections()
+        self.flag_deployed_from_resources_with_missing_license()
+        self.create_local_files_packages()
+
+    def fetch_and_scan_remote_pom(self):
+        """Fetch and scan remote POM files."""
+        scanning_errors = fetch_and_scan_remote_pom(
+            self.purl, self.scan_output_location
+        )
+        if scanning_errors:
```
tdruez: Is the "if" required?

tdruez (changes requested):

scanpipe/views.py
chinyeungli: It will be removed.

scanpipe/templates/scanpipe/includes/vulnerability_id.html
chinyeungli: This was likely auto‑corrected when I ran `make check`, but in any case the missing tag has now been added and fixed.

scanpipe/templates/scanpipe/tabset/tab_vulnerabilities.html
chinyeungli: This was likely auto‑corrected when I ran `make check`, but in any case the missing tag has now been added and fixed.

scanpipe/pipelines/scan_maven_package.py
chinyeungli: how about 

Scan a single Maven package and perform a deployment to development relation scan.

I think we still want to include the "D2D" context (perhaps in different wording)?

scanpipe/pipes/maven.py
chinyeungli: The path needs to be stored as a list because it is saved to `self.from_files` and `self.to_files`. Later, the pipeline calls `self.extract_inputs_to_codebase_directory()` from deploy_to_develop.py, which expects both `self.from_files` and `self.to_files` to be of list type.

scanpipe/pipes/maven.py
chinyeungli: code updated

scanpipe/pipelines/scan_maven_package.py
```diff
@@ -0,0 +1,182 @@
...
+            self.d2d_javascript()
+            self.d2d_process()
+
+    def d2d_ecosystem_config(self):
+        # The following langages will be included:
+        # - Java
+        # - Kotlin
+        # - Scala
+        # - JavaScript
+        options = ["Java", "Kotlin", "Scala", "JavaScript"]
```
tdruez: We can see that clearly in the options list, the comment is not useful

scanpipe/pipelines/scan_maven_package.py
```diff
@@ -0,0 +1,182 @@
...
+class ScanMavenPackage(ScanSinglePackage, DeployToDevelop):
+    """
+    Scan a single Maven package and perform a deployment to development
+    relation scan.
+
+    This pipeline takes a Maven PURL as input, fetches the binary and
+    source archives (if they exist), and then performs scans for package
+    metadata, declared dependencies, licenses, license clarity scores, and
+    copyrights. It also performs a D2D scan if both the source and binary
+    archives are available.
```
tdruez: Replace the "D2D" wording in user documentation.

scanpipe/pipelines/scan_maven_package.py
```diff
@@ -0,0 +1,182 @@
...
+            self.collect_input_information()
+
+    def extract_input(self):
+        """Extract the input to the codebase directory."""
+        if not self.d2d_enabled:
+            self.extract_input_to_codebase_directory()
+        else:
+            self.extract_inputs_to_codebase_directory()
+
+    def maven_d2d_steps(self):
```
tdruez, line 103: Move the method after the `d2d_*` ones for a more natural reading hierarchy.

scanpipe/pipelines/scan_maven_package.py
```diff
@@ -0,0 +1,182 @@
...
+        for scanning_error in scanning_errors:
+            for resource_path, errors in scanning_error.items():
+                self.project.add_error(
+                    description="\n".join(errors),
+                    model=self.pipeline_name,
+                    details={"resource_path": resource_path.removeprefix("codebase/")},
+                )
+
+    def update_package_license_from_resource_if_missing(self):
+        """Update PACKAGE license from the license detected in RESOURCES if missing."""
```
tdruez: "PACKAGE" and "RESOURCES" are not consistent with the existing codebase conventions.

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,229 @@
...
+    if len(input_sources) != 1:
+        error_msg = "Only 1 maven purl is accepted."
+        raise ValueError(error_msg)
+    project_input = str(input_sources[0])
+    input_purl = PackageURL.from_string(project_input)
+
+    if input_purl.type != "maven":
+        error_msg = "Only maven purl is supported."
+        raise ValueError(error_msg)
+    # Version is required
```
tdruez: Comment is not useful

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,229 @@
...
+    if not input_purl.version:
+        error_msg = "Version is required."
+        raise ValueError(error_msg)
+
+    purl = PackageURL(
+        type=input_purl.type,
+        namespace=input_purl.namespace,
+        name=input_purl.name,
+        version=input_purl.version,
+    )
```
tdruez: Why recreating a `PackageURL` from `input_purl`?

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+def fetch_and_scan_remote_pom(input_purl, scan_output_location):
+    """Fetch the .pom file from from maven.org if not present in codebase."""
+    with open(scan_output_location) as file:
+        data = json.load(file)
+        # Return and do nothing if data has pom.xml
+        for file in data["files"]:
+            if "pom.xml" in file["path"]:
+                return []
```
tdruez: Same issue as previously reported, does the parsing need to be wrapped in the context processor?

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,229 @@
...
+    import os
+    import fetchcode
+    downloaded_pom = fetchcode.fetch(pom_url)
+    location = str(downloaded_pom.location)
+    path = location + ".pom"
+    # The fetch function from fetchcode save the file as /tmp/name
+    # without an extension. We need to add the ".pom" extension so that
+    # the package scan can work properly for this file.
+    os.rename(location, path)
+    """
```
tdruez: This is not the right place for discussing the implementation.
Move this to the issue or the PR, not in the code.

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,229 @@
...
+    try:
+        downloaded_pom = fetch.fetch_http(pom_url)
+    except requests.RequestException:
+        # Return an empty dictionary
```
tdruez: `return {}` is explicit enough the comment is not adding any value

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,229 @@
...
+        return {}
+    path = str(downloaded_pom.path)
+    pom_file_dict = {}
+    pom_file_dict["pom_file_path"] = path
+    pom_file_dict["output_path"] = path + "-output.json"
+    pom_file_dict["pom_url"] = pom_url
+    return pom_file_dict
```
tdruez:
```suggestion
    path = str(downloaded_pom.path)
    return {
        "pom_file_path": path,
        "output_path": f"{path}-output.json",
        "pom_url": pom_url,
    }
```

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,229 @@
...
+    scanning_errors = scancode.run_scan(
+        location=pom_file_path,
+        output_file=scanned_pom_output_path,
+        run_scan_args={
+            "package": True,
+        },
+    )
+    if scanning_errors:
+        scan_errors.append(scanning_errors)
+    return scan_errors
```
tdruez: This is confusing, could you clarify?

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,229 @@
...
+    pom_url = pom_file_dict.get("pom_url", "")
+
+    with open(scanned_pom_output_path) as scanned_pom_file:
+        scanned_pom_data = json.load(scanned_pom_file)
+
+    scanned_packages = scanned_pom_data.get("packages", [])
+    scanned_dependencies = scanned_pom_data.get("dependencies", [])
+
+    for scanned_package in scanned_packages:
+        # Replace the 'datafile_path' with the pom_url
```
tdruez: Avoid comments that restates the code

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,229 @@
...
+        scanned_package["datafile_paths"] = [pom_url]
+        scanned_pom_packages.append(scanned_package)
+    for scanned_dep in scanned_dependencies:
+        # Replace the 'datafile_path' with empty string
```
tdruez: Avoid comments that restates the code

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,229 @@
...
+        # See https://github.com/aboutcode-org/scancode.io/issues/1763#issuecomment-3525165830
+        scanned_dep["datafile_path"] = ""
+        scanned_pom_deps.append(scanned_dep)
+    return scanned_pom_packages, scanned_pom_deps
+
+
+def update_package_license_from_resource_if_missing(project):
+    """Populate missing licenses to packages based on resource data."""
+    from license_expression import Licensing
```
tdruez: Inline import?

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,229 @@
...
+            for resource in project.codebaseresources.has_license_expression():
+                for for_package in resource.for_packages:
+                    if for_package == package_uid:
+                        detected_lic_exp = resource.detected_license_expression
+                        if detected_lic_exp not in detected_lics:
+                            detected_lics.append(detected_lic_exp)
+            if detected_lics:
+                lic_exp = " AND ".join(detected_lics)
+                declared_lic_exp = str(Licensing().dedup(lic_exp))
+                package.declared_license_expression = declared_lic_exp
```
tdruez: `lics` `exp` avoid those, use the real words

scanpipe/tests/pipes/test_maven.py
```diff
@@ -0,0 +1,309 @@
...
+        )
+        self.assertEqual(result, ["Failed to download the POM file."])
+
+    @mock.patch("builtins.open", new_callable=mock.mock_open)
+    @mock.patch("json.load")
+    @mock.patch("json.dump")
+    @mock.patch("scanpipe.pipes.maven.get_pom_url")
+    @mock.patch("scanpipe.pipes.maven.download_pom_file")
+    @mock.patch("scanpipe.pipes.maven.scan_pom_file")
+    @mock.patch("scanpipe.pipes.maven.update_datafile_paths")
```
tdruez: Wow, what is the purpose of this test? what are we even testing here?

scanpipe/tests/test_pipelines.py
tdruez: Still missing

tdruez (changes requested):

scanpipe/pipelines/scan_maven_package.py
chinyeungli: what does that mean?

tdruez: Cut and paste this method after d2d_process for better readability

scanpipe/pipelines/scan_maven_package.py
chinyeungli: Do you mean I should use word "DiscoveredPackage" and "CodebaseResource" ?

tdruez: "Fill in missing package licenses using licenses detected in their resources."

scanpipe/pipelines/scan_maven_package.py
chinyeungli: I have no problem doing that, but don’t we want the methods’ order to follow the steps’ order?

scanpipe/tests/test_pipelines.py
chinyeungli: I've added `self.assertEqual(11, project1.codebaserelations.count())` to check for the d2d relation scan.

Note that the d2d requires the source_inspector which is not supported in macos: https://github.com/aboutcode-org/scancode.io/blob/main/uv.lock#L2984

as the source_inspector will not be installed and will throw error:
https://github.com/aboutcode-org/scancode.io/actions/runs/29320243229/job/87054870086#step:7:1074

scanpipe/tests/pipes/test_maven.py
chinyeungli: I've broken down the `fetch_and_scan_remote_pom()` and added the new `update_scan_data()`, then created a test there.

scanpipe/pipes/maven.py
chinyeungli: code refactored and added comment.

scanpipe/pipelines/scan_maven_package.py
```diff
@@ -0,0 +1,183 @@
...
+        self.purl = check_input_and_return_purl(self.project)
+
+    def fetch_inputs(self):
+        """Fetch the binary and source of the given PURL."""
+        from_file, to_file = fetch_inputs(self.purl)
+        # The pipeline will call self.extract_inputs_to_codebase_directory()
+        # from deploy_to_develop.py, which expects both self.from_files and
+        # self.to_files to be of list type.
+        self.from_files = [from_file]
+        self.to_files = [to_file]
```
tdruez: `fetch_inputs` can return None for either file, so `self.from_files` becomes `[None]`, which is True. `d2d_check` then enables D2D even when only one archive exists, and `collect_input_info` may set `input_path` to `None`.

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,204 @@
...
+def fetch_path(url, package_type):
+    """Fetch the url and return the location of the fetched tarball"""
+    try:
+        return fetch.fetch_url(url=url).path
+    except Exception as e:
+        logger.info("Failed to fetch %s package: %s", package_type, e)
+        return None
+
+
+def fetch_and_scan_remote_pom(input_purl, scan_output_location):
```
tdruez, line 82: The return type is inconsistent: `[]`, a list of strings, or `[scanning_errors]` where `scanning_errors` is whatever `run_scan` returns (possibly `None` or `{}`). The caller does `scanning_error.items()`, which raises `AttributeError` on strings and `None`. Also `return [scanning_errors]` wraps a false success value into a non-empty list.

Please pick one error shape and return `[]` on success.

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,204 @@
...
+    with open(scan_output_location, "w") as file:
+        json.dump(updated_data, file, indent=2)
+    return [scanning_errors]
+
+
+def update_scan_data(data, scanned_packages, scanned_dependencies):
+    """Update packages and dependencies data"""
+    data["packages"] = data.get("packages", []) + scanned_packages
+    data["dependencies"] = scanned_dependencies
+    return data
```
tdruez, line 113: Packages are appended but existing dependencies are fully replaced. If intentional (no pom.xml in codebase implies no maven dependencies), a short comment explaining why would help, otherwise dependencies should be merged like packages.

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,204 @@
...
+        scanned_package["datafile_paths"] = [pom_url]
+        scanned_pom_packages.append(scanned_package)
+    for scanned_dep in scanned_dependencies:
+        # See https://github.com/aboutcode-org/scancode.io/issues/1763#issuecomment-3525165830
+        scanned_dep["datafile_path"] = ""
+        scanned_pom_deps.append(scanned_dep)
+    return scanned_pom_packages, scanned_pom_deps
+
+
+def update_package_license_from_resource_if_missing(project):
```
tdruez, line 190: This re-scans all project resources for every package.. The package/resource relation is already in the database, so this could be done at the queryset level, filtering `codebaseresources.has_license_expression()` by `for_packages` containment per package.

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,204 @@
...
+
+    return input_purl
+
+
+def fetch_inputs(purl):
+    """Fetch the binary and source for the given input purl"""
+    purl_str = PackageURL.to_string(purl)
+
+    purl_bin_path = fetch_path(purl_str, "binary")
+    purl_src_path = fetch_path(f"{purl_str}?classifier=sources", "source")
```
tdruez, line 64: `fetch_path` is called with a PURL string but its parameter is named `url`. Rename the parameter or clarify the docstring.

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,204 @@
...
+        raise ValueError(err_msg)
+
+    return purl_src_path, purl_bin_path
+
+
+def fetch_path(url, package_type):
+    """Fetch the url and return the location of the fetched tarball"""
+    try:
+        return fetch.fetch_url(url=url).path
+    except Exception as e:
```
tdruez: `except Exception` with `logger.info` silently swallows programming errors, `logger.warning` and a narrower exception type would be safer.

scanpipe/tests/pipes/test_maven.py
```diff
@@ -0,0 +1,270 @@
...
+    @mock.patch("builtins.open", new_callable=mock.mock_open)
+    @mock.patch("json.load")
+    @mock.patch("scanpipe.pipes.maven.get_pom_url")
+    @mock.patch("scanpipe.pipes.maven.download_pom_file")
+    def test_scanpipe_maven_fetch_and_scan_remote_pom_no_pom_file(
+        self, mock_download_pom_file, mock_get_pom_url, mock_json_load, mock_open
+    ):
+        mock_json_load.return_value = {"files": []}
+        mock_get_pom_url.return_value = "https://example.com/test.pom"
+        mock_download_pom_file.return_value = None
```
tdruez: `download_pom_file` returns `{}` on failure but the mock here returns `None`. It passes thanks to the truth check, but the test doesn't exercise the real contract. Either make the function return `None` on failure or fix the mock to return `{}`.

pyproject.toml
```diff
@@ -166,6 +166,7 @@ resolve_dependencies = "scanpipe.pipelines.resolve_dependencies:ResolveDependenc
 scan_codebase = "scanpipe.pipelines.scan_codebase:ScanCodebase"
 scan_for_virus = "scanpipe.pipelines.scan_for_virus:ScanForVirus"
 scan_single_package = "scanpipe.pipelines.scan_single_package:ScanSinglePackage"
+scan_maven_package = "scanpipe.pipelines.scan_maven_package:ScanMavenPackage"
```
tdruez, line 168: `scan_maven_package` should go before `scan_single_package`

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,204 @@
...
+from packageurl import PackageURL
+from packageurl.contrib import purl2url
+
+from scanpipe.pipes import fetch
+from scanpipe.pipes import scancode
+
+logger = logging.getLogger(__name__)
+
+
+def check_input_and_return_purl(project):
```
tdruez, line 39: Since the PURL is no longer rebuilt, input qualifiers are now kept. A purl submitted with `?classifier=sources` will make `fetch_inputs` produce `?classifier=sources?classifier=sources`. Either strip qualifiers here (with a docstring note) or reject purls with qualifiers explicitly.

tdruez (changes requested):
@chinyeungli a few issues still need to be addressed. Please also add or update unit tests to cover them, so we catch regressions on those.

scanpipe/pipes/maven.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+def fetch_and_scan_remote_pom(input_purl, scan_output_location):
+    """Fetch the .pom file from from maven.org if not present in codebase."""
+    with open(scan_output_location) as file:
+        data = json.load(file)
+    # Return and do nothing if data has pom.xml
+    for file in data["files"]:
+        if "pom.xml" in file["path"]:
+            return []
```
tdruez:
```suggestion

    # Skip fetching the remote POM if a pom.xml is already present in the codebase.
    for file_entry in data["files"]:
        if PurePosixPath(file_entry["path"]).name == "pom.xml":
            return []
```

This would be a bit cleaner, mak esure to add this import too: `from pathlib import PurePosixPath`

tdruez (changes requested):
@chinyeungli see one last suggestion before the merge/

scanpipe/pipes/maven.py
chinyeungli: updated

PR #1955, opened 2025-11-14
Alternative worker to avoid TLS connection issues to DB

rogu-beta:
The tech stack of django-rq seems to have issues with TLS connections getting corrupted when using `Worker` (forks/multiprocessing). This is relevant when an external database is used. The alternative is to also implement a worker based on `SimpleWorker` which does not fork and does not experience this bug. This alternative worker could be configured via the `--worker-class` option of the `rqworker`, just as the `scancodeio.worker.ScanCodeIOWorker` is currently set in the docker compose files. For reference see: 
- https://github.com/aboutcode-org/scancode.io/issues/1523
- https://github.com/rq/django-rq/issues/650 
- https://github.com/aboutcode-org/scancode.io/blob/a89488815cad3e1e8987c0ddc98afff147d43c2a/docker-compose.yml#L86

### DIFF
--- a/scancodeio/worker.py
+++ b/scancodeio/worker.py
@@ -23,13 +23,43 @@
 from django.apps import apps
 
 from rq.queue import Queue
-from rq.worker import Worker
+from rq.worker import Worker, SimpleWorker
 
 scanpipe_app = apps.get_app_config("scanpipe")
 
 
 class ScanCodeIOWorker(Worker):
     """Modified version of RQ Worker including ScanCode.io customizations."""
+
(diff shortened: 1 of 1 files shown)


rogu-beta:
Due to the lack of heartbeats during job execution the scans sometimes seem to get stuck and thus the workers receive a SIGTERM. It does still work better than the regular worker, given the TLS connection issues, but it is not a full solution.

rogu-beta:
Another solution is proposed in https://github.com/aboutcode-org/scancode.io/issues/1523#issuecomment-3558213573

PR #1959, opened 2025-11-20
Close existing connection before job execution in worker

rogu-beta:
The tech stack of django-rq seems to have issues with TLS connections getting corrupted when using the Worker class (forks). This is relevant when an external database is used. This problem can be avoided by closing the existing connection prior to the call to `execute_job`. For reference see: 
- https://github.com/aboutcode-org/scancode.io/issues/1523
- https://github.com/rq/django-rq/issues/650

### DIFF
--- a/scancodeio/worker.py
+++ b/scancodeio/worker.py
@@ -31,14 +31,6 @@ scanpipe_app = apps.get_app_config("scanpipe")
 class ScanCodeIOWorker(Worker):
     """Modified version of RQ Worker including ScanCode.io customizations."""
 
-    def execute_job(self, job, queue):
-        """
-        Terminate existing connection to avoid issues with TLS.
-        https://github.com/aboutcode-org/scancode.io/issues/1523
-        """
-        connection.close()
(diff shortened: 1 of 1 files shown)


rogu-beta:
@tdruez Sure, no problem. That change has been moved to a separate class.

tdruez:
@rogu-beta Perfect!

rogu-beta:
@tdruez I just noticed the unit test failing. For some reason the import statement went missing. I'm very sorry about that. Should I open another PR or would you be so kind to add `from django.db import connection`?

tdruez:
Let me add it in main ;)

tdruez:
Fixed in https://github.com/aboutcode-org/scancode.io/commit/cfb1df9893128a1a62fae65755db5c4b3d9e61a1

rogu-beta:
@tdruez Thank you! By the way, the patch has worked very reliably for me, just yesterday I ran a scan of over 600 packages with 8 worker on a Kubernetes deployment with zero issues.

Issue #1875, opened 2025-09-17
Compiled `.class` May Differ from Source Filename - map_deploy_to_devel (scala)


chinyeungli:
In Scala, a compiled `.class` file may not directly correspond to its source filename. For example, `Hello.scala` can produce both `Hello.class` and `World.class`.

Use case
----------
Project: pekko-cluster-sharding-typed

source: https://repo1.maven.org/maven2/org/apache/pekko/pekko-cluster-sharding-typed_3/1.2.0/pekko-cluster-sharding-typed_3-1.2.0-sources.jar
binary: https://repo1.maven.org/maven2/org/apache/pekko/pekko-cluster-sharding-typed_3/1.2.0/pekko-cluster-sharding-typed_3-1.2.0.jar

With filename mapping:
source
```
from/org/apache/pekko/cluster/sharding/typed/ClusterShardingQuery.scala
```
binary
```
to/org/apache/pekko/cluster/sharding/typed/ClusterShardingQuery.class
to/org/apache/pekko/cluster/sharding/typed/GetClusterShardingStats.class
to/org/apache/pekko/cluster/sharding/typed/GetShardRegionState.class
```
Only the `to/org/apache/pekko/cluster/sharding/typed/ClusterShardingQuery.class` are considered mapped.
`GetClusterShardingStats.class` and `GetShardRegionState.class` will be considered as "missing source".

However, if look into the `ClusterShardingQuery.scala`, 
```
final case class GetShardRegionState(...)
.
.
.
final case class GetClusterShardingStats(...)
```
`GetClusterShardingStats.class` and `GetShardRegionState.class` are inner class and should be treated as compiled from `ClusterShardingQuery.scala
`

chinyeungli:
Similar to https://github.com/aboutcode-org/scancode.io/issues/1993 , the solution will likely be similar.

PR #1969, opened 2025-11-30
Fix issue 1875 scala case classes

pradhyum6144:
Problem
In Scala, case classes and inner classes defined within sealed traits or objects compile to separate .class files that may not follow the standard $-separated naming convention.
Solution
Implemented a fallback mechanism specifically for Scala that searches for source files in the same package directory when an exact match is not found.
Testing
All syntax checks passed
New test case covers the specific scenario described in issue #1875
Maintains backward compatibility with existing Java, Kotlin, and other JVM language mappings

### DIFF
--- a/scanpipe/models.py
+++ b/scanpipe/models.py
@@ -2498,9 +2498,10 @@ class CodebaseResourceQuerySet(ComplianceAlertQuerySetMixin, ProjectRelatedQuery
         )
 
     def executable_binaries(self):
-        return self.win_exes().order_by().union(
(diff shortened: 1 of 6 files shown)


pradhyum6144:
Hey @chinyeungli  I've created a fix for this issue! I’d like you to review my code and accept it.

chinyeungli:
@tdruez perhaps you are in a better position for the review

pradhyum6144:
@tdruez can you please review this and merge it.
thank you.

PR #1975, opened 2025-12-02
Add support for CycloneDX spec v1.7

tdruez:
- new default version for outputs


mjherzog:
We may want to wait until CDX 1.7 standard is approved later this month. I would recommend postponing 1.7 as default until at least Jan 2026

tdruez:
@mjherzog Noted. CycloneDX 1.6 will remain the default for now.

Issue #1754, opened 2025-07-23
Remove the dependency on `scipy`


tdruez:
`scipy` was introduced in https://github.com/aboutcode-org/scancode.io/pull/1622

The only usage of this lib is the `scipy.spatial.distance.jensenshannon` but the price to pay is enormous.

`scipy` is a large library and have a few heavy dependencies itself, such as `numpi`, see https://github.com/scipy/scipy/blob/main/pyproject.toml#L19-L32

This results in over 140M of Python code added to the stack for a simple function that could be implemented as a few lines of Python.

Let's remove this dependency and replace with a simple implementation of the Jensen-Shannon divergence in the symbolmap module.

PR #1977, opened 2025-12-03
Remove the dependency on scipy #1754

tdruez:
Fixes: https://github.com/aboutcode-org/scancode.io/issues/1754


keshav-space (approved):
Thanks @tdruez, LGTM!

PR #1985, opened 2025-12-05
Fix typos in docstrings for pathmap.py

pranavkeerti66-hue:


tdruez:
Hey @pranavkeerti66-hue thank you for your contribution! Could you fix the DCA so we can merge this one?

Issue #1987, opened 2025-12-08
Support new Authors attribute in ORTs package-list.yml


FabianRolfMatthiasNoll:
Previously when scanning with Scancode and creating a package-list.yml there was no way to transport author / maintainer information betweent ORT and Scancode.

On the ORT side the support was already implemented:
https://github.com/oss-review-toolkit/ort/commit/3bc2f50a65456f70b232d37dd3635089d120983c

Which only needs to be adapted on the ScanCode side.

tdruez:
@FabianRolfMatthiasNoll Thanks again for your help on this, released as https://github.com/aboutcode-org/scancode.io/releases/tag/v36.0.1

PR #1988, opened 2025-12-08
Add support for authors in ORT package list generation

FabianRolfMatthiasNoll:
This updates the ORT pipe to include the 'authors' field in the generated package list YAML. The authors are extracted from the package parties where the role is 'maintainer'.

Ref: #1987


tdruez:
@FabianRolfMatthiasNoll Thanks for the PR, looks good!

PR #1997, opened 2025-12-12
Regen test fixtures and fix tests after scancode-toolkit upgrade

AyanSinhaMahapatra:
Reference: https://github.com/aboutcode-org/scancode.io/pull/1984

### DIFF
--- a/scanpipe/tests/pipes/test_output.py
+++ b/scanpipe/tests/pipes/test_output.py
@@ -339,9 +339,7 @@ class ScanPipeOutputPipesTest(TestCase):
         call_command("loaddata", fixtures, **{"verbosity": 0})
 
         project = Project.objects.get(name="asgiref")
-        package = project.discoveredpackages.get(
-            uuid="856001ca-aabf-4c6a-8ac2-cc1317f3e05d"
-        )
+        package = project.discoveredpackages.get(datasource_ids=["pypi_wheel_metadata"])
 
         package.other_license_expression_spdx = "Apache-2.0 AND LicenseRef-test"
         data = self.data / "cyclonedx/django-4.0.10-vulnerability.json"


scanpipe/pipes/spdx.py
```diff
@@ -30,9 +30,10 @@
 from pathlib import Path
 
 import saneyaml
+from scancode_config import spdx_license_list_version
 
 SCHEMAS_LOCATION = Path(__file__).parent / "schemas"
-SPDX_LICENSE_LIST_VERSION = "3.20"
```
AyanSinhaMahapatra, line 35: It is probably best we import this value from scancode where it is updated once we update the license list. Otherwise this might be quite outdated like it was presently.

scanpipe/tests/regen_test_data.py
```diff
@@ -75,7 +76,8 @@ def test_regen_asgiref_test_files(self):
         run = project1.add_pipeline(pipeline_name)
         pipeline = run.make_pipeline_instance()
 
-        exitcode, _ = pipeline.execute()
+        with override_settings(SCANCODEIO_PROCESSES=-1):
```
AyanSinhaMahapatra, line 79: I always have my CPU/memory max out and crash my desktop session on my local machine wherever I regen the SCIO tests using `./manage.py test --pattern "regen*.py"
` so I'm updating the regen tests function here to disable multiprocessing. This is fast without multiprocessing so this is fine.

AyanSinhaMahapatra:
@tdruez I had some more changes related to https://github.com/aboutcode-org/scancode.io/pull/1984 as that was not complete.

I cannot reproduce the warnings in https://github.com/aboutcode-org/scancode.io/pull/1984#issue-3697602008 though I think I've seen them before. It could be related to the click failures which we fixed earlier. Let me know if they still persist. Same about the `UNSET_files` but I can track these seperately. 

We are also failing on the scorecard data checks as I think the scorecard API data format has changed. Also opening an issue and investigating/fixing this seperately.

scanpipe/tests/pipes/test_output.py
```diff
@@ -340,7 +340,7 @@ def test_scanpipe_pipes_outputs_to_cyclonedx(self, regen=FIXTURES_REGEN):
 
         project = Project.objects.get(name="asgiref")
         package = project.discoveredpackages.get(
-            uuid="7969de5e-5589-4441-bffa-a60e12b43280"
```
AyanSinhaMahapatra, line 343: We always have to search for this and update this manually, maybe we should consider a different approach to get the package instance here?

tdruez: Fixed in https://github.com/aboutcode-org/scancode.io/pull/1997/commits/7f40d105687296afff3c48a338ea43c073ff9850

tdruez:
> I cannot reproduce the warnings in #1984 (comment) though I think I've seen them before. It could be related to the click failures which we fixed earlier. Let me know if they still persist. Same about the UNSET_files but I can track these seperately.

@AyanSinhaMahapatra This is an issue with commoncode, entered as https://github.com/aboutcode-org/scancode.io/issues/1998

PR #2005, opened 2025-12-19
fix: Sanitize ORT package IDs to handle colons in versions

### LINKED ISSUE #2004: ORT import failing for some debian packages
**Describe the bug**

So if a image has the a package like this:
```
- id: deb::passwd:1:4.13+dfsg1-4ubuntu3.2
    purl: pkg:deb/ubuntu/passwd@1:4.13%2Bdfsg1-4ubuntu3.2?arch=amd64
    vcs:
      type:
      url:
      revision:
      path:
    sourceArtifact:
      url:
    declaredLicenses:
      - BSD-3-Clause
    description: |
      change and administer password and group data
       This package includes passwd, chsh, chfn, and many other programs to
       maintain password and group data.
       .
       Shadow passwords are supported.  See /usr/share/doc/passwd/README.Debian
    homepageUrl: https://github.com/shadow-maint/shadow
    authors:
      - Ubuntu Developers
```

the ort import from the package-list will fail because the id is not correctly formatted.
As ORT uses : to separate different part of the identifier, the parts themselves must not contain :. So if any original part does contain :, it needs to be escaped. 

The Error:
```
Caused by: java.lang.IllegalArgumentException: An identifier's properties must not contain ':' because that character is used as a separator in the string representation: type='deb', namespace='', name='passwd', version='1:4.13+dfsg1-4ubuntu3.2'.
```

**System configuration**
- Which version of ScanCode.io are you running? latest
- Are you running the app using Docker? yes
- On which OS? linux runner
- What inputs are you using? container image
- Which pipeline are you running? image pipeline

**To Reproduce**
1. Take a image with the above mentioned package
2. Create a package-list
3. Import it into ORT

**Expected behavior**
No error

**Screenshots**

FabianRolfMatthiasNoll:
ORT uses colons as separators in identifiers. Package versions containing colons (e.g. Debian epochs) were breaking the identifier format. This change replaces colons with underscores in the ID components.

closes #2004

### DIFF
--- a/scanpipe/pipes/ort.py
+++ b/scanpipe/pipes/ort.py
@@ -130,14 +130,28 @@ def get_ort_project_type(project):
         return "docker"
 
 
+def sanitize_id_part(value):
+    """
+    Sanitize an identifier part by replacing colons with underscores.
+    ORT uses colons as separators in the identifier string representation.
+    """
+    if value:
+        return value.replace(":", "_")
+    return value
+
+
 def to_ort_package_list_yml(project):
     """Convert a project object into a YAML string in the ORT package list format."""
     project_type = get_ort_project_type(project)
 
     dependencies = []
     for package in project.discoveredpackages.all():
+        type_ = sanitize_id_part(project_type or package.type)
+        name = sanitize_id_part(package.name)
+        version = sanitize_id_part(package.version)
+
         dependency = Dependency(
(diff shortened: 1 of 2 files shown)


tdruez:
@FabianRolfMatthiasNoll Looks good, thanks for this fix!

PR #2012, opened 2025-12-25
Fix Java D2D mapping when compiled .class name differs from source fi…

Adityakk9031:
Problem

The current Deploy-to-Develop (D2D) Java mapping assumes that a compiled .class file name always matches its source .java filename.

This assumption breaks for some projects (e.g. Lombok), where:

The declared class name does not match the source file name

As a result, valid Java sources are not mapped to their compiled .class files

Example from Lombok:

Tasks.class ← source file is DelombokTask.java

AnnotationProcessorHider.class ← source file is AnnotationProcessor.java

This causes these .class files to be reported as having no corresponding source mapping.

Solution

This PR improves Java D2D mapping by:

Extracting the declared class name from .java source files

Using the declared class name (instead of only the filename) to build the JVM mapping index

Allowing .class files to correctly map even when the source filename differs

The change is limited to Java JVM mapping logic and does not affect other ecosystems.

Related issue

Fixes: Compiled .class May Differ from Source Filename – map_deploy_to_devel (Java)
🔗 https://github.com/aboutcode-org/scancode.io/issues/1993

### DIFF
--- a/scanpipe/pipes/d2d.py
+++ b/scanpipe/pipes/d2d.py
@@ -201,6 +201,47 @@ def _map_jvm_to_class_resource(
             )
 
 
+def extract_fqcn_from_java(location_path):
+    """
+    Extract fully qualified class/interface/enum names from a Java source file.
+
+    Example:
+        package to.lombok.delombok.ant;
+        class Tasks {}
+
+    Returns:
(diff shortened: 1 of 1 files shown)


Adityakk9031:
@keshav-space please have a look

Issue #2019, opened 2025-12-29
scan_for_virus with ClamAV docker container failed: CodebaseResource matching query does not exist


cw1427:
Version: v36.0.1

Error log
```
CodebaseResource matching query does not exist.

Traceback:
  File "/opt/scancodeio/aboutcode/pipeline/__init__.py", line 199, in execute
    step(self)
    ~~~~^^^^^^
  File "/opt/scancodeio/scanpipe/pipelines/scan_for_virus.py", line 40, in scan_for_virus
    clamav.scan_for_virus(self.project)
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
  File "/opt/scancodeio/scanpipe/pipes/clamav.py", line 50, in scan_for_virus
    resource = project.codebaseresources.get(path=resource_path)
  File "/opt/scancodeio/.venv/lib/python3.13/site-packages/django/db/models/manager.py", line 87, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/opt/scancodeio/.venv/lib/python3.13/site-packages/django/db/models/query.py", line 635, in get
    raise self.model.DoesNotExist(
        "%s matching query does not exist." % self.model._meta.object_name
    )
```

<img width="499" height="450" alt="Image" src="https://github.com/user-attachments/assets/90f909ad-7f65-42d5-b02d-64d7dca38ffa" />

docker compose
 clamav:
    image: ******/clamav/clamav:latest
    volumes:
      - ./clamav_data:/var/lib/clamav
      - ./workspace:/var/scancodeio/workspace/
    restart: always

tdruez:
@cw1427 Thanks for reporting. Could you clarify which inputs you are using? As well as all the pipeline you are running before the `scan_for_virus` one.

cw1427:
@tdruez  I used scan_codebase which I can't pass the scancode parameter to scan the -cple(copyright pacpage link email) only used its default paramters? 

<img width="569" height="167" alt="Image" src="https://github.com/user-attachments/assets/a890824a-6578-4c1b-96ba-bfa5a4ad6846" />

After that I adjust the scancode.io source code to avoid scan some ignore files that scan_codebase bypassed but the sacn_for_virus didn't. then it passed.

[...]

PR #2020, opened 2025-12-29
Fix scan_for_virus crash when CodebaseResource is missing

Adityakk9031:
Fix:#2019
Issue
The scan_for_virus pipeline fails with CodebaseResource matching query does not exist when ClamAV reports a file path that is not present in the database.

Solution
Convert the scanned resource path to a string before querying and safely handle missing CodebaseResource entries by recording a project error and skipping the resource instead of crashing the pipeline.

### DIFF
--- a/scanpipe/pipes/clamav.py
+++ b/scanpipe/pipes/clamav.py
@@ -23,6 +23,7 @@
 from pathlib import Path
 
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


tdruez:
Fixed in https://github.com/aboutcode-org/scancode.io/pull/2190

PR #2028, opened 2026-01-05
Restructure documentation index

AyanSinhaMahapatra:
Reference: https://github.com/aboutcode-org/www.aboutcode.org/issues/43


DennisClark:
@AyanSinhaMahapatra   thanks for the fixes!

Issue #1918, opened 2025-10-23
Direct dependencies missing if package is used as both direct and transitive dependency


rogu-beta:
**Describe the bug**
The dependency graph gathered from `load_sbom` does not accurately represent what is contained within the SBOM. If a package is both a direct dependency and transitive dependency at the same time, the listing will only show the package as a transitive dependency.

In the example given below, the package `pkg:npm/%40angular/animations@18.2.9` is both a dependency to the global `dejacode-demo` representing the project and the package `pkg:npm/@angular/material@18.2.9`. This can be seen in the section:

```
"dependencies": [
        {
            "ref": "pkg:npm/dejacode-demo",
            "dependsOn": [
                "pkg:npm/@angular/animations@18.2.9",
                "pkg:npm/@angular/common@18.2.9",
                "pkg:npm/@angular/compiler@18.2.9",
                "pkg:npm/@angular/core@18.2.9",
                "pkg:npm/@angular/forms@18.2.9",
                "pkg:npm/@angular/material@18.2.9",
                "pkg:npm/@angular/platform-browser-dynamic@18.2.9",
                "pkg:npm/@angular/platform-browser@18.2.9",
                "pkg:npm/@angular/router@18.2.9",
                "pkg:npm/@jsverse/transloco@7.6.1",
                "pkg:npm/@ngrx/effects@18.1.1",
                "pkg:npm/@ngrx/store@18.1.1",
                "pkg:npm/ngx-echarts@19.0.0",
                "pkg:npm/ngx-toastr@17.0.2",
                "pkg:npm/tslib@2.8.1"
            ]
        },
```
```
{
            "ref": "pkg:npm/@angular/material@18.2.9",
            "dependsOn": [
                "pkg:npm/@angular/animations@18.2.9",
                "pkg:npm/@angular/cdk@18.2.9",
                "pkg:npm/@angular/common@18.2.9",
                "pkg:npm/@angular/core@18.2.9",
                "pkg:npm/@angular/forms@18.2.9",
                "pkg:npm/@angular/platform-browser@18.2.9",
                "pkg:npm/rxjs@7.8.2",
                "pkg:npm/tslib@2.8.1"
            ]
        },
```

However, the result in ScanCode.io only shows `pkg:npm/@angular/animations@18.2.9` as a dependency of `pkg:npm/@angular/material@18.2.9`.

**System configuration**
- ScanCode.io 35.4.0
- Running with custom Helm deployment
- Linux
[...]

PR #2035, opened 2026-01-21
Fix missing project-level dependencies when importing SPDX SBOMs #1918

Monal-Reddy:
When importing SPDX SBOMs, dependencies originating from the document root (SPDXRef-DOCUMENT) were treated as package-to-package dependencies instead of project-level dependencies.
This caused direct dependencies to be lost when the same package also appeared transitively.
This change maps SPDXRef-DOCUMENT to project-level dependencies (for_package=None), preserving both direct and transitive relationships.


Monal-Reddy:
hi @tdruez  , could you please take a look at this PR when you have time? It addresses the SPDX dependency issue in #1918. Thankyou!

PR #2049, opened 2026-02-07
Improve SPDX import by generating generic PURLs when missing

### LINKED ISSUE #1914: Improve SPDX import from Python SBOMs
**Describe the bug**

I run a the `load_sbom` pipeline using this input
https://www.python.org/ftp/python/3.13.9/Python-3.13.9.tgz.spdx.json

The SBOM has a lot of problems including:
- listing each and every file in relationships to embedded packages which creates an SBOM of 93,000+ lines. Not sure we can do much about that
- the lack of package type and PURLs for some generic packages, that we then report as "unknown" type, like with "pkg:unknown/expat@2.7.3" when we could do better especially when there are download URLs and we could instead create a proper "generic" PURL
- pip vendored dependencies are reported but not that they are patched or the download URL may be misleading.... I would be surprised that Python 3.13's pip 25.2 depends on a msgpack binary wheel built for macOS. 

Overall there are issues in the SPDX, but there are also issues on what we do with it.

For reference, attached are the SPDX, its export to CycloneDX from SCIO and also its re-export to SPDX from SCIO:

[Python-3.13.9.tgz.spdx.json](https://github.com/user-attachments/files/23022087/Python-3.13.9.tgz.spdx.json)
[scancodeio_sbom-round-trip_results-2025-10-21-12-59-02.cdx.json](https://github.com/user-attachments/files/23022089/scancodeio_sbom-round-trip_results-2025-10-21-12-59-02.cdx.json)
[scancodeio_sbom-round-trip_results-2025-10-21-13-44-58.spdx.json](https://github.com/user-attachments/files/23022088/scancodeio_sbom-round-trip_results-2025-10-21-13-44-58.spdx.json)

**System configuration**
- Which version of ScanCode.io are you running? 35.4.0
- Are you running the app using Docker? yes
- On which OS? linux
- What inputs are you using? see above
- Which pipeline are you running? load_sbom

Monal-Reddy:
Fixes #1914

This PR improves the `load_sbom` pipeline handling of SPDX SBOMs by generating a fallback `pkg:generic` PURL when an SPDX package does not provide an explicit PURL but does include a name and version.

This avoids classifying such packages as `pkg:unknown`, improving package typing and downstream reporting for SPDX-based Python SBOMs.

Tested locally using ScanCode.io with:
- Python-3.13.9.tgz.spdx.json
- `load_sbom` pipeline

Previously affected packages such as CPython, expat, libb2, and mpdecimal are now correctly represented as `pkg:generic/*`.

### DIFF
--- a/scanpipe/pipes/resolve.py
+++ b/scanpipe/pipes/resolve.py
@@ -314,6 +314,7 @@ def convert_spdx_expression(license_expression_spdx):
 def spdx_package_to_package_data(spdx_package):
     """Convert the provided spdx_package into package_data."""
     package_url_dict = {}
+
     # Store the original "SPDXID" as package_uid for dependencies resolution.
     package_uid = spdx_package.spdx_id
 
@@ -321,6 +322,14 @@ def spdx_package_to_package_data(spdx_package):
         if ref.type == "purl":
             purl = ref.locator
             package_url_dict = PackageURL.from_string(purl).to_dict(encode=True)
+
+    if not package_url_dict and spdx_package.name and spdx_package.version:
+        generic_purl = PackageURL(
+            type="generic",
+            name=spdx_package.name,
+            version=spdx_package.version,
+        )
+        package_url_dict = generic_purl.to_dict(encode=True)
 
     checksum_data = {
         checksum.algorithm.lower(): checksum.value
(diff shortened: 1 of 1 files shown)


Monal-Reddy:
For reference, here is the result after applying this fix locally.

Packages previously classified as pkg:unknown/* are now resolved as pkg:generic/* when loading the Python 3.13.9 SPDX SBOM via load_sbom.

Image
image

I found this an interesting area, if this looks good to you, i'd be happy to work on the subsequent issues as u have mentioned

PR #2065, opened 2026-02-23
SPDX: Replace deprecated documentDescribes with DESCRIBES relationships

gyanranjanpanda:
While working on the SPDX output, I noticed we're still emitting the `documentDescribes` top-level field even though it was deprecated in the SPDX 2.3 specification in favour of explicit `DESCRIBES` relationships inside the `relationships` array (see [spdx/spdx-spec#395](https://github.com/spdx/spdx-spec/issues/395)).

This PR removes `documentDescribes` from serialization and emits the equivalent `DESCRIBES` relationships instead, prepended so that document-level entries appear before package-level ones, keeping the ordering deterministic.

Deserialization is fully backward compatible. If an older SPDX document still uses `documentDescribes`, `Document.from_data()` transparently converts it, so existing test data and externally generated SBOMs continue to load correctly.

**What changed:**
- `Document.as_dict()` no longer emits `documentDescribes`; it now prepends `DESCRIBES` relationships derived from the internal `describes` list.
- - `Document.from_data()` reconstructs `describes` from the legacy field if present, or from `DESCRIBES` relationships where `spdxElementId` matches the document ID.
- - The `asgiref-3.3.0.spdx.json` fixture is updated accordingly.
- - Two new tests cover the happy path and the backward-compat round-trip.
**Testing:**
```
Ran 26 tests in 0.167s

### DIFF
--- a/scanpipe/pipes/spdx.py
+++ b/scanpipe/pipes/spdx.py
@@ -607,7 +607,6 @@ class Document:
             "SPDXID": self.spdx_id,
             "name": self.safe_document_name(self.name),
             "documentNamespace": self.namespace,
-            "documentDescribes": self.describes,
             "creationInfo": self.creation_info.as_dict(),
             "packages": [package.as_dict(self.version) for package in self.packages],
         }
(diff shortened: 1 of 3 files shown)


tdruez:
@gyanranjanpanda We do not plan to change the behavior of `documentDescribes`.
See https://github.com/aboutcode-org/scancode.io/issues/564#issuecomment-3269296563, https://github.com/spdx/spdx-spec/issues/395, and https://github.com/aboutcode-org/scancode.io/pull/1856

PR #2070, opened 2026-02-26
Fix crash in assemble_package when manifest is at codebase root

Kaushik-Kumar-CEG:
Fixes the issue reported in : **scancode-action** [#11](https://github.com/aboutcode-org/scancode-action/issues/11)

when a package manifest (e.g. pom.xml) is at the root of the input directory, `assemble_package()` crashes with:

    AttributeError: 'NoneType' object has no attribute 'path'

this happens because packagedcode's maven handler calls `resource.parent()`  which returns **None** when there is no parent path, then passes that **None**  to `assign_package_to_resources()`

### Fix

wrapped `handler.assemble()` in `assemble_package()` with a `try/except AttributeError` so the pipeline logs a warning and continues instead of crashing

### Testing

reproduced and verified locally

### DIFF
--- a/scanpipe/pipes/scancode.py
+++ b/scanpipe/pipes/scancode.py
@@ -619,16 +619,22 @@ def assemble_package(resource, project, processed_paths):
             package_adder=add_resource_to_package,
         )
 
-        for item in extracted_items:
-            logger.info(f"    Processing item: {item}")
-            if isinstance(item, packagedcode_models.Package):
-                pipes.update_or_create_package(project, item.to_dict())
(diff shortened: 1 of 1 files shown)


Kaushik-Kumar-CEG:
Hi @pombredanne @AyanSinhaMahapatra @tdruez @keshav-space 

this fixes the issue reported in **scancode-action** [#11 ](https://github.com/aboutcode-org/scancode-action/issues/11) 

happy to make any changes if needed! 
please let me know your thoughts :)

tdruez:
Catching the exception here just silences the error, it doesn't address why it's happening in the first place.

Issue #1786, opened 2025-07-30
show full filename for source/binary archives in resources


AyanSinhaMahapatra:
In the d2d pipeline inputs, often the filename of the archive has important clues about origin and build (like os/arch built for) among other things. So it would be nice if we can show the full filename of the archive on the `to` and `from` sides so this is clear.

For example, for [Shouldly.4.0.0-beta0003.nupkg](https://github.com/shouldly/shouldly/releases/download/v.4.0.0-beta0003/Shouldly.4.0.0-beta0003.nupkg):

<img width="888" height="1750" alt="Image" src="https://github.com/user-attachments/assets/66ff9dd3-1e64-4ee6-bc42-3ccc96c2aae9" />

The `to` side should have a `Shouldly.4.0.0-beta0003.nupkg` initial directory in front of the `to` resource paths.

PR #2071, opened 2026-02-28
Show full filenames for source resources

uttam282005:
closes-  #1786

before-
image

after-
image


uttam282005:
most of the diff bloat is due to the regenerated expected files for tests that where previously failing.

PR #2132, opened 2026-03-30
Fix 2116: Fix duplicate input filename by adding unique naming

Monal-Reddy:
## Issues

- Closes: #2116

## Changes

Fixed issue where multiple input files with the same filename were being overwritten. if a file with the same name already exists in the input directory, a numeric suffix (like 2, 3...) is added to create a unique filename instead of overwriting the existing file.Added a unit test to verify that duplicate filenames are handled correctly.

## Checklist
- [x] I have read the [contributing guidelines](https://github.com/aboutcode-org/scancode.io/blob/main/CONTRIBUTING.md)
- [x] I have linked an existing issue above
- [x] I have added unit tests covering the new code
- [x] I have reviewed and understood every line of this PR

image

### DIFF
--- a/scanpipe/pipes/input.py
+++ b/scanpipe/pipes/input.py
@@ -66,6 +66,12 @@ def copy_inputs(input_locations, dest_path):
 def move_input(input_location, dest_path):
     """Move the provided ``input_location`` to the ``dest_path``."""
     destination = dest_path / Path(input_location).name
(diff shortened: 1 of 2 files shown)


tdruez:
Merged in https://github.com/aboutcode-org/scancode.io/pull/2193

PR #2181, opened 2026-06-25
Fix xlsx_errors column overwriting the last data column

prajakta128:
## What
Fix `xlsx_errors` column writing error messages to the wrong column, overwriting the last data column's value.

## Why
`errors_col_index = len(fields) - 1` points to the last data column (zero-indexed). But `xlsx_errors` is appended AFTER all fields, so its correct index is `len(fields)`.

For example with 3 fields [col_a, col_b, col_c]:
- Column index 2 = col_c  (last data column)  ← bug writes here
- Column index 3 = xlsx_errors                ← should write here

## How to reproduce
Export any project to XLSX where a field value exceeds 32,767 chars. The truncation error message overwrites col_c instead of appearing in the xlsx_errors column.

## Fix
One line changed in `scanpipe/pipes/output.py`:
- `errors_col_index = len(fields) - 1`
+ `errors_col_index = len(fields)`

Added a regression test in `scanpipe/tests/pipes/test_output.py`.

Fixes #2088

### DIFF
--- a/scanpipe/pipes/output.py
+++ b/scanpipe/pipes/output.py
@@ -382,7 +382,7 @@ def add_xlsx_worksheet(workbook, worksheet_name, rows, fields):
     worksheet.write_row(row=0, col=0, data=header, cell_format=cell_format)
 
     errors_count = 0
-    errors_col_index = len(fields) - 1  # rows and cols are zero-indexed
+    errors_col_index = len(fields)  # "xlsx_errors" column is appended after all fields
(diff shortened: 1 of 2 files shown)


prajakta128:
I noticed PR #2117 addresses the same issue but has not been merged yet.
My PR takes a similar approach. Happy to collaborate or close this one
if the maintainers prefer to proceed with #2117.

PR #2182, opened 2026-06-25
Fix get_data_from_manifests returning wrong type on early exit

prajakta128:
## What
Fix `get_data_from_manifests` returning a bare `[]` instead of `([], [])` when no manifest resources exist.

## Why
All callers unpack the return value as a two-tuple:

    packages, dependencies = resolve.get_data_from_manifests(...)

This is done in two pipelines:
- `scanpipe/pipelines/load_sbom.py` line 59
- `scanpipe/pipelines/resolve_dependencies.py` line 87

When `manifest_resources.exists()` is False, the function hit:

    return []   ← wrong type

This causes a `ValueError: not enough values to unpack` crash, crashing the entire LoadSBOM and ResolveDependencies pipelines
whenever a project has no manifest resources.

## Fix
One character change in `scanpipe/pipes/resolve.py`:

- return []
+ return [], []

Added a regression test to verify the function always returns a two-tuple even when manifest_resources is empty.

### DIFF
--- a/scanpipe/pipes/resolve.py
+++ b/scanpipe/pipes/resolve.py
@@ -95,7 +95,7 @@ def get_data_from_manifests(project, package_registry, manifest_resources, model
             description="No resources containing package data found in codebase.",
             model=model,
         )
-        return []
+        return [], []
 
     for resource in manifest_resources:
         packages = resolve_manifest_resources(resource, package_registry)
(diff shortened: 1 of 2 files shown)


prajakta128:
Hi maintainers! This PR fixes `get_data_from_manifests` returning a bare `[]` instead of `([], [])` when no manifest resources exist, which causes a `ValueError` crash in the LoadSBOM and ResolveDependencies pipelines. Would love to get your feedback when you have a chance. Happy to make any changes needed. Thank you!

Issue #2179, opened 2026-06-23
scanpipe as stand-alone library


JonoYang:
In the purldb, I want to be able to match resources/directories in a codebase to see if they are part of an indexed package. I am getting the codebase from a scancode.io scan where fingerprints have been computed for the directories and resources.

Currently, I do this by creating a separate django project with scanpipe in it and import all the scancode.io settings (https://github.com/aboutcode-org/purldb/tree/main/matchcode_project), and run it in the same docker compose file as purldb. This setup is untenable and I want to make it work better. I am wondering if it's reasonable to have scanpipe as a django app that I can add to a django project, and create a codebase and run pipelines without having to instantiate the entirety of scancode.io.

I want (from my purldb django app) to be able to create a scanpipe project (instantiate codebase, resources, packages, dependencies) from this codebase scan and run a pipeline that checks fingerprints against the purldb using the django orm.

The aboutcode.pipelines library provides the framework for the pipelines and execution, but does not have the other stuff for dealing with codebases, resources, and packages. 

Alternatively, is there a way to run scancode.io in such a way that I can:

- Add purldb to DATABASES in scancode.io settings
- Query the purldb using the django orm from scancode.io without scancode.io trying to manage the model or having to declare purldb as an app in scancode.io

tdruez:
Scancode.io is already available as a wheel, so we should mostly have everything we need to use it as a library within another Django app. 

The main issue in the current state is the setting management that would need some adjustments for this app library setup.
For now, we can `from scancodeio.settings import *` but I'm expecting some conflicts with this approach.

## POC:

### Setup a new Django app
```
uv init app --python 3.14
cd app
uv add django scancodeio
uv run django-admin startproject config .
source .venv/bin/activate
```

### Edit config/settings.py
Import ScanCode.io settings in the imports
```
from scancodeio.settings import *
```

Add `scanpipe` in `INSTALLED_APPS`
```
INSTALLED_APPS = [
    ...
    'scanpipe',
]
```

### Run migrations to create DB tables
```
./manage.py migrate
```

### Manage projects from Python
```
./manage.py shell
```
```
from scanpipe.management.commands import create_project

project = create_project(
    name="projet1",
    pipelines=["scan_single_package"],
    input_urls=["https://registry.npmjs.org/lodash/-/lodash-4.17.21.tgz"],
    execute=True,
)

# project.discoveredpackages.all()
# etc...
```

### Manage projects from CLI
```
./manage.py create-project projet1 \
[...]

PR #2187, opened 2026-06-30
feat: move scanpipe settings to local settings.py with lazy resolutio…

tdruez:
…n and SCANPIPE dict override

## Issues

- Issue: #2179

## Changes

Move all scanpipe app-level settings out of scancodeio/settings.py into a new scanpipe/settings.py module. This makes scanpipe usable as a standalone library in other Django apps without requiring the host to copy dozens of settings.

The new module exposes a single scanpipe_settings object that resolves each value lazily in order: SCANPIPE dict in Django settings, then the SCANCODEIO_ environment variable, then the built-in default. Tests use override_settings(SCANPIPE={"KEY": value}) instead of flat SCANCODEIO_KEY overrides, which works automatically via Django's setting_changed signal.

scancodeio/settings.py now only contains infrastructure settings that Django needs at boot time: database, Redis, logging, REST framework configuration.


JonoYang:
Some notes when using this branch of scio in purldb:

- I had to add taggit in addition to scanpipe in INSTALLED_APPS for a migration
- I had to add the following envvars to the purldb settings, as some pipelines import variables from the settings file:
```
PURLDB_URL = ""
PURLDB_USER = ""
PURLDB_PASSWORD = ""
PURLDB_API_KEY = ""
MATCHCODEIO_URL = ""
MATCHCODEIO_USER = ""
MATCHCODEIO_PASSWORD = ""
MATCHCODEIO_API_KEY = ""
VULNERABLECODE_URL = ""
VULNERABLECODE_USER = ""
VULNERABLECODE_PASSWORD = ""
VULNERABLECODE_API_KEY = ""
VULNERABLECODE_USER_AGENT = ""
```
- I am now failing some purldb api tests with the error message ` minecode/tests/test_api.py::ScannableURIAPITestCase::test_api_scannable_uri_get_next_download_url - django.db.utils.ProgrammingError: relation "authtoken_token" does not exist`. I do not get this error when I use scancodeio from pypi

tdruez:
> I had to add taggit in addition to scanpipe in INSTALLED_APPS for a migration

Dependency on taggit removed in https://github.com/aboutcode-org/scancode.io/pull/2189
Taggit had also consequence on performances and not needed for our basic "label" usage.

> I had to add the following envvars to the purldb settings, as some pipelines import variables from the settings file:

Ive updated the code in https://github.com/aboutcode-org/scancode.io/pull/2187/changes/ac93c1dde15709e40629eb85ded79e15bff0c5d2 so you do not have to defined those in your main settings.

> I am now failing some purldb api tests with the error message  minecode/tests/test_api.py::ScannableURIAPITestCase::test_api_scannable_uri_get_next_download_url - django.db.utils.ProgrammingError: relation "authtoken_token" does not exist. I do not get this error when I use scancodeio from pypi

The dependency on `authtoken_token` was removed in https://github.com/aboutcode-org/scancode.io/pull/2087
See https://github.com/aboutcode-org/scancode.io/commit/79d7927506e218dabb3c8e69ca99c895096e325c#diff-f11933ef5e81f3fbaf94814cdbab8cfefa04a17d5ee619df9d659b8e71604450

> I do not get this error when I use scancodeio from pypi

Could you confirm why exact version is running?

JonoYang:
> The dependency on authtoken_token was removed in https://github.com/aboutcode-org/scancode.io/pull/2087
> See https://github.com/aboutcode-org/scancode.io/commit/79d7927506e218dabb3c8e69ca99c895096e325c#diff-f11933ef5e81f3fbaf94814cdbab8cfefa04a17d5ee619df9d659b8e71604450

Thanks, the issue is happening here (https://github.com/aboutcode-org/purldb/blob/main/minecode/api.py#L198) where I am calling `get_user_model()`. I have `rest_framework.authtoken` declared in my INSTALLED_APPS, but I don't know why the migration wouldn't run.

> Could you confirm why exact version is running?

I have purldb using scancode.io version 35.5.0

Besides the issue above, I had to add `django.contrib.postgres` to INSTALLED_APPS, and everything else seems to be fine.

tdruez:
> I have rest_framework.authtoken declared in my INSTALLED_APPS

Is the main "purldb" app has a real dependency on `rest_framework.authtoken`?

The scanpipe migration at https://github.com/aboutcode-org/scancode.io/commit/79d7927506e218dabb3c8e69ca99c895096e325c#diff-956b348f9974f70e47874f79d45f4465f2998f2f6c9f1999d651b0366a51456eR53-R58 forces the deletion of the `authtoken_token` table, which is likely the source of the issue if purldb depends on it.

@JonoYang could you run the `test_api_scannable_uri_get_next_download_url` with full verbosity on it?
Also, can you confirm the dependency on `rest_framework.authtoken` for API auth?

Once confirmed, I'll suggest that we start by migrating the PurlDB replacing `rest_framework.authtoken` with `aboutcode.api_auth` to https://github.com/aboutcode-org/dejacode/tree/main/aboutcode/api_auth for enhanced security and consistency across the tools.

See https://github.com/aboutcode-org/scancode.io/commit/79d7927506e218dabb3c8e69ca99c895096e325c for the implementation on the SCIO side.

JonoYang:
Thanks, the purldb users use rest_framework.authtoken. I will update it to use aboutcode.api_auth

